PR #67, opened 2015-08-21
Fix some typos in Localizable.strings

aceman444:
Is this the right place to fix the typos so that they are also picked up in Weblate for translation?


vshcherb:
They are also available in Weblate, both places are fine and weblate doesn't require personal acceptance

PR #69, opened 2016-02-14
OsmAnd in Apple Watch

and01:
I have written an extension for the Apple Watch. Check it out..


WalterGlobal:
> I have written an extension for the Apple Watch. Check it out..

How do I start using this? Sorry for being clueless.

2xsaiko:
Would love to have this. Might try rebasing it when I have time but I have never touched iOS development before so no idea how far I’d get.

PR #138, opened 2018-09-13
Fixed an issue when the import of files wouldn't work on cold start

PaulStets:


Sources/Controllers/Help/OAHelpViewController.mm
```diff
@@ -0,0 +1,366 @@
...
+        [cell.textView setText:item[@"title"]];
+        [cell.descView setText:item[@"description"]];
+        [cell showImage:NO];
+    }
+    return cell;
+}
+
+- (CGFloat) tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    return 50.0;
```
crimean: Are you sure 50 is enough? What if text will not fit this height? Also method for estimating row's height is a must.

PR #327, opened 2019-06-11
Add support for WunderLINQ, motorcycle smartphone control.

kconger:
Hi,

I'm one of the developers of the [WunderLINQ](http://www.wunderlinq.com). The WunderLINQ enables you to use your motorcycle's handlebar controls to control your smartphone. Essentially the device acts as a Bluetooth keyboard.

My patch to the Android version was [merged](https://github.com/osmandapp/Osmand/pull/7000) recently, this is the iOS equivalent

Here is a video of the integration in action: [https://youtu.be/gTXtVagOQqk](https://youtu.be/gTXtVagOQqk)

### DIFF
--- a/Sources/AppHost/OAAppDelegate.mm
+++ b/Sources/AppHost/OAAppDelegate.mm
@@ -217,11 +217,11 @@
                 if (title.length > 0)
                     targetPoint.title = title;
                 if ([navigate  isEqual: @"navigate"]){
-                    [[OARootViewController instance].mapPanel navigate:targetPoint];
-                    [[OARootViewController instance].mapPanel closeRouteInfo];
-                    [[OARootViewController instance].mapPanel startNavigation];
+                    [_rootViewController.mapPanel navigate:targetPoint];
+                    [_rootViewController.mapPanel closeRouteInfo];
+                    [_rootViewController.mapPanel startNavigation];
                 } else {
-                    [[OARootViewController instance].mapPanel showContextMenu:targetPoint];
+                    [_rootViewController.mapPanel showContextMenu:targetPoint];
                 }
             });
             
--- a/Sources/AppHost/OAAppDelegate.mm
+++ b/Sources/AppHost/OAAppDelegate.mm
@@ -188,36 +188,42 @@
     }
     else if ([scheme isEqualToString:@"osmandmaps"])
     {
-        NSDictionary *params = [OAUtilities parseUrlQuery:url];
+		NSDictionary *params = [OAUtilities parseUrlQuery:url];
         
-        // osmandmaps://?lat=45.6313&lon=34.9955&z=8&title=New+York
-        double lat = [params[@"lat"] doubleValue];
-        double lon = [params[@"lon"] doubleValue];
-        double zoom = [params[@"z"] doubleValue];
-        NSString *title = params[@"title"];
+		        // osmandmaps://?lat=45.6313&lon=34.9955&z=8&title=New+York
+		        double lat = [params[@"lat"] doubleValue];
+		        double lon = [params[@"lon"] doubleValue];
+		        double zoom = [params[@"z"] doubleValue];
+		        NSString *title = params[@"title"];
+		        NSString *navigate = [url host];
         
-        Point31 pos31 = [OANativeUtilities convertFromPointI:OsmAnd::Utilities::convertLatLonTo31(OsmAnd::LatLon(lat, lon))];
-        dispatch_async(dispatch_get_main_queue(), ^{
+		        Point31 pos31 = [OANativeUtilities convertFromPointI:OsmAnd::Utilities::convertLatLonTo31(OsmAnd::LatLon(lat, lon))];
+		        dispatch_async(dispatch_get_main_queue(), ^{
             
-            OAMapViewController* mapViewController = [_rootViewController.mapPanel mapViewController];
+		            OAMapViewController* mapViewController = [_rootViewController.mapPanel mapViewController];
             
-            UIViewController *top = _rootViewController.navigationController.topViewController;
+		            UIViewController *top = _rootViewController.navigationController.topViewController;
             
-            if (![top isKindOfClass:[JASidePanelController class]])
-                [_rootViewController.navigationController popToRootViewControllerAnimated:NO];
+		            if (![top isKindOfClass:[JASidePanelController class]])
+		                [_rootViewController.navigationController popToRootViewControllerAnimated:NO];
 
-            if (_rootViewController.state != JASidePanelCenterVisible)
-                [_rootViewController showCenterPanelAnimated:NO];
+		            if (_rootViewController.state != JASidePanelCenterVisible)
+		                [_rootViewController showCenterPanelAnimated:NO];
 
-            [_rootViewController.mapPanel closeDashboard];
+		            [_rootViewController.mapPanel closeDashboard];
             
-            [mapViewController goToPosition:pos31 andZoom:zoom animated:NO];
-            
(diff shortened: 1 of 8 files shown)


vshcherb:
Is it possible to display settings only if WunderLink connected?

kconger:
That should be possible, although it may mean giving the app Bluetooth permissions.

kconger:
> Is it possible to display settings only if WunderLink connected?

I took a look and I would need to add Bluetooth permissions to OsmAnd-ios,  if you're ok with this I will add the code to only display the setting when connected.

Thanks,
Keith

vshcherb:
I was thinking about it and I've got a feeling that a direction is almost good. I believe that WunderLinq is an external keyboard and in that sense we need to support "extra keys". Outcome
1) It would make sense to have a setting called "External input devices" (and in description mention bluetooth keyboards and WunderLinq). The setting should be not a checkbox but a List as of today it could be [None, WunderLinq].
2) I think it would make sense to backport to Android in similar layout as well.

So, in future it could be easy to connect a normal keyboard or whatever keyboard, obviously it won't redirect to "wunderlink:" but it can similar useful things like zoom in / out or even more. 

P.S. do you have another keys supported by this device?

kconger:
> I was thinking about it and I've got a feeling that a direction is almost good. I believe that WunderLinq is an external keyboard and in that sense, we need to support "extra keys". Outcome
> 
> 1. It would make sense to have a setting called "External input devices" (and in description mention bluetooth keyboards and WunderLinq). The setting should be not a checkbox but a List as of today it could be [None, WunderLinq].
> 2. I think it would make sense to backport to Android in similar layout as well.
> 
> So, in future it could be easy to connect a normal keyboard or whatever keyboard, obviously it won't redirect to "wunderlink:" but it can similar useful things like zoom in / out or even more.
> 
> P.S. do you have another keys supported by this device?

Hi,

I think that sounds like an excellent idea.  Currently, our device supports: UP, DOWN, LEFT, RIGHT, Escape and Enter.  

I'll start working in this direction and will reply back once I have something for you to review.

Keith

vshcherb:
👍

kconger:
> I was thinking about it and I've got a feeling that a direction is almost good. I believe that WunderLinq is an external keyboard and in that sense we need to support "extra keys". Outcome
> 
> 1. It would make sense to have a setting called "External input devices" (and in description mention bluetooth keyboards and WunderLinq). The setting should be not a checkbox but a List as of today it could be [None, WunderLinq].
> 2. I think it would make sense to backport to Android in similar layout as well.
> 
> So, in future it could be easy to connect a normal keyboard or whatever keyboard, obviously it won't redirect to "wunderlink:" but it can similar useful things like zoom in / out or even more.
> 
> P.S. do you have another keys supported by this device?

Here is my first pass at your suggestion let me know what you think.  If this is good all update the Android app.

vshcherb:
Idea and settings layout looks good

Sources/AppHost/OAAppDelegate.mm
```diff
@@ -189,37 +189,44 @@ - (BOOL) application:(UIApplication *)application openURL:(NSURL *)url sourceApp
...
+                    [_rootViewController showCenterPanelAnimated:NO];
+                
+                [_rootViewController.mapPanel closeDashboard];
+                
+                [mapViewController goToPosition:pos31 andZoom:zoom animated:NO];
+                OATargetPoint *targetPoint = [mapViewController.mapLayers.contextMenuLayer getUnknownTargetPoint:lat longitude:lon];
+                if (title.length > 0)
+                    targetPoint.title = title;
+                if ([navigate  isEqual: @"navigate"]){
+                    [[OARootViewController instance].mapPanel navigate:targetPoint];
```
crimean: You have _rootViewController

Sources/AppHost/OAAppDelegate.mm
```diff
@@ -189,37 +189,44 @@ - (BOOL) application:(UIApplication *)application openURL:(NSURL *)url sourceApp
...
+                if (_rootViewController.state != JASidePanelCenterVisible)
+                    [_rootViewController showCenterPanelAnimated:NO];
+                
+                [_rootViewController.mapPanel closeDashboard];
+                
+                [mapViewController goToPosition:pos31 andZoom:zoom animated:NO];
+                OATargetPoint *targetPoint = [mapViewController.mapLayers.contextMenuLayer getUnknownTargetPoint:lat longitude:lon];
+                if (title.length > 0)
+                    targetPoint.title = title;
+                if ([navigate  isEqual: @"navigate"]){
```
crimean: What if we already navigating? Did you test it?

Sources/Controllers/OARootViewController.m
```diff
@@ -130,6 +131,8 @@ - (void) viewDidLoad
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(productsRestored:) name:OAIAPProductsRestoredNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(requestPurchase:) name:OAIAPRequestPurchaseProductNotification object:nil];
     [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(reachabilityChanged:) name:kReachabilityChangedNotification object:nil];
+    
+    _mapViewController = [OARootViewController instance].mapPanel.mapViewController;
```
crimean: Calling [OARootViewController instance] at OARootViewController...
I would suggest to use mapPanel.mapViewController instead of class field.

Sources/Controllers/OARootViewController.m
```diff
@@ -696,4 +699,43 @@ - (void) popoverControllerDidDismissPopover:(UIPopoverController *)popoverContro
...
+        NSString *wunderlinqAppURL = @"wunderlinq://datagrid";
+        BOOL canOpenURL = [[UIApplication sharedApplication]
+                       canOpenURL:[NSURL URLWithString:wunderlinqAppURL]];
+        if ( canOpenURL ) [[UIApplication sharedApplication]
+                       openURL:[NSURL URLWithString:wunderlinqAppURL] options:@{} completionHandler:nil];
+    }
+}
+
+- (void)chooseSelection {
+    printf("Enter\n");
```
crimean: Is it needed?

kconger: Sorry, that was for future use.  I'll remove.

Sources/AppHost/OAAppDelegate.mm
kconger: Yes, so the app pops up a menu to add as an additional location or new route.  I think thats what we would want?

Sources/AppHost/OAAppDelegate.mm
kconger: Fixed

Sources/Controllers/OARootViewController.m
kconger: Not sure how to fix this one.

crimean: Just drop it and use mapPanel.mapViewController where you need

Sources/AppHost/OAAppDelegate.mm
crimean: yes, ok

vshcherb:
@kconger can you please recreate these changes to OsmAnd android as well?

crimean:
@kconger BTW - do you need testflight version for tests? If yes - give me your email please.

kconger:
> @kconger can you please recreate these changes to OsmAnd android as well?

Already on it :) I'm going on vacation so it may be a little bit before I get it completed.

> @kconger BTW - do you need testflight version for tests? If yes - give me your email please.

Would love to, my email is keith.conger@blackboxembedded.com

PR #360, opened 2019-07-17
Added obf file import + added purchase screens to regional banners

PaulStets:


Sources/Map/OAMapImportHelper.mm
```diff
@@ -0,0 +1,97 @@
...
+    {
+        _app = [OsmAndApp instance];
+        _fileManager = [NSFileManager defaultManager];
+        NSArray *paths = [_fileManager URLsForDirectory:NSDocumentDirectory inDomains:NSUserDomainMask];
+        NSURL *documentsURL = [paths lastObject];
+        _documentsDir = documentsURL.path;
+        
+        BOOL isDir = YES;
+        if (![_fileManager fileExistsAtPath:_documentsDir isDirectory:&isDir])
+            [_fileManager createDirectoryAtPath:_documentsDir withIntermediateDirectories:YES attributes:nil error:nil];
```
crimean: It is possible that doc dir non exists?

PR #500, opened 2019-11-26
Hide menu buttons when quick action menu is on

iyerin:


Sources/Controllers/2.0/OAQuickActionHudViewController.mm
```diff
@@ -209,6 +208,7 @@ - (void)hideActionsSheetAnimated
     } completion:^(BOOL finished) {
         [_actionsView removeFromSuperview];
     }];
+    [_mapHudController showTopControls];
```
max-klaus: Why viewWillDisappear or viewDidDisappear is not suitable?

Sources/Controllers/2.0/OAQuickActionHudViewController.mm
```diff
@@ -233,6 +233,7 @@ - (IBAction)quickActionButtonPressed:(id)sender
             [self adjustMapViewPort];
         }];
         [self setPinPosition];
+        [_mapHudController hideTopControls];
```
max-klaus: Why viewWillAppear or viewDidAppear is not suitable?

PR #509, opened 2019-12-06
Gpx issues fixed 4/5

iyerin:


Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -276,9 +276,10 @@ - (void) handleKmlImport:(NSData *)data
     }
 }
 
-- (void) processUrl:(NSURL *)url showAlwerts:(BOOL)showAlerts
+- (void) processUrl:(NSURL *)url showAlwerts:(BOOL)showAlerts showDownloadedGPX:(BOOL)showDownloadedGPX
```
max-klaus: showAlwerts > showAlerts
showDownloadedGPX > openGpxView

Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -326,6 +327,11 @@ - (void) processUrl:(NSURL *)url showAlwerts:(BOOL)showAlerts
                    otherButtonTitles:nil];
         }
     }
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [self doPush];
+        [[OARootViewController instance].mapPanel openTargetViewWithGPX:item pushed:YES];
```
max-klaus: what if item = nil?

Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -374,6 +379,7 @@ -(void)doImport:(BOOL)doRefresh
         [self generateData];
         [self setupView];
     }
+    return(item);
```
max-klaus: () ?

PR #525, opened 2019-12-27
Rtl refactoring

iyerin:


Sources/Controllers/2.0/OATripRecordingSettingsViewController.mm
```diff
@@ -129,13 +129,10 @@ -(UIView *) getTopView
...
 
--(UIView *) getMiddleView
-{
-    return _tableView;
-}
-
 - (void) setupView
 {
+    if ([self.backButton isDirectionRTL])
+        self.backButton.transform = CGAffineTransformMakeRotation(M_PI);
```
max-klaus: Why transform?

Sources/Controllers/2.0/Routing/OARouteInfoView.mm
```diff
@@ -1144,7 +1146,9 @@ - (UITableViewCell *) tableView:(UITableView *)tableView cellForRowAtIndexPath:(
         {
             cell.backgroundColor = UIColor.whiteColor;
             cell.dividerColor = UIColorFromRGB(color_divider_blur);
-            cell.dividerInsets = [item[@"custom_insets"] boolValue] ? UIEdgeInsetsMake(0., 62., 0., 0.) : UIEdgeInsetsZero;
+            CGFloat leftInset = DirectionIsRTL ? 0. : 62.0;
```
max-klaus: DirectionIsRTL - drop it

Sources/Controllers/2.0/Routing/OARouteInfoView.mm
```diff
@@ -1305,16 +1311,21 @@ - (CGFloat)tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexPa
...
-        return [OADescrTitleIconCell getHeight:item[@"title"] value:item[@"descr"] cellWidth:tableView.bounds.size.width];
+        return UITableViewAutomaticDimension;
     else if ([item[@"cell"] isEqualToString:@"OARouteProgressBarCell"])
         return 2.0;
     return 44.0;
 }
 
+- (CGFloat)tableView:(UITableView *)tableView estimatedHeightForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    return kEstimatedRowHeight;
```
max-klaus: Do not override with single return kEstimatedRowHeight

Sources/Controllers/2.0/Routing/OARouteTripSettingsViewController.mm
```diff
@@ -129,14 +129,19 @@ - (CGFloat) heightForRow:(NSIndexPath *)indexPath tableView:(UITableView *)table
...
     }
     else
     {
         return 48.0;
     }
 }
 
+- (CGFloat)tableView:(UITableView *)tableView estimatedHeightForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    return kEstimatedRowHeight;
```
max-klaus: remove

Sources/Controllers/Resources/OADonationSettingsViewController.mm
```diff
@@ -591,7 +591,7 @@ - (CGFloat) tableView:(UITableView *)tableView heightForRowAtIndexPath:(NSIndexP
     }
     else if ([type isEqualToString:kCellTypeTextInput])
     {
-        return [OATextInputCell getHeight:item[@"title"] desc:nil cellWidth:1000.0];
+        return 44.0;
```
max-klaus: why constant height?

Sources/Helpers/OAAppSettings.m
```diff
@@ -1300,9 +1300,9 @@ - (instancetype) init
         [_autoZoomMap setModeDefaultValue:@NO mode:[OAApplicationMode PEDESTRIAN]];
         
         _mapSettingSaveTrackInterval = [OAProfileInteger withKey:mapSettingSaveTrackIntervalKey defValue:SAVE_TRACK_INTERVAL_DEFAULT];
-        [_mapSettingSaveTrackInterval setModeDefaultValue:@3000 mode:[OAApplicationMode CAR]];
-        [_mapSettingSaveTrackInterval setModeDefaultValue:@5000 mode:[OAApplicationMode BICYCLE]];
-        [_mapSettingSaveTrackInterval setModeDefaultValue:@10000 mode:[OAApplicationMode PEDESTRIAN]];
+        [_mapSettingSaveTrackInterval setModeDefaultValue:@3 mode:[OAApplicationMode CAR]];
```
max-klaus: Is it related to RTL? Why SAVE_TRACK_INTERVAL_DEFAULT left 5000? What will be with old users and wrong values?

Sources/Controllers/2.0/OAOsmEditingSettingsViewController.mm
```diff
@@ -98,6 +98,7 @@ - (OAButtonCell *) getButtonCell
     {
         [cell.button setTitle:OALocalizedString(@"shared_string_edit") forState:UIControlStateNormal];
         [cell.button addTarget:self action:@selector(editPressed) forControlEvents:UIControlEventTouchDown];
+        [cell showImage:NO];
```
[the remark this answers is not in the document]
iyerin: showImage is called in getButtonCell. getButtonCell is called in viewDidLoad. So it is called once. I dont understand what should I fix.

Sources/Helpers/OAAppSettings.m
iyerin: Not related with RTL. Was noticed during RTL fixing, that default value in cell is set to "50 minutes". Tried to investigate. I think, problem is happened when figures were copied from Android, where time is measured in milliseconds.
SAVE_TRACK_INTERVAL_DEFAULT left 5000, because I didnt know about its existence. I think, it also should be changed to 5. 
What will be with old users and wrong values? - Cant answer this question.

Sources/Controllers/2.0/OATripRecordingSettingsViewController.mm
iyerin: because "self.backButton.imageView.image = self.backButton.imageView.image.imageFlippedForRightToLeftLayoutDirection" is rewriting image, and when we go to inner settings and go back, it flips image 1 more time.

PR #663, opened 2020-04-09
UI fix in Overlay/Underlay

AnnBibyk:


Sources/Controllers/2.0/OAMapSettingsOverlayUnderlayScreen.mm
```diff
@@ -476,11 +493,12 @@ - (UITableViewCell*) tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
         
         if (cell)
         {
-            cell.titleLabel.text = item[@"title"];
+            cell.titleLabel.text = _mapSettingType == EMapSettingOverlay ? item[@"title"] : OALocalizedString(@"map_settings_base_transp");
```
max-klaus, line 496: Wrong place for the fix. Must be fixed at item[@"title"]

Sources/Controllers/2.0/OAMapSettingsOverlayUnderlayScreen.mm
```diff
@@ -181,7 +181,7 @@ - (void)setupView
         }
     }
     
-    
+    [tblView reloadData];
```
max-klaus, line 184: Looks wrong

PR #682, opened 2020-05-07
Selection of cell in editing mode changed.

AnnBibyk:
+ Extra space at top of the table removed

### DIFF
--- a/Sources/Controllers/2.0/OAFavoriteListViewController.mm
+++ b/Sources/Controllers/2.0/OAFavoriteListViewController.mm
@@ -86,7 +86,7 @@ typedef enum
     OAEditGroupViewController *_groupController;
 
     CALayer *_horizontalLine;
-    NSMutableArray *_selectedUnsortedItems;
+    NSMutableArray<NSIndexPath *> *_selectedItems;
 }
 
 static UIViewController *parentController;
@@ -129,7 +129,7 @@ static UIViewController *parentController;
     self.editToolbarView.backgroundColor = UIColorFromRGB(kBottomToolbarBackgroundColor);
     [self.editToolbarView.layer addSublayer:_horizontalLine];
     
-    _selectedUnsortedItems = [[NSMutableArray alloc] init];
+    _selectedItems = [[NSMutableArray alloc] init];
 }
 
 -(void)viewWillLayoutSubviews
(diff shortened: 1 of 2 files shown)


Sources/Controllers/2.0/OAFavoriteListViewController.mm
```diff
@@ -86,6 +86,7 @@ @implementation OAFavoriteListViewController
     OAEditGroupViewController *_groupController;
 
     CALayer *_horizontalLine;
+    NSMutableArray *_selectedUnsortedItems;
```
max-klaus, line 89: Type

Sources/Controllers/2.0/OAFavoriteListViewController.mm
```diff
@@ -600,13 +603,14 @@ - (void)setupColor
         [app saveFavoritesToPermamentStorage];
     }
     
+    [_selectedUnsortedItems removeAllObjects];
     [self editButtonClicked:nil];
```
max-klaus, line 607: Needs refactoring. We need two new methods - startEditing / finishEditing. These methods should be called from editButtonClicked. [_selectedUnsortedItems removeAllObjects] should be in finishEditing. Call finishEditing instead of [self editButtonClicked:nil].

Sources/Controllers/2.0/OAFavoriteListViewController.mm
```diff
@@ -1157,21 +1168,27 @@ - (void)removeItemsFromUnsortedFavoriteItems:(NSArray *)selectedRows
...
-            [self.favoriteTableView beginUpdates];
             app.favoritesCollection->removeFavoriteLocation(item.favorite);
             [groupData.groupItems removeObjectAtIndex:dataIndex];
-            [self.favoriteTableView deleteRowsAtIndexPaths:[NSArray arrayWithObject:[NSIndexPath indexPathForRow:selectedItem.row inSection:selectedItem.section]] withRowAnimation:UITableViewRowAnimationLeft];
-            [self.favoriteTableView endUpdates];
+            
+            if (groupData.isOpen)
+            {
+                [self.favoriteTableView beginUpdates];
+
```
max-klaus, line 1177: extra line?

Sources/Controllers/2.0/OAFavoriteListViewController.mm
```diff
@@ -1157,21 +1168,27 @@ - (void)removeItemsFromUnsortedFavoriteItems:(NSArray *)selectedRows
...
+            }
             [app saveFavoritesToPermamentStorage];
         }
     }
+    [_selectedUnsortedItems removeAllObjects];
     [self editButtonClicked:nil];
 }
 
-- (void)removeFavoriteItems:(NSArray *)selectedRows
+- (void) removeFavoriteItems:(NSArray *)selectedRows
```
max-klaus, line 1188: Looks like half of work done. I think we need handle selection via array for both modes sorted / unsorted: _selectedItems

PR #699, opened 2020-05-20
bottom toolbar height fix

AnnBibyk:
+ destination line starting point moved


Sources/Controllers/Map/Widgets/OADestinationsLineWidget.mm
```diff
@@ -122,6 +122,7 @@ - (void) layoutSubviews
     [super layoutSubviews];
    
     self.frame = CGRectMake(0., 0., DeviceScreenWidth, DeviceScreenHeight);
+    _destinationLineSublayer.frame = CGRectMake(0., 0., DeviceScreenWidth, DeviceScreenHeight);
```
max-klaus, line 125: _destinationLineSublayer.frame = self.bounds?

PR #775, opened 2020-07-24
Vehicle parameters

AnnBibyk:


Sources/Controllers/2.0/Settings/NavigationSettings/OAVehicleParametersSettingsViewController.mm
```diff
@@ -0,0 +1,371 @@
...
+- (NSInteger) tableView:(nonnull UITableView *)tableView numberOfRowsInSection:(NSInteger)section {
+    return _data[section].count;
+}
+
+- (NSInteger) numberOfSectionsInTableView:(UITableView *)tableView
+{
+    return 2;
+}
+
+- (CGFloat) heightForLabel:(NSString *)text
```
max-klaus, line 291: 1. Similar method should be in OAUtilities.
2. Why UILabel is created at all? It is possible to calculate height with calculateTextBounds.

PR #789, opened 2020-08-04
Menu elements

AnnBibyk:


Sources/Controllers/2.0/Settings/GeneralSettings/OAProfileGeneralSettingsViewController.m
```diff
@@ -308,7 +308,7 @@ - (void) setupView
     [tableData addObject:otherArr];
     _data = [NSArray arrayWithArray:tableData];
     [self updateNavBar];
-    [self.tableView reloadData];
+    //[self.tableView reloadData];
```
PaulStets: Always remove comments like this.

PR #808, opened 2020-08-20
Settings: General Settings for profiles 30

AnnBibyk:


Sources/Controllers/2.0/OACoordinatesFormatViewController.m
```diff
@@ -44,7 +44,7 @@ - (instancetype) initWithMode:(OAApplicationMode *)applicationMode
 - (void) applyLocalization
 {
     self.titleLabel.text = OALocalizedString(@"coords_format");
-    self.subtitleLabel.text = OALocalizedString(@"app_mode_car");
+    self.subtitleLabel.text = self.appMode.toHumanString;
```
PaulStets, line 47: This should be achieved by calling ```[super applyLocalization]``` rather than by copy-pasting. (See all other classes).

PR #842, opened 2020-09-15
Download map source

AnnBibyk:


Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -1639,6 +1640,16 @@ - (void) refreshMap
         [[OAMapViewTrackingUtilities instance] refreshLocation];
 }
 
+- (void) disableRotationAnd3DView:(BOOL)disabled
+{
+    _rotationAnd3DViewDisabled = disabled;
+}
+
+- (void) restoreViewAngle
```
max-klaus, line 1648: restoreViewAngle > resetViewAngle

PR #977, opened 2020-12-12
Lost dialog before track recording

AnnBibyk:


Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -1166,7 +1166,7 @@ - (void)startStopRecPressed
     }
     else
     {
-        if (!settings.mapSettingSaveTrackIntervalApproved && ![_savingHelper hasData])
+        if (![settings.mapSettingSaveTrackIntervalApproved get:settings.applicationMode] && ![_savingHelper hasData])
```
PaulStets, line 1169: If we do not need a custom AppMode we use the ```get``` method.

PR #1025, opened 2021-01-16
Add more keyboard shortcuts

Fossil01:
## Why?:
Lots of apps seem to implement their own shortcuts which creates confusion for the end-user. Zooming on an Apple Mac or iOS app has always used `CMD +` and `CMD -`. Resetting the zoom level normally is `CMD 0`.

So to be more in line with Apple's default shortcuts from apps like Apple Maps which use `CMD +` and `CMD -`. I have added these as defaults.

A dedicated re-center button (`CMD 0`) is useful when the map was moved or zoomed, to quickly get back to the current location.

## Added:

`CMD +` for Zoom In (With a hidden alternative of CMD = to also make it work on non numeric keyboards without much confusion.
`CMD -` for Zoom Out
`CMD 0` for recentering the map.

## Changed:
The **original zoom in and zoom out** on the arrow keys are still active but not shown in the help popup when holding down CMD for backwards compatibility. We might want to re-use these keys to move the map around sometime in the future.

## Notes:
Someone else added the check for an external input device a while back. I left it in there because am not sure if this is needed. But there is no way to use keyboard shortcuts in iOS without a hardware keyboard so IMHO that could maybe be removed.

### DIFF
--- a/Sources/Controllers/OARootViewController.m
+++ b/Sources/Controllers/OARootViewController.m
@@ -831,7 +831,7 @@ typedef enum : NSUInteger {
              [UIKeyCommand keyCommandWithInput:@"+" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn) discoverabilityTitle:OALocalizedString(@"key_hint_zoom_in")],
              [UIKeyCommand keyCommandWithInput:@"=" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn)],
              [UIKeyCommand keyCommandWithInput:UIKeyInputEscape modifierFlags:0 action:@selector(goBack) discoverabilityTitle:OALocalizedString(@"key_hint_goback")],
-             [UIKeyCommand keyCommandWithInput:@"0" modifierFlags:UIKeyModifierCommand action:@selector(recenterMap) discoverabilityTitle:OALocalizedString(@"sett_arr_map")]];
+             [UIKeyCommand keyCommandWithInput:@"0" modifierFlags:UIKeyModifierCommand action:@selector(recenterMap) discoverabilityTitle:OALocalizedString(@"key_hint_recenter_map")]];
 }
 
 - (void) zoomOut
--- a/Sources/Controllers/OARootViewController.m
+++ b/Sources/Controllers/OARootViewController.m
@@ -831,7 +831,7 @@ typedef enum : NSUInteger {
              [UIKeyCommand keyCommandWithInput:@"+" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn) discoverabilityTitle:OALocalizedString(@"key_hint_zoom_in")],
              [UIKeyCommand keyCommandWithInput:@"=" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn)],
              [UIKeyCommand keyCommandWithInput:UIKeyInputEscape modifierFlags:0 action:@selector(goBack) discoverabilityTitle:OALocalizedString(@"key_hint_goback")],
-             [UIKeyCommand keyCommandWithInput:@"0" modifierFlags:UIKeyModifierCommand action:@selector(recenterMap) discoverabilityTitle:@"Recenter the map"]];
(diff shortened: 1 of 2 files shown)


Sources/Controllers/OARootViewController.m
```diff
@@ -824,9 +825,13 @@ - (BOOL) canBecomeFirstResponder
...
-             [UIKeyCommand keyCommandWithInput:UIKeyInputEscape modifierFlags:0 action:@selector(goBack) discoverabilityTitle:OALocalizedString(@"key_hint_goback")]];
+    return @[[UIKeyCommand keyCommandWithInput:UIKeyInputDownArrow modifierFlags:0 action:@selector(zoomOut)],
+             [UIKeyCommand keyCommandWithInput:@"-" modifierFlags:UIKeyModifierCommand action:@selector(zoomOut) discoverabilityTitle:OALocalizedString(@"key_hint_zoom_out")],
+             [UIKeyCommand keyCommandWithInput:UIKeyInputUpArrow modifierFlags:0 action:@selector(zoomIn)],
+             [UIKeyCommand keyCommandWithInput:@"+" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn) discoverabilityTitle:OALocalizedString(@"key_hint_zoom_in")],
+             [UIKeyCommand keyCommandWithInput:@"=" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn)],
+             [UIKeyCommand keyCommandWithInput:UIKeyInputEscape modifierFlags:0 action:@selector(goBack) discoverabilityTitle:OALocalizedString(@"key_hint_goback")],
+             [UIKeyCommand keyCommandWithInput:@"0" modifierFlags:UIKeyModifierCommand action:@selector(recenterMap) discoverabilityTitle:@"Recenter the map"]];
```
PaulStets, line 834: Please localize the ```discoverabilityTitle``` here.

Fossil01: Done

Sources/Controllers/OARootViewController.m
```diff
@@ -824,9 +825,13 @@ - (BOOL) canBecomeFirstResponder
...
-             [UIKeyCommand keyCommandWithInput:UIKeyInputEscape modifierFlags:0 action:@selector(goBack) discoverabilityTitle:OALocalizedString(@"key_hint_goback")]];
+    return @[[UIKeyCommand keyCommandWithInput:UIKeyInputDownArrow modifierFlags:0 action:@selector(zoomOut)],
+             [UIKeyCommand keyCommandWithInput:@"-" modifierFlags:UIKeyModifierCommand action:@selector(zoomOut) discoverabilityTitle:OALocalizedString(@"key_hint_zoom_out")],
+             [UIKeyCommand keyCommandWithInput:UIKeyInputUpArrow modifierFlags:0 action:@selector(zoomIn)],
+             [UIKeyCommand keyCommandWithInput:@"+" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn) discoverabilityTitle:OALocalizedString(@"key_hint_zoom_in")],
+             [UIKeyCommand keyCommandWithInput:@"=" modifierFlags:UIKeyModifierCommand action:@selector(zoomIn)],
+             [UIKeyCommand keyCommandWithInput:UIKeyInputEscape modifierFlags:0 action:@selector(goBack) discoverabilityTitle:OALocalizedString(@"key_hint_goback")],
+             [UIKeyCommand keyCommandWithInput:@"0" modifierFlags:UIKeyModifierCommand action:@selector(recenterMap) discoverabilityTitle:OALocalizedString(@"sett_arr_map")]];
```
PaulStets, line 834: It would be better to create a new string with a key that resembles others e.g. ```"key_hint_recenter_map"``` and a corresponding text

Fossil01: But then it would have to be translated in a million languages before that works? I don't speak all of those :-) Or does it fallback to English?

PaulStets: You need to add the string only to the Localizable.strings (English). It will default to English for others until our Weblate contributors translate it.

PaulStets:
Looks good, thank you for the request:)

PR #1058, opened 2021-01-28
Track groups

AnnBibyk:


Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -536,27 +534,27 @@ -(void)viewWillAppear:(BOOL)animated {
     [self applySafeAreaMargins];
 }
 
--(UIView *) getTopView
+- (UIView *) getTopView
```
PaulStets, line 537: If you are using constraints (which you should), please remove this and three methods below that were used before the introduction of autolayout and are completely useless now.

Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -594,11 +592,11 @@ - (void)onTrackRecordingChanged
     });
 }
 
-- (void)onGpxRouteCanceled
+- (void) onGpxRouteCanceled
 {
     dispatch_async(dispatch_get_main_queue(), ^{
         
-        if (_viewMode == kActiveTripsMode && _routeSectionIndex > -1)
+        if (_routeSectionIndex > -1)
```
PaulStets, line 599: _routeSectionIndex is used here but never assigned?

Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -671,33 +656,89 @@ -(void)generateData
...
+        ];
+    
     if (self.gpxList.count > 0)
     {
         // Sort items by date-time added desc
         NSArray *sortedArrayGroups = [self.gpxList sortedArrayUsingComparator:^NSComparisonResult(OAGPX* obj1, OAGPX* obj2) {
             return [obj2.importDate compare:obj1.importDate];
         }];
-        [self.gpxList setArray:sortedArrayGroups];
+        [_gpxList setArray:sortedArrayGroups];
```
PaulStets, line 681: The reason behind this change?

Sources/Controllers/Cells/OAIconTitleValueCell.h
```diff
@@ -16,6 +16,7 @@
 @property (weak, nonatomic) IBOutlet UIImageView *leftImageView;
 @property (strong, nonatomic) IBOutlet NSLayoutConstraint *noImageTextLeadingMargin;
 @property (strong, nonatomic) IBOutlet NSLayoutConstraint *imageTextLeadingMargin;
+@property (strong, nonatomic) IBOutlet UIButton *openCloseGroupButton;
```
PaulStets, line 19: Will this break other places where this cell is used?

Sources/Controllers/2.0/OAGPXListViewController.mm
PaulStets: Maybe it would be a great idea not to use constants for sections at all since we will have a dynamic number of folders in the future.

Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -894,249 +811,44 @@ - (void)onImportClicked
...
-        }
-        
-        if (cell) {
-            [cell.textView setText:OALocalizedString(@"create_new_trip")];
-            [cell.iconView setImage: [UIImage imageNamed:@"icon_info"]];
-        }
-        return cell;
-    }
-    else
+    NSString *fileName = [NSString stringWithFormat:@"Trip_%@.gpx", [fmt stringFromDate:[NSDate date]]];
```
PaulStets, line 821: Since you're changing Trip to track it would be a great idea to replace all occurrences of "trip" throughout the entire project to avoid confusion in the UI.

Sources/Controllers/2.0/OAGPXListViewController.mm
```diff
@@ -1283,90 +994,512 @@ - (void)doSaveTrack
...
+        if (groupData.isOpen)
+            return [groupData.groupItems count] + 1;
+        return 1;
+    }
+    return [[_data objectAtIndex:section] count];
+}
+
+- (UITableViewCell *) tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath
+{
+    id sectionObject = [_data objectAtIndex:indexPath.section];
```
PaulStets, line 1061: Please refactor this entirely. Create an internal class(es) that will contain all the necessary section information and use them instead.

PR #1116, opened 2021-03-03
Distance by tap

AnnBibyk:


Sources/Controllers/Map/Layers/OARulerByTapControlLayer.mm
```diff
@@ -0,0 +1,387 @@
...
+}
+
+- (void) layoutSubviews
+{
+    // resize your layers based on the view's new bounds
+    [super layoutSubviews];
+    _fingerDistanceSublayer.frame = self.bounds;
+}
+
+- (void) drawRect:(CGRect)rect
```
PaulStets, line 160: This method declaration is redundant

Sources/Controllers/Map/Layers/OARulerByTapControlLayer.mm
```diff
@@ -0,0 +1,387 @@
...
+    }
+    CGContextRestoreGState(ctx);
+}
+
+- (void) touchDetected:(UITapGestureRecognizer *)recognizer
+{
+    // Handle gesture only when it is ended
+    if (recognizer.state != UIGestureRecognizerStateEnded)
+        return;
+    if (![_settings.showDistanceRuler get])
```
PaulStets, line 332: Perhaps implementing UIGestureRecognizer's shouldReceiveTouch would be better

Sources/Controllers/Map/Widgets/OARulerWidget.mm
```diff
@@ -208,7 +160,8 @@ - (void) commonInit
     _centerIconDay = [UIImage imageNamed:@"ic_ruler_center.png"];
     _centerIconNight = [UIImage imageNamed:@"ic_ruler_center_light.png"];
     _imageView.image = _settings.nightMode ? _centerIconNight : _centerIconDay;
-    self.hidden = YES;
+    self.hidden = NO;
```
PaulStets, line 163: Why did you change it to NO?

PR #1168, opened 2021-04-19
Delete custom categories

Skalii:


Sources/Controllers/2.0/OADeleteCustomFiltersTableViewController.mm
```diff
@@ -0,0 +1,226 @@
...
+        [self.tableView endUpdates];
+        [self.tableView reloadRowsAtIndexPaths:@[[NSIndexPath indexPathForRow:0 inSection:indexPath.section], indexPath] withRowAnimation:UITableViewRowAnimationNone];
+    }
+}
+
+#pragma mark - UITableViewDataSource
+
+- (nonnull UITableViewCell *)tableView:(nonnull UITableView *)tableView cellForRowAtIndexPath:(nonnull NSIndexPath *)indexPath
+{
+    NSString* identifierCell = indexPath.row == 0 ? kCellTypeSelectionButton : kCellTypeTitle;
```
PaulStets, line 125: This approach is incorrect. Instead of casting the cell, please check against the identifier and create an appropriate cell (see any other view controller with the tableview in it)

Sources/Controllers/2.0/OAQuickSearchViewController.mm
```diff
@@ -29,6 +29,7 @@
 #import "OACategoriesTableViewController.h"
 #import "OAHistoryTableViewController.h"
 #import "OACustomPOIViewController.h"
+#import "OACompoundViewController.h"
```
PaulStets, line 32: Seems like you do not need this import here

Sources/Controllers/2.0/OAQuickSearchViewController.mm
```diff
@@ -1824,13 +1826,25 @@ - (void) reloadAddressData
...
 {
     OAPOIUIFilter *filter = [[OAPOIFiltersHelper sharedInstance] getCustomPOIFilter];
     [filter clearFilter];
     OACustomPOIViewController *customPOI = [[OACustomPOIViewController alloc] initWithFilter:filter];
     customPOI.delegate = self;
-    [self.navigationController pushViewController:customPOI animated:YES];    
+    [self.navigationController pushViewController:customPOI animated:YES];
+}
+
+- (void)showRemoveFiltersView:(NSArray<OAPOIUIFilter *> *)filters
```
PaulStets, line 1838: You do not show a view (a UIView is just an element of a view controller like a button or a tableview) in this method, so this name is confusing. Please rename it to something like: ```showRemoveFiltersScreen```

Sources/Controllers/2.0/OADeleteCustomFiltersViewController.mm
```diff
@@ -0,0 +1,243 @@
...
+    NSMutableArray *_selectedItems;
+}
+
+-(instancetype)initWithFilters:(NSArray<OAPOIUIFilter *> *)filters
+{
+    self = [super init];
+    if (self)
+    {
+        _selectedItems = [NSMutableArray new];
+        _items = [[NSArray arrayWithArray:filters] mutableCopy];
```
PaulStets, line 41: Again, NSMutableArray arrayWithArray:

PR #1191, opened 2021-04-30
T10889

alisa911:


Sources/OAAbbreviations.mm
```diff
@@ -0,0 +1,61 @@
...
+//  Copyright © 2021 OsmAnd. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "OAAbbreviations.h"
+#import "OASearchPhrase.h"
+
+@implementation OAAbbreviations
+
+static NSMutableDictionary<NSString *, NSString *> *_ABBREVIATIONS;
```
PaulStets, line 15: This should be immutable

Sources/OAAbbreviations.mm
```diff
@@ -0,0 +1,61 @@
...
+
++ (void) initialize
+{
+    _ABBREVIATIONS = [NSMutableDictionary new];
```
PaulStets, line 19: You can create immutable dictionary inline @{key : value, key : value}

Sources/OAAbbreviations.h
```diff
@@ -0,0 +1,23 @@
...
+//  OAAbbreviations.h
+//  OsmAnd Maps
+//
+//  Created by plotva on 30.04.2021.
+//  Copyright © 2021 OsmAnd. All rights reserved.
+//
+//  OsmAnd-java/src/main/java/net/osmand/binary/Abbreviations.java
+//  git revision /54e26c5a6195beb371c210746c5d89674016b9f7
+
+#ifndef OAAbbreviations_h
```
PaulStets, line 11: No need for #ifndef and other related code in obj c header.

PR #1219, opened 2021-05-21
Refactoring global settings

Skalii:


Sources/Controllers/2.0/Settings/ImportExport/OAImportCompleteViewController.mm
```diff
@@ -325,6 +342,14 @@ -(void) loadCurrentRoutingMode
     [OARoutingHelper.sharedInstance setAppMode:currentRoutingMode];
 }
 
+- (NSString *)getTitleForSection
+{
+    NSString *headerText = OALocalizedString(@"import_complete_description");
+    if (_needRestart)
```
PaulStets, line 348: Are you sure the restart is needed in ios?

Sources/Controllers/2.0/Settings/ImportExport/OAExportItemsSelectionViewController.mm
```diff
@@ -33,9 +33,13 @@
 #import "OAGlobalSettingsItem.h"
 #import "OADestination.h"
 #import "OAPOIHelper.h"
+#import "OATableViewCustomHeaderView.h"
 
 #include <OsmAndCore/ArchiveReader.h>
 
+#define titleWithDescrCellHeight 60.0
```
PaulStets, line 40: Why is this code here? When you switch a task, please rebase onto master before writing the new code.

Sources/Controllers/2.0/Settings/ImportExport/OAImportCompleteViewController.mm
Skalii: yes, some settings are not applied without restart

Sources/Helpers/OAAppSettings.h
```diff
@@ -152,9 +152,94 @@ typedef NS_ENUM(NSInteger, EOAAutoZoomMap)
 
 @end
 
-@interface OAProfileSetting : NSObject
+typedef NS_ENUM(NSInteger, EOAWikiArticleShowConstant)
+{
+    ON = 0,
```
PaulStets, line 157: Since obj c enums do not have namespaces, it is generally a good idea to avoid such generic names for public enums. Prefer using ```EOAWikiArticleShowConstantOn``` format in such cases

Sources/Helpers/OAAppSettings.m
```diff
@@ -32,14 +32,15 @@
 #define metricSystemChangedManuallyKey @"metricSystemChangedManuallyKey"
 #define liveUpdatesPurchasedKey @"liveUpdatesPurchasedKey"
 #define settingOsmAndLiveEnabledKey @"settingOsmAndLiveEnabledKey"
+#define liveUpdatesRetryesKey @"liveUpdatesRetryesKey"
```
PaulStets, line 35: Did you mean retries?

Skalii: yes, but the setting storage key will remain "retryes", as it is written in android

PR #1275, opened 2021-06-11
Support generic GPX Uniform Type Identifier

xremix:
# Changes

Switching the import and export UTIs from `net.osmand.gpx` to the more generic `com.topografix.gpx`.

# Background Information

Other Apps could get into trouble when they register the generic UTI, but apple reserves `net.osmand.gpx` for `.gpx` files because OsmAnd is installed. I recognized this on my own App, which led to bad App Store reviews and other Apps that try to register for the generic UTI might run into the same issue. Please let me know if this PR makes sense and I also can extend this for the other file types that are registered in OsmAnd.

The problems that occured were, that the `UIDocumentPickerViewController` didn't recognize the `.gpx` files at all and the **Share Extension** recognized the files, but didn't read the contents of it. My work around was to support the generic UTI as well as the OsmAnd UTI.

### DIFF
--- a/OsmAnd Maps copy-Info.plist
+++ b/OsmAnd Maps copy-Info.plist
@@ -12,12 +12,12 @@
 			<key>CFBundleTypeIconFiles</key>
 			<array/>
 			<key>CFBundleTypeName</key>
-			<string>GPS eXchange format</string>
+			<string>GPS Exchange Format (GPX)</string>
 			<key>LSHandlerRank</key>
 			<string>Owner</string>
 			<key>LSItemContentTypes</key>
 			<array>
-				<string>net.osmand.gpx</string>
+				<string>com.topografix.gpx</string>
 			</array>
 		</dict>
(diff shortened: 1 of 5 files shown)


PaulStets:
Hi, @xremix, thank you for your contribution. I will test the functionality today. One minor piece of advice for the future: please do not commit auto-indentation edits as it makes it harder to review the pull request.

xremix:
@PaulStets thanks a lot for merging this in! And... whoops, sorry for the auto-indentations, won't do again! P.S.: Github has a great feature for that. In the `Files changed` section you can remove any whitespace changes from being displayed.

PaulStets:
@xremix Wow, never knew this feature existed:) Still, I try to avoid committing changes that do not contribute to the actual functionality of the pull request. Anyway, it's just a formatting preference, so it doesn't matter that much:) I will test the sharing functionality later today to make sure the import works as before.

PR #1297, opened 2021-06-24
Context menu add long press to copy and nearby poi

Skalii:


Sources/Controllers/2.0/OAOsmEditTargetViewController.mm
```diff
@@ -156,6 +156,11 @@ - (BOOL) showNearestWiki
     return NO;
 }
 
+- (BOOL) showNearestPoi
```
PaulStets, line 159: No need to override the default behavior (check in all places)

Sources/Controllers/2.0/OATargetInfoViewController.mm
```diff
@@ -687,7 +741,24 @@ - (void) addNearbyImagesIfNeeded
     
     _nearbyImagesRowInfo = nearbyImagesRowInfo;
 }
-	
+
+-(void)handleLongPressToCopyText:(UILongPressGestureRecognizer *)gestureRecognizer
```
PaulStets, line 745: Wouldn't it be much easier to handle this in ```didSelectCellForRow```?

PR #1448, opened 2021-08-19
Configure map - Routes

Skalii:


Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsMapTypeScreen.h
```diff
@@ -8,6 +8,17 @@
...
+@required
+
+- (void)updateSkimapRoutesParameter:(OAMapSource *)source;
+- (void)refreshMenuRoutesParameters;
+
+@end
+
 @interface OAMapSettingsMapTypeScreen : NSObject<OAMapSettingsScreen>
 
+@property (nonatomic) id<OAMapTypeDelegate> delegate;
```
PaulStets, line 22: Always make delegates weak!

Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsRoutesScreen.mm
```diff
@@ -0,0 +1,400 @@
...
+#import "OASettingSwitchCell.h"
+#import "OASettingsTitleTableViewCell.h"
+#import "Localization.h"
+#import "OAColors.h"
+#import "OAMapStyleSettings.h"
+
+static const NSInteger visibilitySection = 0;
+static const NSInteger colorsSection = 1;
+
+typedef enum
```
PaulStets, line 24: Use only ```typedef NS_ENUM()```

Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsRoutesScreen.mm
```diff
@@ -0,0 +1,400 @@
...
+
+- (void)initData
+{
+    NSMutableArray *dataArr = [@[
+            @[
+                    @{@"type": [OADividerCell getCellIdentifier]},
+                    @{@"type": [OASettingSwitchCell getCellIdentifier]},
+                    @{@"type": [OADividerCell getCellIdentifier]}
+            ]
+    ] mutableCopy];
```
PaulStets, line 96: Use ```addObject:``` instead of creating pointless copies.

Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsRoutesScreen.mm
```diff
@@ -0,0 +1,400 @@
...
+                        @"value": value.name,
+                        @"title": value.title
+                }];
+            }
+        }
+    }
+    [colorsArr addObject:@{@"type": [OADividerCell getCellIdentifier]}];
+    [dataArr addObject:colorsArr];
+
+    _data = [NSArray arrayWithArray:dataArr];
```
PaulStets, line 129: You can set NSMutableArray to an array here. It is completely safe as you lose the reference to the mutable array anyway.

PR #1468, opened 2021-08-27
Refactoring configure map

Skalii:


Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsMainScreen.mm
```diff
@@ -121,321 +361,26 @@ - (NSString *)getMapLangValueStr
...
     }
-    
+
     return languageValue;
 }
 
 - (NSArray *) getAllCategories
 {
-    NSMutableArray *res = [NSMutableArray array];
+    NSMutableArray *res = [NSMutableArray new];
```
PaulStets, line 374: Both variants of initialization are acceptable. No need to change one for the other. (I personally prefer using array/dictionary initialization though it's just a matter of personal preference)

Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsMainScreen.mm
```diff
@@ -458,8 +403,10 @@ - (NSString *) getPOIDescription
 
 - (CGFloat) heightForHeader:(NSInteger)section
 {
-    NSDictionary* data = (NSDictionary*)[((NSArray*)[((NSDictionary*)tableData[section]) objectForKey:@"cells"]) objectAtIndex:0];
-    if ([[data objectForKey:@"type"] isEqualToString:[OAAppModeCell getCellIdentifier]])
+    NSDictionary *sectionData = (NSDictionary *) tableData[section];
```
PaulStets, line 406: Absolutely no need for such casts

PR #1483, opened 2021-09-07
Fix population tag formatting

Skalii:


Sources/Controllers/TargetMenu/POI/OAPOIViewController.mm
```diff
@@ -218,6 +218,7 @@ - (void) buildRows:(NSMutableArray<OARowInfo *> *)rows
         BOOL isText = NO;
         BOOL isDescription = NO;
         BOOL needLinks = ![@"population" isEqualToString:key];
+        BOOL needIntFormatting = [@"population" isEqualToString:key];
```
PaulStets, line 221: Extract key to constants

Sources/Controllers/TargetMenu/POI/OAPOIViewController.mm
```diff
@@ -254,6 +256,18 @@ - (void) buildRows:(NSMutableArray<OARowInfo *> *)rows
                 skip = YES;
             }
         }
+        else
+        {
+            if (needIntFormatting)
```
PaulStets, line 261: Can be simplified to ```else if```

PR #1512, opened 2021-09-22
Track Menu - Tabs and Overview

Skalii:


Sources/Router/Colorization/OAColoringType.mm
```diff
@@ -160,6 +160,21 @@ - (BOOL) isGradient
     return self == self.class.SPEED || self == self.class.ALTITUDE || self == self.class.SLOPE;
 }
 
+- (BOOL) isSlope
```
PaulStets, line 163: This class was ported by me from Android, there are no such methods there, so they shouldn't be here either.

Sources/Controllers/OABaseScrollableHudViewController.mm
```diff
@@ -306,12 +313,21 @@ - (void) show:(BOOL)animated state:(EOADraggableMenuState)state onComplete:(void
...
+                frame.origin.x = -_scrollableView.bounds.size.width;
+                frame.origin.y = DeviceScreenHeight - self.additionalLandscapeOffset;
+            }
+            else
+            {
+                frame.origin.x = 0.0;
+                frame.origin.y = 0.0;
+            }
+
             _scrollableView.frame = frame;
```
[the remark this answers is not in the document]
PaulStets: The frame is still set two times

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -2778,18 +2778,15 @@ - (void)openTargetViewWithRouteDetailsGraph:(OAGPXDocument *)gpx
...
 
     [self enterContextMenuMode];
     [self showTargetPointMenu:NO showFullMenu:NO onComplete:^{
-        if (trackMenuDelegate)
-        {
-            OARouteDetailsGraphViewController *graphController = (OARouteDetailsGraphViewController *) self.targetMenuView.customController;
-            graphController.trackMenuDelegate = trackMenuDelegate;
-            [graphController onNewModeSelected:modeType];
-        }
+        if ([self.targetMenuView.customController isKindOfClass:OARouteDetailsGraphViewController.class])
```
PaulStets, line 2788: The problem was not in creating a variable, you should not cast anything here, the method you call should be in the customController protocol (better to make it optional and check for ```respondsToSelector```)

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -2335,7 +2335,11 @@ - (BOOL) findTrack:(CLLocationCoordinate2D)location
     auto activeGpx = _selectedGpxHelper.activeGpx;
     for (auto it = activeGpx.begin(); it != activeGpx.end(); ++it)
     {
-        const auto& doc = it.value();
+        NSString *gpxFilePath = [it.key().toNSString()
+                stringByReplacingOccurrencesOfString:[_app.gpxPath stringByAppendingString:@"/"]
+                                          withString:@""];
+        auto doc = std::const_pointer_cast<OsmAnd::GeoInfoDocument>(it.value());
+//        auto gpx = std::dynamic_pointer_cast<OsmAnd::GpxDocument>(doc);
```
PaulStets, line 2342: Куьщму сщььуте

PR #1516, opened 2021-09-26
Fix Wikipedia Layer part 2

Skalii:


Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsWikipediaLanguagesScreen.mm
```diff
@@ -97,6 +95,7 @@ - (void)viewDidLoad
     self.tableView.dataSource = self;
     self.tableView.editing = YES;
     self.tableView.tintColor = UIColorFromRGB(color_primary_purple);
+    self.tableView.separatorInset = UIEdgeInsetsMake(0., [OAUtilities getLeftMargin] + 52., 0., 0.);
```
PaulStets, line 98: This will probably be incorrect on device rotation. You need to call it again in ViewWillTransitionToSize

PR #1577, opened 2021-11-03
Fix tab bar in "My Places"

Skalii:


Sources/Controllers/MyPlaces/OAFavoritesTabBarViewController.m
```diff
@@ -22,6 +23,15 @@ - (void)viewDidLoad
         [self setViewControllers:newTabs];
     }
     [self applyLocalization];
+
+    if (@available(iOS 15.0, *))
```
PaulStets, line 27: This is not a correct fix for this issue. See https://developer.apple.com/forums/thread/682432

Sources/Controllers/MyPlaces/OAFavoritesTabBarViewController.m
```diff
@@ -9,6 +9,7 @@
 #import "OAFavoritesTabBarViewController.h"
 #import "OAIAPHelper.h"
 #import "Localization.h"
+#import "OAColors.h"
```
PaulStets, line 12: Remove this

PR #1603, opened 2021-11-13
Fix Track Menu - Options (part 2)

Skalii:


Sources/Controllers/RoutePlanning/OASaveTrackViewController.h
```diff
@@ -24,6 +27,10 @@
 
 @property (weak, nonatomic) id<OASaveTrackViewControllerDelegate> delegate;
 
-- (instancetype) initWithFileName:(NSString *)fileName filePath:(NSString *)filePath showOnMap:(BOOL)showOnMap simplifiedTrack:(BOOL)simplifiedTrack;
+- (instancetype) initWithFileName:(NSString *)fileName
```
PaulStets, line 30: Was this taken from the Android? If not, does it work in Android without this fix?

PR #1607, opened 2021-11-17
Fix Wikipedia Layer part 3

Skalii:


Sources/Plugins/OAPlugin.mm
```diff
@@ -385,7 +385,7 @@ - (void)prepareExtraTopPoiFilters:(NSSet<OAPOIUIFilter *> *)poiUIFilters
 
 + (void)onPrepareExtraTopPoiFilters:(NSSet<OAPOIUIFilter *> *)poiUIFilters
 {
-    for (OAPlugin *plugin in [self getEnabledPlugins])
+    for (OAPlugin *plugin in [self.class getEnabledPlugins])
```
PaulStets, line 388: No need for this, it is correct to call self when calling another class static methods

PR #1646, opened 2021-11-26
Fix Wikipedia Layer - sync code with Android

Skalii:


Sources/POI/OAPOIFiltersHelper.h
```diff
@@ -18,6 +18,7 @@
 
 - (OAPOIUIFilter *) getSearchByNamePOIFilter;
 - (OAPOIUIFilter *) getCustomPOIFilter;
++ (NSString*) getTopWikiPoiFilterId;
```
PaulStets, line 21: Static methods should be declared separately.

Sources/Plugins/OAPlugin.mm
```diff
@@ -264,35 +274,28 @@ - (void) clearContextMenuRows() {
...
-
     allPlugins.add(new AccessibilityPlugin(app));
-    allPlugins.add(new OsmEditingPlugin(app));
     allPlugins.add(new OsmandDevelopmentPlugin(app));
     */
 
-    [allPlugins addObject:[[OAParkingPositionPlugin alloc] init]];
+    [allPlugins addObject:[[OAWikipediaPlugin alloc] init]];
     [allPlugins addObject:[[OAMonitoringPlugin alloc] init]];
+    [allPlugins addObject:[[OASRTMPlugin alloc] init]];             //checkMarketPlugin
```
PaulStets, line 289: No point in these comments

PR #1679, opened 2021-12-10
Fix saving GPX file (4.1)

Skalii:


PaulStets:
I don't see the code that resets the appearance values in the database

PR #1681, opened 2021-12-10
Fix Favorites - access calendar after saving

Skalii:


Sources/Controllers/TargetMenu/PointEditing/OAFavoriteEditingHandler.mm
```diff
@@ -159,7 +159,6 @@ - (void)savePoint:(OAPointEditingData *)data newPoint:(BOOL)newPoint
     {
         [OAFavoritesHelper editFavoriteName:_favorite newName:data.name group:data.category descr:[_favorite getDescription] address:[_favorite getAddress]];
     }
-    [OAFavoritesHelper loadFavorites];
```
PaulStets, line 162: This is not the fix for this issue. You need t find out why the calendar is called in the first place and add a check to call it only if we import/export parking points

Sources/Helpers/OAFavoritesHelper.mm
```diff
@@ -694,6 +694,29 @@ + (void) removeParkingReminderFromCalendar
...
+        NSDate *startDate = event.startDate;
+        NSDate *endDate = event.endDate;
+        NSDate *pickupDate = [NSDate dateWithTimeIntervalSince1970:plugin.getParkingTime / 1000];
+        CLLocation *parkingPosition = [plugin getParkingPosition];
+        exists = parkingPoint
+                && [OAUtilities isCoordEqual:parkingPosition.coordinate.latitude
+                                      srcLon:parkingPosition.coordinate.longitude
+                                     destLat:[parkingPoint getLatitude]
+                                     destLon:[parkingPoint getLongitude]]
+                && [startDate isEqualToDate:pickupDate]
```
PaulStets, line 714: Both start and end date shoulf be equal to pickup date?

PR #1747, opened 2022-01-31
MGRS coordinates

jBackhaus:
Hey,
I added support for the MGRS coordinate system, which is commonly known as "NATO standard" but is also widely used for civil disaster management services.
Input and output were added to all places, also the possibility to choose it as the coordinate system for map profiles.

This resembles a pull request I added to the android version in 2020 (https://github.com/osmandapp/OsmAnd/pull/10107) and now I had the time to bring the same functionality and behaviour to iOS.

Please notice 2 things:
- I had to edit one function that is not directly connected to MGRS coordinates (getUTMCoordinateString in OALocationConvert.mm) to prevent crashes which are caused by latitude overflow when faulty MGRS coordinates are searched but UTM coordinates should be displayed. I chose the least invasive way and just added a try-catch and in case the already implemented error message is automatically shown.
- I only translated it to English, as 'MGRS' is the same abbreviation for all languages. Only the description text might be good to be translated to other languages later.

### DIFF
--- a/Resources/Localizations/en.lproj/Localizable.strings
+++ b/Resources/Localizations/en.lproj/Localizable.strings
@@ -808,7 +808,6 @@
 "navigate_point_zone" = "Zone";
 "navigate_point_olc" = "Open Location Code";
 "navigate_point_olc_short" = "OLC";
-"navigate_point_mgrs" = "MGRS";
 "approximate_file_size" = "Approximate file size: %@";
 
 "coordinates_example" = "Example";
--- a/Resources/Localizations/en.lproj/Localizable.strings
+++ b/Resources/Localizations/en.lproj/Localizable.strings
@@ -801,16 +801,19 @@
 "navigate_point_format_DMS" = "DDD MM SS.SS";
 "navigate_point_format_OLC" = "Open Location Code (OLC)";
 "navigate_point_format_UTM" = "UTM";
+"navigate_point_format_MGRS" = "MGRS";
 "navigate_point_format_unknown" = "Unknown format";
 "navigate_point_northing" = "Northing";
 "navigate_point_easting" = "Easting";
 "navigate_point_zone" = "Zone";
 "navigate_point_olc" = "Open Location Code";
 "navigate_point_olc_short" = "OLC";
+"navigate_point_mgrs" = "MGRS";
 "approximate_file_size" = "Approximate file size: %@";
 
 "coordinates_example" = "Example";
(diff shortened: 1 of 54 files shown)


Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -801,16 +801,19 @@
...
 "navigate_point_format_OLC" = "Open Location Code (OLC)";
 "navigate_point_format_UTM" = "UTM";
+"navigate_point_format_MGRS" = "MGRS";
 "navigate_point_format_unknown" = "Unknown format";
 "navigate_point_northing" = "Northing";
 "navigate_point_easting" = "Easting";
 "navigate_point_zone" = "Zone";
 "navigate_point_olc" = "Open Location Code";
 "navigate_point_olc_short" = "OLC";
+"navigate_point_mgrs" = "MGRS";
```
PaulStets, line 811: Please remove this line and use the one above, we try not to duplicate string resources.

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -577,6 +634,68 @@ - (NSString *) mgrsZoneToUTMZone:(NSString *)mgrsZone
...
+    if (s.length < 3
+        || !([[NSCharacterSet decimalDigitCharacterSet] characterIsMember:[s characterAtIndex:0]]
+             || [s characterAtIndex:0] == 'A' || [s characterAtIndex:0] == 'a'
+             || [s characterAtIndex:0] == 'B' || [s characterAtIndex:0] == 'b'
+             || [s characterAtIndex:0] == 'Y' || [s characterAtIndex:0] == 'y'
+             || [s characterAtIndex:0] == 'Z' || [s characterAtIndex:0] == 'z'
+             )
+        )
+    {
+        return false;
```
PaulStets, line 694: Obj c convention for BOOL is ```YES/NO```.

Resources/Localizations/en.lproj/Localizable.strings
jBackhaus: Done with 31fe5a0

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
jBackhaus: Done with bc953ed

PaulStets:
Perfect, thank you for the contribution!

PR #1752, opened 2022-02-01
Fix Current Recording Track - Waypoints, Appearance, Show/Hide, Tap

Skalii:


Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -2947,11 +2953,21 @@ - (void) initRendererWithGpxTracks
...
+
         [_mapLayers.gpxMapLayer refreshGpxTracks:docs];
     }
 }
 
+- (void)initRendererWithRecordingGpxTrack
+{
+    if (_gpxDocsRec.size() > 0)
+    {
+        QHash<QString, std::shared_ptr<const OsmAnd::GpxDocument>> docs;
```
PaulStets, line 2965: Why is is it added each time now?

PR #1759, opened 2022-02-03
Fix right widget frame when download map widget is visible

Skalii:


Sources/Controllers/Map/OAMapHudViewController.mm
```diff
@@ -827,7 +827,7 @@ - (void) updateControlsLayout:(CGFloat)y statusBarColor:(UIColor *)statusBarColo
     
     if (_widgetsView)
     {
-        CGFloat widgetsOffset = _downloadMapWidget.isVisible ? _downloadMapWidget.shadowOffset : 2.0;
+        CGFloat widgetsOffset = 2.0;
```
PaulStets, line 830: Make it constant

PR #1798, opened 2022-03-08
Save as new track UI while exporting track from Navigation

yuliiastets:


Sources/Helpers/OASaveGpxToTripsActivity.mm
```diff
@@ -0,0 +1,96 @@
...
+        
+        NSString *folderName = @"";
+        if ([NSFileManager.defaultManager fileExistsAtPath:[OsmAndApp.instance.gpxPath stringByAppendingPathComponent:kImportFolderName]])
+            folderName = kImportFolderName;
+        NSString *shortPath = [folderName stringByAppendingPathComponent:fileName];
+        NSString *fullFolderPath = [OsmAndApp.instance.gpxPath stringByAppendingPathComponent:folderName];
+        
+        NSInteger index = 1;
+        while ([NSFileManager.defaultManager fileExistsAtPath:[[fullFolderPath stringByAppendingPathComponent:fileName] stringByAppendingPathExtension:kGpxFileExtension]]) {
+            fileName = [NSString stringWithFormat:@"%@(%ld)", name, index];
```
alex-osm, line 85: Do we need some separator between %@(%ld)?

Sources/Controllers/RoutePlanning/OASaveTrackViewController.mm
```diff
@@ -266,6 +295,31 @@ - (IBAction)saveButtonPressed:(id)sender
                             simplifiedTrack:_simplifiedTrack
                                   openTrack:YES];
         }
+        
+        if (_exportingFromNavigation)
```
alex-osm, line 299: Use delegate instead of _exportingFromNavigation

PR #1833, opened 2022-03-15
"OsmAnd Maps" folder

Skalii:


Resources/OsmAnd-Info.plist
```diff
@@ -111,7 +111,7 @@
 			<key>CFBundleTypeRole</key>
 			<string>Viewer</string>
 			<key>LSHandlerRank</key>
-			<string>Owner</string>
+			<string>Alternate</string>
```
PaulStets, line 114: obf should remain owner

Resources/OsmAnd-Info.plist
```diff
@@ -72,7 +72,7 @@
...
```
PaulStets, line 75: osf should remain owner

PR #1861, opened 2022-04-04
Weather - Plugin

Skalii:


Sources/Plugins/Weather/OAWeatherPlugin.mm
```diff
@@ -0,0 +1,63 @@
...
+    [[OsmAndApp instance].data setWeather:enabled ? [_lastUsedWeather get] : NO];
+}
+
+- (void)weatherChanged:(BOOL)isOn
+{
+    [_lastUsedWeather set:isOn];
+    [[OsmAndApp instance].data setWeather:isOn];
+}
+
++ (NSArray<NSString *> *)getWeatherSettingKeys
```
PaulStets, line 58: Where is - (void) registerWidget method and other methods that create controls (see OAParkingPositionPlugin)?

PaulStets, line 58: Where is - (void) registerWidget method and other methods that create controls (see OAParkingPositionPlugin)?

PR #1868, opened 2022-04-12
Fix Import - files from custom plugin

Skalii:


Sources/Controllers/Settings/ImportExport/OAImportSettingsViewController.h
```diff
@@ -13,6 +13,6 @@
 
 - (instancetype) initWithItems:(NSArray<OASettingsItem *> *)items;
 
-- (void)onItemsCollected:(NSArray<OASettingsItem *> *)items;
+- (void)onItemsCollected:(NSArray<OASettingsItem *> *)items file:(NSString *)file;
```
PaulStets, line 16: ```filePath``` is not a file

PR #1954, opened 2022-05-14
Fix history - duplicate entries

Skalii:


Sources/History/OAHistoryHelper.m
```diff
@@ -42,6 +42,7 @@ - (instancetype)init
 
 - (void)addPoint:(OAHistoryItem *)item
 {
+    [_db deleteDuplicate:item];
     [_db addPoint:item.latitude longitude:item.longitude time:[item.date timeIntervalSince1970] name:item.name type:item.hType iconName:item.iconName typeName:item.typeName];
```
PaulStets, line 46: Instead of deleting, update the item if it exists

PR #2111, opened 2022-08-10
Weather forecast - Multi-select & Info screen

Skalii:


Sources/Controllers/Resources/OAOutdatedResourcesViewController.mm
```diff
@@ -397,15 +398,24 @@ - (UITableViewCell*)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NS
...
     if (indexPath.section == _updatesSection)
     {
-        cellTypeId = liveUpdatesCell;
-        title = OALocalizedString(@"osmand_live_updates");
+        if (indexPath.row == 0)
+        {
+            cellTypeId = liveUpdatesCell;
+            title = OALocalizedString(@"osmand_live_updates");
+        }
+        else if (indexPath.row == 1)
```
PaulStets, line 414: variable in the first case but not in here?

Sources/Controllers/Resources/OAWeatherForecastViewController.mm
```diff
@@ -0,0 +1,1181 @@
...
+                        [_weatherHelper downloadForecastsByRegionIds:forecastsToDownload];
+                        [forecastsToDownload removeAllObjects];
+                    }]];
+                    [self presentViewController:alert animated:YES completion:nil];
+                }
+            };
+        }
+    }
+
+    dispatch_block_t executingBlock = ^{
```
PaulStets, line 741: Check these blocks for retain cycles) They are more than possible here.

Sources/Controllers/Resources/OAManageResourcesViewController.mm
```diff
@@ -3344,4 +3356,18 @@ - (void)clearMultipleResources
     _multipleItems = nil;
 }
 
+#pragma mark - OAWeatherForecastDetails
+
+- (void)onRemoveForecast
```
PaulStets, line 3361: Why do you need 2 methods that do the same thing?

PR #2121, opened 2022-08-14
Fix Weather Toolbar

Skalii:


Sources/Controllers/Map/Toolbars/OAWeatherToolbar.mm
```diff
@@ -91,11 +96,37 @@ - (void)commonInit
...
+    }
+    [self updateInfo];
+}
+
+- (void)dealloc
+{
+    for (OAAutoObserverProxy *observer in _layerChangeObservers)
+        [observer detach];
+
+    [_layerChangeObservers removeAllObjects];
```
PaulStets, line 114: Use _layerChangeObservers = nil instead

PR #2243, opened 2022-10-19
Fix bearing

ivanPyrohivskyi:
Issue https://github.com/osmandapp/OsmAnd-Issues/issues/1597
Android PR https://github.com/osmandapp/OsmAnd/pull/15499


Sources/Router/OARouteCalculationResult.mm
```diff
@@ -384,6 +385,20 @@ - (CLLocation *) getNextRouteLocation:(int)after
     return nil;
 }
 
+- (CLLocation *) getRouteLocationByDistance:(int)meters {
```
PaulStets, line 388: We format obj c to start { from the new line

PR #2282, opened 2022-11-01
OpenStreetMap OAuth login and Fix bug

DmitrySvetlichny:
https://user-images.githubusercontent.com/111898301/199243895-9281bb6e-6739-4514-a2e8-4dd975535361.mp4

### DIFF
--- a/Sources/Controllers/Settings/OSMEditing/OABenefitsOsmContributorsViewController.m
+++ b/Sources/Controllers/Settings/OSMEditing/OABenefitsOsmContributorsViewController.m
@@ -178,11 +178,11 @@
 
 #pragma mark - OAAccountSettingDelegate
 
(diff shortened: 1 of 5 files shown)


Sources/Controllers/Settings/OSMEditing/OABenefitsOsmContributorsViewController.m
```diff
@@ -178,11 +178,11 @@ - (CGFloat)heightForRow:(NSIndexPath *)indexPath estimated:(BOOL)estimated
 
 #pragma mark - OAAccountSettingDelegate
 
-- (void)onAccountInformationUpdated
+- (void)onAccountInformationUpdatedWithBenefitsShown:(BOOL)shouldOpenBenefits
```
PaulStets, line 181: Why do we need this parameter if it's not used?

PR #2294, opened 2022-11-04
Follow track

hasanov080:


Sources/Controllers/RouteInfoMenu/OARoutingSettingsCell.mm
```diff
@@ -76,10 +75,14 @@ - (void) adjustInsets:(UIButton *)btn
 
 - (void) adjustButtonSize
```
PaulStets, line 76: Rename to adjustButtonsSize

PR #2296, opened 2022-11-08
Separate Weather layer settings

Skalii:


Sources/Helpers/OAAppSettings.h
```diff
@@ -927,7 +927,11 @@ typedef NS_ENUM(NSInteger, EOARateUsState)
 // Custom plugins
 @property (nonatomic) NSString *customPluginsJson;
 
+// Weather
+
 @property (nonatomic) OACommonBoolean *weatherButtonIsOn;
+@property (nonatomic) BOOL weatherToolbarVisible;
```
PaulStets, line 933: We specifically introduced settings classes to avoid using primitives

Sources/Controllers/Map/OAMapInfoController.mm
```diff
@@ -480,7 +476,7 @@ - (void)showWeatherToolbar
 
         [_mapHudViewController showBottomControls:[OAUtilities isLandscape] ? 0. : _weatherToolbar.frame.size.height - [OAUtilities getBottomMargin]
                                          animated:YES];
-        [mapPanelViewController setTopControlsVisible:YES];
+        [[OARootViewController instance].mapPanel setTopControlsVisible:YES];
```
PaulStets, line 479: Why not use the variable you created at 460?

PR #2593, opened 2023-02-21
Add several recycling waste types

eunikolsky:
This adds four new types of recycling waste that I see in OSM. They are all present in the table for [`amenity=recycling`][table].

I've verified the changes locally after creating a piece of the map with these recycling POIs.

recycling@2x

[table]: https://wiki.openstreetmap.org/wiki/Tag:amenity=recycling#Accepts

### DIFF
--- a/Resources/phrases.xml
+++ b/Resources/phrases.xml
@@ -389,6 +389,10 @@
 	<string name="poi_recycling_car_batteries">Car batteries</string>
 	<string name="poi_recycling_cars">Cars</string>
 	<string name="poi_recycling_bicycles">Bicycles</string>
+	<string name="poi_recycling_plastic_bottle_caps">Plastic bottle caps</string>
+	<string name="poi_recycling_aluminium_cans">Aluminium cans</string>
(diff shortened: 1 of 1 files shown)


eunikolsky:
@vshcherb, I didn't know that, thanks! I'm confused here:
1. Do I need to do anything else? Submit a PR into `OsmAnd`? When `resources` is on my PR https://github.com/osmandapp/OsmAnd-resources/pull/937, I already see the changes in iOS in English, but not translations.
2. How to update the iOS project after changing `OsmAnd/res/values/phrases.xml`? `resources/poi/copy_phrases.sh` (but it's a commented build step in the `OsmAnd Maps` target)?
3. After running that script, it reverted the `ru`/`uk` localization files and added a bunch of other changes. Does it mean that the translations from my PR such as https://github.com/osmandapp/OsmAnd-resources/pull/937/files#diff-154c5b2d5399c287090e7a116a90ab9f6f1645cd4e56d3bebbb05f24e4573f9fR291 need to be moved to the `android` repository?

PR #2637, opened 2023-03-07
Wikipedia plugin - settings / Wikipedia article screen redesign

Skalii:


Sources/Helpers/OAAppSettings.m
```diff
@@ -1266,6 +1266,115 @@ - (NSString *)description
 
 @end
 
+@interface OADownloadMode()
```
PaulStets, line 1269: Extract to a separate file. Why not use the static getters like in other Java-like enums that we turned into classes?

Sources/Models/OATableSectionData.h
```diff
@@ -23,6 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
 - (OATableRowData *) getRow:(NSUInteger)index;
 
 - (void)addRow:(OATableRowData *)rowData;
+- (void)addRows:(NSArray<OATableRowData *> *)rowsData;
```
PaulStets, line 26: rename parameter to simply ```rows``` or ```rowsToAdd```

PR #2694, opened 2023-03-21
Accessibility add value

DmitrySvetlichny:


Sources/Controllers/Map/OAMapHudViewController.mm
```diff
@@ -1474,6 +1496,34 @@ - (void) updateMapSettingsButton
     _mapSettingsButton.tintColorDay = UIColorFromRGB(mode.getIconColor);
     _mapSettingsButton.tintColorNight = UIColorFromRGB(mode.getIconColor);
     [_mapSettingsButton updateColorsForPressedState:NO];
+    
+    NSString *stringKey = _settings.applicationMode.get.stringKey;
```
PaulStets, line 1500: Extract to a method and remove settings from class fields since you use it only once.

PR #2701, opened 2023-03-23
Hide Default speed parameter from Public transport profile

DmitrySvetlichny:


Sources/Controllers/Settings/NavigationSettings/OAVehicleParametersViewController.mm
```diff
@@ -153,7 +153,8 @@ - (void)generateData
         [tableData addObject:defaultSpeedArr];
         _otherSection = tableData.count - 1;
     }
-    _data = [NSArray arrayWithArray:tableData];
+    if (!isPublicTransport)
```
PaulStets, line 156: Why do you populate the data if you don't use it?

PR #2737, opened 2023-03-27
Customise route line - fix back button

Skalii:


Sources/Controllers/RouteInfoMenu/OARouteLineAppearanceHudViewController.mm
```diff
@@ -251,6 +255,19 @@ - (instancetype)initWithAppMode:(OAApplicationMode *)appMode
     return self;
 }
 
+- (instancetype)initWithAppMode:(OAApplicationMode *)appMode prevScreenKey:(NSString *)prevScreenKey
```
PaulStets, line 258: Make prevScreenKey an enum instead

PR #2825, opened 2023-04-22
Favorites - show emoji

Skalii:


alex-osm:
Drop the checkEmoticons method if it's no longer used

PR #2885, opened 2023-05-17
chore: Fix some typos in German translation

ffflorian:
Hi, while using OsmAnd on iOS in German I noticed some typos and found a few more while fixing the others. :)

I also noticed that the strings are switching between the informal "Du" and the formal "Sie". Is that intended?


vshcherb:
CC @sonora

sonora:
@ffflorian  Very good, thanks! :wink:

PR #2967, opened 2023-06-26
fix free space check for map update

Skalii:


Sources/Controllers/Resources/OAOutdatedResourcesViewController.mm
```diff
@@ -230,13 +230,24 @@ - (void)offerDownloadAndUpdateMultiple:(NSArray*)items
...
+    }];
+    
     for (OAOutdatedResourceItem* item in items)
     {
         const auto resourceInRepository = _app.resourcesManager->getResourceInRepository(item.resourceId);
-
         totalDownloadSize += resourceInRepository->packageSize;
-        totalSpaceNeeded += resourceInRepository->packageSize + resourceInRepository->size;
     }
+    totalSpaceNeeded = totalDownloadSize + _app.resourcesManager->getResourceInRepository(((OAOutdatedResourceItem *) items.firstObject).resourceId)->size;
```
vshcherb, line 250: This probably doesn't look  correct cause there is no check if the map needs to be jsut updated or downloaded again.

So in case of update you don't need to sum the maps size, you need to find maximum

PR #3074, opened 2023-08-18
Allow to access OsmAnd Cloud settings while checking status Fix

DmitrySvetlichny:


Sources/Backup/OABackupHelper.mm
```diff
@@ -31,6 +31,7 @@
 #import "OARegisterUserCommand.h"
 #import "OARegisterDeviceCommand.h"
 #import "OAURLSessionProgress.h"
+#import "OACloudBackupViewController.h"
```
alex-osm, line 34: OABackupHelper should not have any links to UI classes

PR #3124, opened 2023-09-19
Terrain maps review

DmitrySvetlichny:


Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsTerrainParametersViewController.mm
```diff
@@ -147,15 +159,36 @@ - (void)generateData
     else
     {
         [topSection addRowFromDictionary:@{
-            kCellKeyKey : @"zoomSlider",
-            kCellTypeKey : [OARangeSliderCell getCellIdentifier]
+            kCellTypeKey : [OAValueTableViewCell getCellIdentifier],
+            kCellTitleKey: OALocalizedString(@"rec_interval_minimum"),
+            @"value" : @(_minZoom)
+        }];
+        _minValueIndexPath = [NSIndexPath indexPathForRow:[_data rowCount:[_data sectionCount] - 1] - 1 inSection:[_data sectionCount] - 1];
```
alex-osm, line 166: Since we introduced `OATableSectionData` and `OATableRowData`, we can add index field to both and use it instead of calling `_data rowCount` and `_data sectionCount` when needed

PR #3126, opened 2023-09-20
Avoid series of same types alarms

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/17723
Code porting from Java https://github.com/osmandapp/OsmAnd/pull/18090

### DIFF
--- a/Sources/Router/OAWaypointHelper.mm
+++ b/Sources/Router/OAWaypointHelper.mm
@@ -37,6 +37,7 @@
 // don't annoy users by lots of announcements
 #define APPROACH_POI_LIMIT 1
 #define ANNOUNCE_POI_LIMIT 3
+#define SAME_ALARM_INTERVAL 30//in seconds
 
 @implementation OAWaypointHelper
 {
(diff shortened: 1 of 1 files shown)


Sources/Router/OAWaypointHelper.mm
```diff
@@ -404,6 +408,17 @@ - (void) announceVisibleLocations
                                             proceed = NO;
                                         }
                                     }
+                                    NSNumber * timeLastAlarm = [_lastAnnouncedAlarmsTime objectForKey:@(t)];
+                                    if (timeLastAlarm && proceed)
+                                    {
+                                        long ms = CACurrentMediaTime();
```
alex-osm, line 414: long > NSTimeInterval

Sources/Router/OAWaypointHelper.mm
```diff
@@ -404,6 +408,17 @@ - (void) announceVisibleLocations
...
+                                        if (ms - [timeLastAlarm longValue] < SAME_ALARM_INTERVAL * 1000)
```
alex-osm, line 415: Extract doubleValue and do not * 1000

PR #3163, opened 2023-10-12
Night theme for UI set

DmitrySvetlichny:
Simulator Screenshot - iPhone 12 - 2023-10-12 at 12 16 26


Sources/AppHost/OsmAndAppImpl.mm
```diff
@@ -723,6 +724,9 @@ - (BOOL) initializeImpl
     OAPOIFiltersHelper *helper = [OAPOIFiltersHelper sharedInstance];
     [helper reloadAllPoiFilters];
     [helper loadSelectedPoiFilters];
+    
+    // Configure ThemeManager
+    [[ThemeManager shared] configureWithAppMode:initialAppMode];
```
alex-osm, line 729: Might be called in non UI thread. Is it ok?

PR #3188, opened 2023-10-25
fix configure map - 3d relief purchase

Skalii:


Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsTerrainScreen.mm
```diff
@@ -801,18 +805,23 @@ - (void) onLocalResourcesChanged:(id<OAObservableProtocol>)observer withKey:(id)
 
 - (void)productPurchased:(NSNotification *)notification
 {
-    dispatch_async(dispatch_get_main_queue(), ^{
-        [self initData];
-        [self.tblView reloadData];
-    });
+    [_plugin.enable3DMaps set:YES];
```
alex-osm, line 808: It will not work if purchase will be done from other screen

Sources/Plugins/SRTMPlugin/OASRTMPlugin.mm
```diff
@@ -88,11 +88,14 @@ - (void) onProfileSettingSet:(NSNotification *)notification
...
-            [OARootViewController.instance.mapPanel.mapViewController recreateHeightmapProvider];
-        });
-    }
+        [self updateLayers];
+}
+
+- (void)updateLayers
+{
+    dispatch_async(dispatch_get_main_queue(), ^{
+        [OARootViewController.instance.mapPanel.mapViewController recreateHeightmapProvider];
```
alex-osm, line 97: Looks like updateLayers is called from updateCurrentMapSource like recreateHeightmapProvider. So it might be wrong fix.

PR #3198, opened 2023-11-03
UI Fix

DmitrySvetlichny:


Sources/Controllers/Settings/OSMEditing/OAOsmNoteViewController.mm
```diff
@@ -54,6 +54,7 @@ - (instancetype)initWithEditingPlugin:(OAOsmEditingPlugin *)plugin points:(NSArr
         _bugPoints = points;
         _plugin = plugin;
         _screenType = type;
+        _messageText = ((OAOsmNotePoint *)_bugPoints.firstObject).getText;
```
alex-osm, line 57: Not clear why `_bugPoints` array is needed if we are using single point

PR #3260, opened 2023-12-04
Ble add running sensor

tigrim:


Sources/Plugins/ExternalSensorsPlugin/BLE/BLEManager/SensorWidgetDataField.swift
```diff
@@ -27,7 +27,8 @@ class SensorSpeedWidgetDataField: SensorWidgetDataField {
         if let value = numberValue?.floatValue {
             let formattedSpeed = OAOsmAndFormatter.getFormattedSpeed(value).components(separatedBy: " ")
```
alex-osm, line 28: `getFormattedSpeed` should be able to return separated value / unit

Sources/Plugins/ExternalSensorsPlugin/BLE/BLEManager/SensorWidgetDataField.swift
```diff
@@ -42,7 +43,8 @@ class SensorDistanceWidgetDataField: SensorWidgetDataField {
         if let value = numberValue?.floatValue {
             let formattedDistance = OAOsmAndFormatter.getFormattedDistance(value, forceTrailingZeroes: false).components(separatedBy: " ")
```
alex-osm, line 44: `getFormattedDistance ` should be able to return separated value / unit

PR #3263, opened 2023-12-05
fix start navigation button on CarPlay

Skalii:


tigrim (approved):

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -612,6 +612,18 @@ - (void)onMapViewAttached
     if (route && start && _routingHelper.isRouteCalculated)
     {
         [self enterRoutePreviewMode];
+        if ([_routingHelper isFollowingMode])
+            [self onTripStartTriggered];
+    }
+}
+
+- (void)onMapViewDettached
```
alex-osm, line 620: typo

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -612,6 +612,18 @@ - (void)onMapViewAttached
...
+{
+    [NSNotificationCenter.defaultCenter removeObserver:self];
```
alex-osm, line 622: Why notifications and location observer are not created at `onMapViewAttached` in this case?
Remove `dealloc` method

PR #3270, opened 2023-12-06
fix gpx analyze on map

Skalii:


Sources/Controllers/TargetMenu/OARouteBaseViewController.mm
```diff
@@ -709,6 +709,14 @@ - (void) centerMapOnRoute
             [self centerMapOnBBox:transportHelper.getBBox];
         }
     }
+    else if (_gpx)
+    {
+        routeBBox.top = _gpx.bounds.topLeft.latitude;
```
tigrim, line 714: From the improvements, gpx.bounds can be extracted into a separate variable

PR #3280, opened 2023-12-08
Crash import favourites replace all action applied

tigrim:


Sources/Controllers/TargetMenu/Favorites/OAFavoriteImportViewController.mm
```diff
@@ -18,6 +18,8 @@
 #import "OAColors.h"
 #import "Localization.h"
 
+NSNotificationName const OAFavoriteImportViewControllerDidDismissNotification = @"OAFavoriteImportViewControllerDidDismissNotification";
```
alex-osm, line 21: Do not use notification for that. It is better to use delegate.

Sources/Controllers/TargetMenu/Favorites/OAFavoriteImportViewController.mm
```diff
@@ -279,7 +284,7 @@ - (BOOL)isFavoritesValid
                                 OAPointsGroup *group = _gpxFile.pointsGroups[keyGroup];
                                 for (OAWptPt *wptPt in group.points)
                                 {
-                                    for (OAFavoriteItem *localFavortite in favoriteItems)
+                                    for (OAFavoriteItem *localFavortite in [favoriteItems copy])
```
alex-osm, line 287: `copy` should not be used there

PR #3283, opened 2023-12-11
fix gpx analyze on map;

Skalii:


Sources/Controllers/TargetMenu/OARouteBaseViewController.mm
```diff
@@ -693,21 +693,17 @@ - (void) centerMapOnRoute
     routeBBox.bottom = DBL_MAX;
     routeBBox.left = DBL_MAX;
     routeBBox.right = DBL_MAX;
-    if ([_routingHelper isRouteCalculated] && !error && !_routingHelper.isPublicTransportMode)
+    if ([_routingHelper isRoutePlanningMode] && [_routingHelper isRouteCalculated] && !error && !_routingHelper.isPublicTransportMode)
```
tigrim, line 696: [_routingHelper isRoutePlanningMode] && [_routingHelper isRouteCalculated] is this possible? Maybe there should be "||" there?

PR #3285, opened 2023-12-11
use deviceServiceName for empty device name

tigrim:


Sources/Plugins/ExternalSensorsPlugin/BLE/DeviceHelper.swift
```diff
@@ -184,6 +184,7 @@ extension DeviceHelper {
     }
     
     func clearConnectedDevicesList() {
```
alex-osm, line 186: It is not clear why this method needed and why it has public access? Looks like disconnectAllDevices should be enough and clear for callers.

PR #3294, opened 2023-12-12
fix gpx ble

Skalii:
close screen after selecting sensor;


tigrim (approved):

Sources/Plugins/ExternalSensorsPlugin/BLE/BLEPairedSensors/BLEPairedSensorsViewController.swift
```diff
@@ -171,7 +171,7 @@ final class BLEPairedSensorsViewController: OABaseNavbarViewController {
             }
             onSelectDeviceAction?(currentSelectedDevice)
         }
-        tableView.reloadData()
```
tigrim, line 174: I think tableView.reloadData() should be kept. The user wants to see what they selected before the controller closes.

tigrim (approved):

PR #3332, opened 2023-12-22
Remove all old cells

DmitrySvetlichny:


Sources/Controllers/Cloud/OACloudBackupViewController.mm
```diff
@@ -501,21 +500,7 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
 {
     OATableRowData *item = [_data itemForIndexPath:indexPath];
     NSString *cellId = item.cellType;
-    if ([cellId isEqualToString:OATitleRightIconCell.getCellIdentifier])
```
tigrim, line 504: Was it unnecessary? Or should it be replaced with something?

PR #3344, opened 2023-12-25
Fixed

DmitrySvetlichny:


Sources/Controllers/Cloud/OACloudBackupViewController.mm
```diff
@@ -501,21 +500,7 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
 {
     OATableRowData *item = [_data itemForIndexPath:indexPath];
     NSString *cellId = item.cellType;
-    if ([cellId isEqualToString:OATitleRightIconCell.getCellIdentifier])
```
tigrim, line 504: Was it unnecessary? Or should it be replaced with something?

DmitrySvetlichny: This type of cell apparently was replaced earlier and they forgot to remove it from the cellForRow + there are no dates in the generateData for this type of cell anymore.

tigrim (approved):

PR #3383, opened 2024-01-24
fix carplay 3d switching

Skalii:


tigrim (approved):

Sources/Controllers/Map/OAMapViewTrackingUtilities.mm
```diff
@@ -186,6 +186,13 @@ - (void) onMapModeChanged
                     
 
                     _mapView.mapAnimator->resume();
+
+                    OAMapViewController *mapViewController = [OARootViewController instance].mapPanel.mapViewController;
+                    if (mapViewController.needToSwitchTo3DMode && ![self isIn3dMode])
+                    {
+                        mapViewController.needToSwitchTo3DMode = NO;
+                        [self switchMap3dMode];
```
alex-osm, line 194: Seems that 3D mode is turned on regardless of whether it was turned on before or not. Looks wrong.

Sources/Controllers/Map/OAMapViewController.h
```diff
@@ -106,6 +106,7 @@
 
 @property (nonatomic, assign) BOOL isCarPlayActive;
 @property (nonatomic, assign) BOOL isCarPlayDashboardActive;
+@property (nonatomic, assign) BOOL needToSwitchTo3DMode;
```
alex-osm, line 109: Bad approach.

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -133,6 +133,7 @@ - (void) enterBrowsingState
 
 - (void) enterRoutePreviewMode
 {
+    [OARootViewController instance].mapPanel.mapViewController.needToSwitchTo3DMode = YES;
```
alex-osm, line 136: Shouldn't this work for the entire application, not just CarPlay?

PR #3408, opened 2024-02-02
Review Fixed

DmitrySvetlichny:


Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsMainScreen.mm
```diff
@@ -189,7 +189,8 @@ - (void)setupView
     [showSectionData addObject:@{
         @"name": OALocalizedString(@"shared_string_gpx_tracks"),
         @"image": @"ic_custom_trip",
-        @"type": [OASimpleTableViewCell getCellIdentifier],
+        @"value": [NSString stringWithFormat:@"%lu", (unsigned long)_settings.mapSettingVisibleGpx.get.count],
```
tigrim, line 192: @"%d", (int)_settings.mapSettingVisibleGpx.get.count] ?

Sources/Controllers/DashboardOnMap/MapSettings/MapSettingsGpxViewController.swift
```diff
@@ -862,6 +861,16 @@ extension MapSettingsGpxViewController: UISearchBarDelegate {
     }
 }
 
+extension MapSettingsGpxViewController: UISearchControllerDelegate {
+    func presentSearchController(_ searchController: UISearchController) {
+        DispatchQueue.main.async {
```
tigrim, line 866: Do we really need to use DispatchQueue.main.async here?

DmitrySvetlichny: Yes

tigrim (approved):

PR #3411, opened 2024-02-05
Fixed Crash

DmitrySvetlichny:


Sources/Controllers/Settings/WidgetSettings/WidgetsListViewController.swift
```diff
@@ -412,27 +412,41 @@ extension WidgetsListViewController {
             }
         }
         
-        let destinationItem = tableData.item(for: proposedDestinationIndexPath)
+        let sectionCount = tableData.sectionCount()
+        var correctedIndexPath = proposedDestinationIndexPath
+        if proposedDestinationIndexPath.section >= sectionCount {
```
tigrim, line 417: if proposedDestinationIndexPath.section >= sectionCount || proposedDestinationIndexPath.row >= rowCount {
return sourceIndexPath
}
?

PR #3433, opened 2024-02-12
GPX - nearest city

Skalii:


Sources/Address/OACity.h
```diff
@@ -39,4 +39,9 @@ typedef NS_ENUM(NSInteger, EOACitySubType)
 
 - (instancetype)initWithCity:(const std::shared_ptr<const OsmAnd::StreetGroup>&)city;
 
++ (NSString *)getLocalizedTypeStr:(EOACitySubType)type;
```
tigrim, line 42: For header files with reference type parameters, use _Nullable/_Nonnull

tigrim (approved):

PR #3462, opened 2024-02-27
Sort for tracks in configure map review 2

DmitrySvetlichny:


Sources/GPX/OAGPXDatabase.mm
```diff
@@ -365,6 +366,7 @@ - (void)load
         OAGPX *gpx = [self generateGpxItem:gpxData];
 
         // Make compatible with old database data
+        gpx.creationDate = gpx.creationDate ?: gpx.importDate;
```
tigrim, line 369: if  !(gpx.creationDate){}

tigrim (approved):

Sources/GPX/OAGPXDatabase.mm
```diff
@@ -365,6 +366,9 @@ - (void)load
         OAGPX *gpx = [self generateGpxItem:gpxData];
 
         // Make compatible with old database data
+        if (!gpx.creationDate)
+            gpx.creationDate = gpx.importDate;
```
alex-osm, line 370: You should implement [similar](https://github.com/osmandapp/OsmAnd/blob/c37664ae270490687cf4e171c929a2cd258c62fc/OsmAnd-java/src/main/java/net/osmand/gpx/GPXUtilities.java#L1392) method to get creation date

PR #3470, opened 2024-02-28
fix ble temperature

Skalii:


Sources/Plugins/ExternalSensorsPlugin/OAExternalSensorsPlugin.mm
```diff
@@ -217,19 +217,19 @@ - (void)onAnalysePoint:(OAGPXTrackAnalysis *)analysis point:(NSObject *)point at
 {
     if ([point isKindOfClass:OAWptPt.class])
     {
+        NSNumberFormatter *numberFormatter = [[NSNumberFormatter alloc] init];
+        numberFormatter.decimalSeparator = @".";
```
tigrim, line 221: Can there be ',' here depending on the locale?

tigrim (approved):

PR #3476, opened 2024-03-03
fix basemap language

Skalii:


Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -136,6 +136,8 @@
 #define ZONE_1_ZOOM_THRESHOLD 0.6f
 #define ZONE_2_ZOOM_THRESHOLD 1.5f
 
+#define kReplaceLocalNamesMaxZoom 6
```
tigrim, line 139: static const

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -2121,6 +2127,36 @@ - (void) resetViewAngle
...
+    @synchronized(_rendererSync)
+    {
+        [self updateMapLocaleLanguage];
+    }
+}
+
+- (void)updateMapLocaleLanguage
+{
+    NSString *langId = [self getMapPreferredLocale:_mapView.zoomLevel];
+    if (_mapPresentationEnvironment != nullptr && ![langId isEqualToString:_mapPresentationEnvironment->getLocaleLanguageId().toNSString()])
```
tigrim, line 2141: It is more optimal to first check _mapPresentationEnvironment != nullptr, and then call NSString *langId = ...

tigrim (approved):

PR #3499, opened 2024-03-11
Wrong interface colors in status bar

DmitrySvetlichny:


Sources/Controllers/Base/OABaseScrollableHudViewController.mm
```diff
@@ -95,8 +96,12 @@ - (void)viewWillAppear:(BOOL)animated
 
 - (UIStatusBarStyle)preferredStatusBarStyle
 {
-    return self.currentState != EOADraggableMenuStateFullScreen && [OAAppSettings sharedManager].nightMode
-            ? UIStatusBarStyleLightContent : UIStatusBarStyleDefault;
+    if (self.currentState != EOADraggableMenuStateFullScreen && [OAAppSettings sharedManager].nightMode)
```
tigrim, line 99: move self.currentState != EOADraggableMenuStateFullScreen in let

tigrim (approved):

PR #3512, opened 2024-03-18
Expand bbox to cover Earth for basemap search

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/18919

vshcherb:
Code incorrect

PR #3518, opened 2024-03-20
Line Width from strokeWidth_0 for nightMode

DmitrySvetlichny:


Sources/Controllers/Map/Layers/OAPreviewRouteLineLayer.mm
```diff
@@ -337,7 +337,7 @@ - (CGFloat)getLineWidth
     }
     else
     {
-        width = [self getParamFromAttr:@"strokeWidth"].floatValue;
+        width = [self getParamFromAttr:[OAAppSettings sharedManager].nightMode ? @"strokeWidth_0" : @"strokeWidth"].floatValue;
```
alex-osm, line 340: Probably not correct fix. Should be compared with android code.

PR #3522, opened 2024-03-22
Fix difference between POI categories in Android / iOS version #18971

Dima-1:


Sources/POI/OAPOIHelper.mm
```diff
@@ -371,6 +372,19 @@ - (NSString *) getSynonyms:(OAPOIBaseType *)type
     return [self getSynonymsByName:type.name];
 }
 
+- (void) setPoiTranslator
+{
+    [self sortList:_poiCategories];
+}
+
+- (void) sortList: categories
```
alex-osm, line 380: Where is `categories` type?

Sources/POI/OAPOIHelper.mm
```diff
@@ -371,6 +372,19 @@ - (NSString *) getSynonyms:(OAPOIBaseType *)type
     return [self getSynonymsByName:type.name];
 }
 
+- (void) setPoiTranslator
```
alex-osm, line 375: This method is not related to translator. Should be renamed.

PR #3523, opened 2024-03-23
fix CarPlay turn arrows

Skalii:


Sources/Controllers/Map/Widgets/OATurnDrawable.mm
```diff
@@ -117,7 +132,7 @@ - (void) drawRect:(CGRect)rect
         //NSMutableParagraphStyle *paragraphStyle = [[NSMutableParagraphStyle alloc] init];
```
tigrim, line 132: remove comments

Sources/Controllers/Map/Widgets/OATurnDrawable.h
```diff
@@ -9,6 +9,13 @@
 #import <UIKit/UIKit.h>
 #import "OATurnPathHelper.h"
 
+typedef NS_ENUM(NSInteger, EOATurnDrawableThemeColor)
+{
+    EOATurnDrawableThemeColorMap = 0,
```
tigrim, line 14: can use without 0

tigrim (approved):

PR #3539, opened 2024-03-29
Fix search large objects by basemap

Dima-1:
transfer from [Android request](https://github.com/osmandapp/OsmAnd/pull/19340)

### DIFF
--- a/Sources/Search/OASearchCoreFactory.mm
+++ b/Sources/Search/OASearchCoreFactory.mm
@@ -733,6 +733,7 @@
     
     searchCriteria->name = QString::fromNSString(searchWord);
     searchCriteria->xy31 = OsmAnd::PointI(bbox.centerX, bbox.centerY);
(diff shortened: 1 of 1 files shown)


Sources/Search/OASearchCoreFactory.mm
```diff
@@ -752,6 +751,10 @@ - (BOOL) search:(OASearchPhrase *)phrase resultMatcher:(OASearchResultMatcher *)
         if (!r)
             continue;
         searchCriteria->localResources = {r};
+        if ([resId containsString:@"basemap"])
+            searchCriteria->bbox31 = OsmAnd::AreaI(0, 0, INT_MAX, INT_MAX);
```
alex-osm, line 755: `bbox31` is `Nullable`, thus it is better to use this property instead of `OsmAnd::AreaI(0, 0, INT_MAX, INT_MAX)`

PR #3544, opened 2024-03-29
fix widget glide ratio

Skalii:


Sources/Controllers/Map/Widgets/Widgets/GlideTargetWidget.swift
```diff
@@ -116,7 +116,7 @@ final class GlideTargetWidget: GlideBaseWidget {
             markUpdated()
             if forceUpdate || metricSystemChanged || !GlideUtils.areAltitudesEqual(cachedTargetAltitude, targetAltitude) {
                 cachedTargetAltitude = targetAltitude
-                if let cachedTargetAltitude {
+                if let cachedTargetAltitude, cachedTargetAltitude != Self.minAltitudeValue {
```
alex-osm, line 119: Why `minAltitudeValue` is even needed? And why in this `if`?

PR #3552, opened 2024-04-02
Use uname() to get correct device model name

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2485

PR #3578, opened 2024-04-10
Glide ratio

DmitrySvetlichny:


Sources/Controllers/Map/Widgets/Widgets/GlideAverageWidget.swift
```diff
@@ -15,30 +15,59 @@ final class GlideAverageWidget: GlideBaseWidget {
...
-        section.headerText = localizedString("average_glide_ratio_time_interval_desc")
+        section.headerText = localizedString("shared_string_settings")
+
+        if let preference = widgetState?.getPreference() {
+            let settingRow = section.createNewRow()
+            settingRow.cellType = OAValueTableViewCell.reuseIdentifier
+            settingRow.key = "value_pref"
+            settingRow.title = localizedString("shared_string_mode")
+            settingRow.setObj(preference, forKey: "pref")
+            settingRow.setObj(getWidgetName() ?? localizedString("average_glide_ratio"), forKey: "value")
```
tigrim, line 67: When can getWidgetName be nil?

Sources/Controllers/Map/Widgets/Widgets/GlideTargetWidget.swift
```diff
@@ -116,11 +116,11 @@ final class GlideTargetWidget: GlideBaseWidget {
             markUpdated()
             if forceUpdate || metricSystemChanged || !GlideUtils.areAltitudesEqual(cachedTargetAltitude, targetAltitude) {
                 cachedTargetAltitude = targetAltitude
-                if let cachedTargetAltitude {
-                    let formattedAltitude: String = OAOsmAndFormatter.getFormattedAlt(cachedTargetAltitude)
+                if let targetAltitude = targetAltitude, targetAltitude != GlideTargetWidget.minAltitudeValue {
```
tigrim, line 119: use short style

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -2466,7 +2466,7 @@ - (void) createGeoTiffCollection
 - (void) recreateHeightmapProvider
 {
     OASRTMPlugin *plugin = (OASRTMPlugin *) [OAPluginsHelper getEnabledPlugin:OASRTMPlugin.class];
-    if (!plugin || ![plugin is3DMapsEnabled] || _app.data.terrainType == EOATerrainTypeDisabled)
```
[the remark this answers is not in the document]
DmitrySvetlichny: To display target height when 3D terrain is turned off and terrain settings are turned off

Sources/Controllers/Map/Widgets/Widgets/GlideAverageWidget.swift
DmitrySvetlichny: When widgetState is nil: If the widgetState is nil, the method attempts to return widgetType?.title. If widgetType itself is nil or if it is non-nil but its title property is not set, then the method will return nil.

tigrim (approved):

PR #3587, opened 2024-04-12
Add cleanupCollatorCache() after ICU calls

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2503

RZR-UA:
Will be replaced with ICU's inner cache cleanup.

PR #3606, opened 2024-04-18
Replace HH-options with single one use_old_routing

RZR-UA:


RZR-UA:
need to rebase into 4.7

PR #3607, opened 2024-04-18
Replace HH-options with single one use_old_routing (r4.7)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2513

Issue #3617, opened 2024-04-20
Route recalculation is broken in OsmAnd Version 4.7 


kevinp2:
### Description

This is a bad bug.  Route recalculation upon a detour is basically 100% broken in 4.7

### Steps to reproduce

Start a route and drive along it.  

At some point, make a detour from the route.

### Actual result

OSMAnd 4.7 simply does not recalculate the route!  Instead, the screen becomes as shown with 0 ft displayed in the upper left corner.  The map continues to pan but the route is not recalculated even after several minutes.  I tried this 3 times and it did not work even a single time.

<table style="width:100%">
  <tr>
<th><img src="https://github.com/osmandapp/OsmAnd/assets/609677/137b4299-e375-4b0a-950f-5cc01f4facf1"  width="300" height="auto"/> </td>
  </tr>
</table>

### Expected result

Route calculation should work upon a detour as it did prior to version 4.7

### Your Environment (required)

```
OsmAnd Version: 4.7
Android/iOS version: 17.4.1
Device model: iPhone 15 Pro Max
```

vshcherb:
We will take a looks and try to release release-patch next week

kevinp2:
> Navigation settings>Route parameters

I took a look but I could not find the option Routing type in this menu.  I don't know if it is available in the iOS app.

DmitryAlexei:
@kevinp2 Please enable OsmAnd development plugin (go to menu > plugins > OSmAnd development) and in plugin settings choose Old routing (A*) option

ivanPyrohivskyi:
Perhaps it is regression
https://github.com/osmandapp/OsmAnd-iOS/commit/83803856a2e746a22d7f18bb42d5a2d281514334

vshcherb:
Probably related https://github.com/osmandapp/OsmAnd-iOS/issues/3409

kevinp2:
I can confirm that this is fixed in OSMAnd 4.7.4  

I did several route detours today and all worked flawlessly.  Thanks for the fix!

vshcherb:
More details is needed to reproduce the issue, it might be another case

vshcherb:
Yes please open, please record track so later we could simulate it by track

PR #3634, opened 2024-04-23
Fix rare route recalculation problems (Issue #3617)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/issues/3617

PR #3663, opened 2024-05-02
fix my places tracks 5

Skalii:


Sources/SwiftExtensions/UIViewController+Extension.swift
```diff
@@ -74,4 +74,18 @@ extension UINavigationController {
         }
         return newCurrentHistory
     }
+
+    @objc func restoreForceHidingScrollableHud() {
+        if !viewControllers.isEmpty {
```
tigrim, line 79: guard

Sources/SwiftExtensions/UIViewController+Extension.swift
```diff
@@ -74,4 +74,18 @@ extension UINavigationController {
...
+            var newHistory: [UIViewController] = Array()
```
tigrim, line 80: `var newHistory = [UIViewController]()`

tigrim (approved):

PR #3677, opened 2024-05-07
fix favorite background size

Skalii:


Sources/Core/OACompoundIconUtils.mm
```diff
@@ -87,7 +87,7 @@ @implementation OACompoundIconUtils
         return result;
 
     // color filled background icon
-    NSString *backgroundIconName = [NSString stringWithFormat:@"ic_bg_point_%@_center%@", shapeName, sizeName];
+    NSString *backgroundIconName = [OANativeUtilities getScaledResourceName:[NSString stringWithFormat:@"ic_bg_point_%@_center%@", shapeName, sizeName]];
```
tigrim, line 90: What are these resources? It's better to transfer them and take them from the assets

tigrim (approved):

Sources/Core/OACompoundIconUtils.mm
```diff
@@ -87,7 +87,7 @@ @implementation OACompoundIconUtils
...
     UIImage *img = [self getIcon:backgroundIconName defaultIconName:@"ic_bg_point_circle_center"];
```
tigrim, line 91: ic_bg_point_circle_center scalefactor?

tigrim (approved):

PR #3704, opened 2024-05-14
Replace car play and my location raster resources

DmitrySvetlichny:


tigrim (commented):
1) You need to fix the conflict.
2) Do not need to make a copy of the resource for the main target and CarPlay, it is sufficient to check the 'Universal' checkbox for the image. If we were using different resources for the same image name in the Main application and CarPlay, then it would be necessary.

tigrim:
Resolved the conflict, still need to fix the second point.

tigrim (approved):

PR #3713, opened 2024-05-16
Disabling autocorrect and capitalization

DmitrySvetlichny:


Sources/Controllers/OsmEditing/OABasicEditingViewController.mm
```diff
@@ -269,6 +269,8 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
                 break;
             case 3:
                 cell.inputField.textView.keyboardType = UIKeyboardTypeURL;
+                cell.inputField.textView.autocapitalizationType = UITextAutocapitalizationTypeNone;
+                cell.inputField.textView.autocorrectionType = UITextAutocorrectionTypeNo;
```
tigrim, line 273: Do we need to re-enable the inputField.textView for steel cells?

PR #3717, opened 2024-05-17
hideContextMenu in mapSettingsButtonClick

DmitrySvetlichny:


Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -907,6 +907,8 @@ - (void) mapSettingsButtonClick:(id)sender mode:(OAApplicationMode *)targetMode
     [self createShadowButton:@selector(closeDashboard) withLongPressEvent:nil topView:_dashboard.view];
     
     [self.targetMenuView quickHide];
```
alex-osm, line 909: Is it needed then?

PR #3718, opened 2024-05-17
[GeoApproximation] Sync with Java, fix lost minPointApproximation, try geo-approximation as default

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2342

PR #3740, opened 2024-05-29
Sync OSM oauth2 scopes with Android

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19841

RZR-UA:
Android (master):

ok

iOS (before):

was

iOS (synced):

now

PR #3742, opened 2024-05-30
Fix travelguide lang prefix

Dima-1:


Sources/Controllers/TravelGuides/TravelGuidesUtils.swift
```diff
@@ -59,4 +59,8 @@ final class TravelGuidesUtils {
         }
         return nil
     }
+    
+    static func getTitleWithotPrefix(title: String) -> String {
+        return title.hasPrefix(EN_LANG_PREFIX) ? String(title.dropFirst(EN_LANG_PREFIX.count)) : title
```
tigrim, line 64: remove return

Sources/Controllers/TravelGuides/TravelGuidesUtils.swift
```diff
@@ -59,4 +59,8 @@ final class TravelGuidesUtils {
         }
         return nil
     }
+    
+    static func getTitleWithotPrefix(title: String) -> String {
```
tigrim, line 63: without

tigrim (approved):

PR #3766, opened 2024-06-09
Update Localizable.strings

flexmak:
I fixed a grammar mistake in the German localization file. I'm a native speaker.

"Öffnen" means to open. "Geöffnet" means that sth is open.


vshcherb:
It needs to be changed https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/res/values-de/strings.xml here cause it's derived resource

PR #3825, opened 2024-07-03
Weather redesign

tigrim:


Sources/Controllers/Map/Layers/OAWeatherContourLayer.mm
```diff
@@ -122,24 +122,40 @@ - (BOOL) updateLayer
...
+        OsmAnd::BandIndex band = WEATHER_BAND_NOTHING;
 
         OAMapStyleParameter *tempContourLinesParam = [_styleSettings getParameter:WEATHER_TEMP_CONTOUR_LINES_ATTR];
         OAMapStyleParameter *pressureContourLinesParam = [_styleSettings getParameter:WEATHER_PRESSURE_CONTOURS_LINES_ATTR];
         OAMapStyleParameter *cloudContourLinesParam = [_styleSettings getParameter:WEATHER_CLOUD_CONTOURS_LINES_ATTR];
         OAMapStyleParameter *windContourLinesParam = [_styleSettings getParameter:WEATHER_WIND_CONTOURS_LINES_ATTR];
         OAMapStyleParameter *precipContourLinesParam = [_styleSettings getParameter:WEATHER_PRECIPITATION_CONTOURS_LINES_ATTR];
-
+        
+        BOOL bandWasChanged = NO;
```
alex-osm, line 133: why `bandWasChanged` is needed? can it be replaced with code like `band != WEATHER_BAND_NOTHING`?

Sources/Controllers/Map/OAMapInfoController.h
```diff
@@ -65,5 +65,6 @@
 - (void)updateLayout;
 - (void)viewWillTransition;
 - (void)updateSpeedometer;
+- (UIMenu *)createContourMenu:(void(^)(void))onTapHandler;
```
alex-osm, line 68: This method should be in weather related class

PR #3831, opened 2024-07-04
fix favorite background size 2

Skalii:


Sources/Controllers/Map/Layers/OAImpassableRoadsLayer.mm
```diff
@@ -45,7 +46,8 @@ - (NSString *) layerId
 - (void) initLayer
 {
     [super initLayer];
-    
+    _textSize = [[OAAppSettings sharedManager].textSize get];
```
alex-osm, line 49: It may change. Where is it handled?

Sources/Core/OACompoundIconUtils.mm
```diff
@@ -132,39 +147,53 @@ @implementation OACompoundIconUtils
...
+    UIImage *img = [self getIcon:iconName
+                 defaultIconName:defaultResourceName ? [OANativeUtilities getScaledResourceName:defaultResourceName] : nil
+                           scale:scale];
+    if (img)
+    {
+        if (color)
+            img = [OAUtilities tintImageWithColor:img color:color];
+        result = [OANativeUtilities skImageFromCGImage:img.CGImage];
+    }
+    return [OANativeUtilities getScaledSkImage:result scaleFactor:scale];
```
alex-osm, line 185: img is already scaled. Why apply scale factor again?

Sources/Core/OACompoundIconUtils.mm
```diff
@@ -132,39 +147,53 @@ @implementation OACompoundIconUtils
...
 {
     UIImage *iconImage = [UIImage imageNamed:iconName];
-    if (!iconImage)
+    if (!iconImage && defaultIconName)
         iconImage = [UIImage imageNamed:defaultIconName];
-    return iconImage;
+    if (!iconImage)
+        return nil;
+    CGSize iconPtSize = { iconImage.size.width * scale, iconImage.size.height * scale };
+    return [OAUtilities resizeImage:iconImage newSize:iconPtSize];
```
alex-osm, line 196: If scale = 1.0, why call resizeImage?

PR #3837, opened 2024-07-11
Increase simple widget size

tigrim:


Sources/Controllers/Settings/WidgetSettings/WidgetConfigurationViewController.swift
```diff
@@ -205,6 +205,7 @@ class WidgetConfigurationViewController: OABaseButtonsViewController, WidgetStat
                     NotificationCenter.default.post(name: .SimpleWidgetStyleUpdated,
                                                     object: widgetInfo,
                                                     userInfo: nil)
+                    OARootViewController.instance().mapPanel.recreateControls()
```
alex-osm, line 208: Do not do such calls directly. Add observer to OAMapPanelViewController.

PR #3838, opened 2024-07-11
Port interpolateEmptyElevationWpts from Java (fix charts elevation)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/19142

PR #3870, opened 2024-07-26
The route name is displayed in the selected map language

DmitrySvetlichny:


Sources/Controllers/Map/OARouteKey.mm
```diff
@@ -64,4 +65,12 @@ - (NSString *)getActivityTypeTitle
     return [res isEqualToString:resourceId] ? [OAUtilities capitalizeFirstLetter:tag] : res;
 }
 
+- (NSString *)getLocalizedTitle
```
alex-osm, line 68: Hold localized title in class field and resolve it in initWithKey once only

PR #3878, opened 2024-07-30
color palette updates

Skalii:


Sources/Helpers/ConcurrentDictionary.swift
```diff
@@ -43,5 +43,11 @@ class ConcurrentDictionary<Key: Hashable, Value> {
         defer { pthread_rwlock_unlock(&rwlock) }
         return Array(dictionary.values)
     }
+
+    func getAllKeys() -> [Key] {
+        pthread_rwlock_rdlock(&rwlock)
```
tigrim, line 48: Why do we need `pthread_rwlock` here?

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,88 @@
...
+            debugPrint("Directory contents changed: \(self.type.path)")
+            NotificationCenter.default.post(name: self.type.notificationName, object: self.type.path)
+        }
+
+        // Set the cancel handler
+        dispatchSource?.setCancelHandler { [weak self] in
+            guard let self else { return }
+
+            close(self.directoryFileDescriptor)
+            self.directoryFileDescriptor = -1
```
tigrim, line 74: dispatchSource = nil ?

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,88 @@
...
+        guard directoryFileDescriptor != -1 else {
+            debugPrint("Unable to open directory: \(type.path)")
+            return
+        }
+
+        // Create the dispatch source
+        dispatchSource = DispatchSource.makeFileSystemObjectSource(
+            fileDescriptor: directoryFileDescriptor,
+            eventMask: .all,
+            queue: DispatchQueue.global()
```
tigrim, line 58: create self concurrent DispatchQueue (private let)

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,88 @@
...
+    var notificationName: NSNotification.Name {
+        switch self {
+        case .colorPalette:
+            return NSNotification.Name("ColorPaletteDicrectoryUpdated")
+        }
+    }
+}
+
+@objcMembers
+class DirectoryObserver: NSObject {
```
tigrim, line 30: final

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,88 @@
...
+        // Start monitoring
+        dispatchSource?.resume()
+    }
+
+    func stopObserving() {
+        dispatchSource?.cancel()
+    }
+
+    deinit {
+        stopObserving()
```
tigrim, line 86: I don’t think it’s necessary; the object will be deleted anyway.

Sources/Plugins/SRTMPlugin/TerrainMode.swift
```diff
@@ -49,6 +49,8 @@ final class TerrainMode: NSObject {
     static let colorSlopePrefix = "slope_"
     static let heightPrefix = "height_"
 
+    private static let queue = DispatchQueue(label: "TerrainModeValues")
```
tigrim, line 52: TerrainModeValuesQueue

Sources/Plugins/SRTMPlugin/TerrainMode.swift
```diff
@@ -150,6 +168,10 @@ final class TerrainMode: NSObject {
         type == .slope
     }
 
+    func isHeight() -> Bool {
```
tigrim, line 171: unused?

tigrim (approved):

Sources/Controllers/DashboardOnMap/MapSettings/OAMapSettingsTerrainParametersViewController.mm
```diff
@@ -121,13 +121,67 @@ - (void)commonInit
...
 
+- (void)onColorPalettesFilesUpdated:(NSNotification *)notification
+{
+    if (![notification.object isKindOfClass:NSDictionary.class] || _terrainType != EOATerrainSettingsTypePalette)
+        return;
+
+    NSDictionary<NSString *, NSString *> *colorPaletteFiles = (NSDictionary *) notification.object;
+    if (!colorPaletteFiles)
+        return;
+    NSString *currentPaletteFile = [_terrainMode getMainFile];
```
alex-osm, line 144: Looks like this code run in background thread. Are you sure it is thread safe?

Sources/Controllers/Map/Layers/OAGPXLayer.mm
```diff
@@ -137,6 +144,33 @@ - (BOOL) updateLayer
...
+
+    NSDictionary<NSString *, NSString *> *colorPaletteFiles = (NSDictionary *) notification.object;
+    if (!colorPaletteFiles)
+        return;
+    BOOL refresh = NO;
+    for (NSString *colorPaletteFile in colorPaletteFiles)
+    {
+        if ([colorPaletteFile hasPrefix:ColorPaletteHelper.routePrefix])
+        {
+            _updatedColorPaletteFiles[colorPaletteFile] = colorPaletteFiles[colorPaletteFile];
```
alex-osm, line 160: Is it thread safe?

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -3683,7 +3683,6 @@ - (void) refreshGpxTracks
 {
     @synchronized(_rendererSync)
     {
-        [_mapLayers.gpxMapLayer resetLayer];
```
alex-osm, line 3686: What reason of dropping this code?

Sources/Plugins/SRTMPlugin/TerrainMode.swift
```diff
@@ -49,6 +49,8 @@ final class TerrainMode: NSObject {
     static let colorSlopePrefix = "slope_"
     static let heightPrefix = "height_"
 
+    private static let queue = DispatchQueue(label: "TerrainModeValuesQueue")
```
alex-osm, line 52: It is not clear why synchronization is needed

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,80 @@
...
+//  DirectoryObserver.swift
+//  OsmAnd Maps
+//
+//  Created by Skalii on 26.07.2024.
+//  Copyright © 2024 OsmAnd. All rights reserved.
+//
+
+import Foundation
+
+@objc enum DirectoryObserverType: Int {
```
alex-osm, line 11: DirectoryObserver should be path based, not specific type. Drop DirectoryObserverType.

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,80 @@
...
+        }
+    }
+}
+
+@objcMembers
+final class DirectoryObserver: NSObject {
+
+    static let updatedKey = "updated"
+    static let deletedKey = "deleted"
+    static let createdKey = "created"
```
alex-osm, line 34: Not directly related to DirectoryObserver. Drop.

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    private let type: DirectoryObserverType
+
+    private var directoryFileDescriptor: CInt = -1
+    private var dispatchSource: DispatchSourceFileSystemObject?
+    private let queue = DispatchQueue(label: "DirectoryObserverQueue", attributes: .concurrent)
```
alex-osm, line 40: Isn't global queue enough?

Sources/Helpers/DirectoryObserver.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    init(_ type: DirectoryObserverType) {
+        self.type = type
+    }
+
+    func startObserving() {
```
alex-osm, line 46: Why you dropped methods that were in the initial test code? `stopObserving`, `deinit` ?

import Foundation

class DirectoryObserver {
    private var directoryFileDescriptor: CInt = -1
    private var dispatchSource: DispatchSourceFileSystemObject?
    private let directoryURL: URL
    
    init(directoryURL: URL) {
        self.directoryURL = directoryURL
    }
    
    func startObserving() {
        // Open the directory
        directoryFileDescriptor = open(directoryURL.path, O_EVTONLY)
        
        guard directoryFileDescriptor != -1 else {
            print("Unable to open directory: \(directoryURL.path)")
            return
        }
        
        // Create the dispatch source
        dispatchSource = DispatchSource.makeFileSystemObjectSource(
            fileDescriptor: directoryFileDescriptor,
            eventMask: .all,
            queue: DispatchQueue.global()
        )
        
        // Set the event handler
        dispatchSource?.setEventHandler { [weak self] in
            guard let self = self else { return }
            print("Directory contents changed")
            self.handleDirectoryChange()
        }
        
        // Set the cancel handler
        dispatchSource?.setCancelHandler { [weak self] in
            guard let self = self else { return }
            close(self.directoryFileDescriptor)
            self.directoryFileDescriptor = -1
        }
        
        // Start monitoring
        dispatchSource?.resume()
    }
    
    func stopObserving() {
        dispatchSource?.cancel()
    }
    
    private func handleDirectoryChange() {
        // Implement your logic to handle directory changes
        print("Handle the changes here")
    }
    
    deinit {
        stopObserving()
    }
}

// Usage
let directoryURL = URL(fileURLWithPath: "/path/to/your/directory")
let observer = DirectoryObserver(directoryURL: directoryURL)
observer.startObserving()

Sources/Helpers/DirectoryObserver.swift
tigrim: I propose that we use our own queue for this purpose. It will function as a separate service. This will prevent the situation where the global queue might be occupied when we expect to receive a notification about a change in the directory

PR #3879, opened 2024-07-31
Fix NO_DIRECTION bug in iOS (change degrees to radians)

RZR-UA:


RZR-UA:
already fixed in r4.8

PR #3935, opened 2024-08-15
fix color palette screen

Skalii:


Sources/Helpers/Appearance/ColorsCollection.swift
```diff
@@ -196,13 +196,29 @@ class ColorsCollection: NSObject {
...
+    func getSorting() -> ((PaletteColor, PaletteColor) -> Bool) {
+        {
+            return $0.getIndex() < $1.getIndex()
+        }
+    }
+
+    func sortColors() {
+        originalOrder.sort(by: getSorting())
+
+        for paletteColor in originalOrder {
```
tigrim, line 208: Does this code do the same thing?
```
for (index, paletteColor) in originalOrder.enumerated() {
        paletteColor.setIndex(index + 1)
   }  
```

Sources/Helpers/Appearance/GradientColorsCollection.swift
```diff
@@ -219,12 +231,43 @@ final class GradientColorsCollection: ColorsCollection {
         return res
     }
 
-    override func saveColors() {
-        for paletteColor in originalOrder {
-            if let index = originalOrder.firstIndex(where: { $0.id == paletteColor.id }) {
-                paletteColor.setIndex(index + 1)
+    override func getSorting() -> ((PaletteColor, PaletteColor) -> Bool) {
+        { [weak self] in
```
tigrim, line 235: Why is [weak self] used here? I think it can be removed

tigrim (approved):

PR #3945, opened 2024-08-22
Size Zero GPX Fixed

DmitrySvetlichny:


tigrim (approved):

Sources/Controllers/TargetMenu/GPX/TrackMenu/OATrackMenuHudViewController.mm
```diff
@@ -1338,7 +1338,7 @@ - (NSString *)getDirName
 - (NSString *)getGpxFileSize
 {
     NSDictionary *fileAttributes = [NSFileManager.defaultManager attributesOfItemAtPath:self.isCurrentTrack
-            ? self.gpx.gpxFilePath : self.doc.path error:nil];
+            ? self.gpx.gpxFilePath : self.gpx.absolutePath error:nil];
```
alex-osm, line 1341: Why use self.gpx.gpxFilePath for self.isCurrentTrack? Does it make sense?

PR #3999, opened 2024-09-10
Fix iOS build for GpxRouteApproximation separated from RoutePlannerFrontEnd

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20280

Sources/Router/OARouteProvider.mm
```diff
@@ -975,7 +975,7 @@ - (OARoutingEnvironment *) getRoutingEnvironment:(OAApplicationMode *)mode start
             return true;
         };
 
-        // env.router->setUseGeometryBasedApproximation(true); // r4.8 is switched to old approximation
+        env.router->setUseGeometryBasedApproximation(true);
```
ivanPyrohivskyi, line 978: ?

RZR-UA: got back

PR #4005, opened 2024-09-12
change compass behaviour

Skalii:


Sources/Controllers/Map/OAMapHudViewController.mm
```diff
@@ -1643,4 +1666,17 @@ - (void) widgetsLayoutDidChange:(BOOL)animated
     [self updateControlsLayout:animated];
 }
 
+#pragma mark - UIGestureRecognizerDelegate
+
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldReceiveTouch:(UITouch *)touch
```
tigrim, line 1671: should this only apply to the Recognizer compass button?

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -1250,6 +1250,9 @@ - (void) zoomAndRotateGestureDetected:(UIGestureRecognizer *)recognizer
     if (_rotationAnd3DViewDisabled && rotationRecognizer)
         return;
 
+    if (rotationRecognizer && [[OAAppSettings sharedManager].rotateMap get] == ROTATE_MAP_NONE)
```
tigrim, line 1253: Will pinch zoom work on Mac(touchpad)/iOS?

PR #4011, opened 2024-09-16
Show if POI is closed or Open

DmitrySvetlichny:


Sources/POI/OAPOIHelper.mm
```diff
@@ -550,6 +552,49 @@ - (NSString *) getPoiStringWithoutType:(OAPOI *)poi
     return [NSString stringWithFormat:@"%@ %@", typeName, n];
 }
 
+- (NSString *) getOpeningHoursStatusChange:(OAPOI *)poi
```
alex-osm, line 555: getFormattedOpeningHours

PR #4024, opened 2024-09-24
fix crash after import osf

Skalii:


tigrim (approved):

Sources/Controllers/Panels/OAOptionsPanelBlackViewController.m
```diff
@@ -94,6 +94,15 @@ - (void) viewWillAppear:(BOOL)animated
                                                         andObserve:[OsmAndApp instance].data.weatherChangeObservable];
 }
 
+- (void)dealloc
```
alex-osm, line 97: Why detach in dealloc when was attached in viewWillAppear? Use viewWillDisappear in this case.

Sources/Controllers/Panels/OAOptionsPanelBlackViewController.m
```diff
@@ -108,6 +117,9 @@ - (void)updateMenu
 
 - (void) updateLayout
 {
+    if (!_menuButtonsArray) // if class is not initialized
```
alex-osm, line 120: Do not use indirect sign of class initialized or not. It even not clear what is it - "initialized"
Looks like `self.isViewLoaded` is needed here

PR #4030, opened 2024-09-27
Set GpxMultiSegmentsApproximation as default (4.9)

RZR-UA:
This reverts commit a201754cdd63573d41f9e8b0039db26075517c0a.


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/20280

PR #4038, opened 2024-10-02
fix map mode screen

Skalii:


Sources/Controllers/DashboardOnMap/MapSettings/MapSettingsMapModeParametersViewController.swift
```diff
@@ -171,20 +167,24 @@ final class MapSettingsMapModeParametersViewController: OABaseScrollableHudViewC
...
+                if let image = UIImage(named: $0.iconName) {
+                    return OAUtilities.resize(image, newSize: CGSize(width: 20, height: 20))
+                }
+                return UIImage()
+            }),
+            "selectedIcons": modes.map({
+                if let image = UIImage(named: $0.selectedIconName) {
+                    return OAUtilities.resize(image, newSize: CGSize(width: 20, height: 20))
+                }
+                return UIImage()
```
tigrim, line 185: Duplicated code can be extracted into a separate function

tigrim (approved):

PR #4042, opened 2024-10-04
fix turn after recalculation in carplay 2

Skalii:


alex-osm:
Looks like synchronization with java code is needed
https://github.com/osmandapp/OsmAnd/blob/02d91f35bc555fddf402053f35413d599df2a828/OsmAnd/src/net/osmand/plus/auto/TripHelper.java#L100

PR #4066, opened 2024-10-16
fix terrain style screens

Skalii:


Sources/Controllers/Cells/Handlers/PaletteCollectionHandler.swift
```diff
@@ -181,15 +181,13 @@ final class PaletteCollectionHandler: OABaseCollectionHandler {
         imageView.gradated(Self.createGradientPoints(colorPalette))
     }
 
-    @objc static func createDescriptionForPalette(_ colorPalette: ColorPalette) -> String {
-        let formatter = NumberFormatter()
-        formatter.numberStyle = .decimal
-        formatter.minimumFractionDigits = 0
-        formatter.maximumFractionDigits = 3
-        formatter.locale = Locale(identifier: "en_US_POSIX")
+    @objc static func createDescriptionForPalette(_ colorPalette: ColorPalette, isTerrain: Bool) -> String {
```
tigrim, line 184:
```
   colorPalette.colorValues
        .map { isTerrain ? GradientUiHelper.formatTerrainTypeValues($0.val) : "\($0.val)" }
        .joined(separator: " • ")
```
Check if it’s possible to replace compactMap with map

tigrim (approved):

PR #4093, opened 2024-11-05
fix import for upd UI, fix upd visible tracks

tigrim:


tigrim:
1)When navigating from "My Places" -> "Tracks" -> "Visible Tracks", perform a long tap to open the context menu and select "Open". Then, go to the "Action" tab and choose: delete/rename/change folder/duplicate. After that, go back to the "Visible Tracks" screen -> "Tracks". On these screens, the new UI with the modified track should not be displayed.

2)Added the ability to update the UI "on the fly" when importing a track via the system folder "Files", using the context menu, "Share". Now, if we are on the screen with tracks or in the "Import" subfolder and we import from "Files", the UI updates immediately.

PR #4180, opened 2024-12-09
Fix shields and destination

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/21362
Transfer PR https://github.com/osmandapp/OsmAnd/pull/21476
Depends from https://github.com/osmandapp/OsmAnd-core-legacy/pull/264 and https://github.com/osmandapp/OsmAnd-core/pull/811

### DIFF
--- a/Sources/Controllers/Map/Widgets/OATopTextView.mm
+++ b/Sources/Controllers/Map/Widgets/OATopTextView.mm
@@ -548,6 +548,23 @@
         }
     }
     
+    //Change several shields to one
+    if (streetName.shields.count > 1)
+    {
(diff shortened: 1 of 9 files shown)


Sources/Controllers/Map/Widgets/OATopTextView.mm
```diff
@@ -548,6 +548,23 @@ - (BOOL) updateInfo
         }
     }
     
+    //Change several shields to one
+    if (streetName.shields.count > 1)
```
ivanPyrohivskyi, line 552: According to @vshcherb comment
Problem described in [the comment to issue](https://github.com/osmandapp/OsmAnd/issues/21362#issuecomment-2493801878)

PR #4198, opened 2024-12-12
fix direction from

Skalii:


tigrim (approved):

alex-osm:
We need confirmation that the changes in the request correspond to the logic requested in the issue.

Skalii:
<table style="width:100%">
  <tr>
    <th>Android</th>
    <th>iOS</th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
  </tr>
</table>

Sources/Controllers/RouteInfoMenu/OARouteInfoView.h
```diff
@@ -18,6 +18,7 @@
 @property (weak, nonatomic) IBOutlet UIButton *goButton;
 @property (weak, nonatomic) IBOutlet UIView *contentContainer;
 @property (weak, nonatomic) IBOutlet UIView *statusBarBackgroundView;
+@property (nonatomic) BOOL selectFromMapTouch;
```
alex-osm, line 21: Not clear why this field added. It's name is confusing. Looks like it should be dropped.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -821,6 +821,12 @@ - (void) closeRouteInfo
     [self closeRouteInfo:YES onComplete:nil];
 }
 
+- (void)closeRouteInfoByTappingOnMap
```
alex-osm, line 824: Why you handle close by tapping. What if menu closed by cancel button?

PR #4239, opened 2024-12-23
Adapt ios code to new class Link (OsmAnd-shared)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-Issues/issues/2277

Issue #4249, opened 2024-12-26
Imported track with multiple segments:segment are joined by a straight line


Baldo81:
iOS: 18.2
Device: iPhone 13 and iPad Pro 11 M4 (issue present on both devices)
OsmAnd: 4.9.1 (issue started from 4.9, on 4.8 there was no issue with segments)

Issue: when I import a gpx with multiple segments, the end of a segment is joined with the beginning of another segment even if the option “join segments” in APPEARANCE  is switched off. 

Screenshot of an example:
on OsmAnd you can see the straight line that joins the 2 segments

<table style="width:100%">
  <tr>
<th><img src="https://github.com/user-attachments/assets/60a44730-7761-45c5-925a-2b7c9540d534"  width="500" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/09b77e1e-2af4-456f-b36f-ab04eb13fc8a"  width="500" height="auto"/> </th>
  </tr>
</table>

Attached the gpx as an example: 
[New file 1.zip](https://github.com/user-attachments/files/18255863/New.file.1.zip)

Baldo81:
Adding some link of people noticing the same issue:

https://m.facebook.com/groups/Transeurotrail.org/posts/hi-tet-rider-does-anyone-know-why-the-tet-gpx-in-osmand-for-ios-is-having-those-/2957451121076679/

https://www.reddit.com/r/OsmAnd/s/dsopezVLxL

https://m.facebook.com/groups/Transeurotrail.org/permalink/2960891577399300/?

dmpr0:
## Description

### How to reproduce
* Menu – Plan a Route – Create new route
* Add two points on the map
* Tap Options – Start new segment
* Add two points 
* Save track

### Actual result

<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/0c0763e2-ed22-477f-9832-dd5b9bc7228d
"  width="300" height="auto"/> </td>
  </tr>
</table>   

* Connected segments

### Expected result

* Segments not connected

**Environment**
OsmAnd Version: 4929
iOS version: 18.2
Device model: iPhone 16 Pro Max

PR #4263, opened 2024-12-30
Fix weird staight lines (tracksWithoutGeneralTrack) for Issue #4249

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/issues/4249

PR #4299, opened 2025-01-15
Fixes cloud bundled assets4.9

vshcherb:


Sources/Backup/BackupUtils.swift
```diff
@@ -63,23 +63,44 @@ final class BackupUtils: NSObject {
                                             infoFiles: Bool) -> [OARemoteFile: OASettingsItem] {
         var res = [OARemoteFile: OASettingsItem]()
```
vshcherb, line 64: https://github.com/osmandapp/OsmAnd/pull/21725

PR #4410, opened 2025-03-05
Used animateZoomToAndPan for double click animation

ihoryakovenko:


Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -1011,9 +1012,10 @@ - (void) storeTargetPosition:(UIGestureRecognizer *)recognizer
 
 - (void) restorePreviousTarget
 {
-    if ([self isTargetChanged] && [self isLastMultiGesture])
+    if ([self isTargetChanged] && ([self isLastMultiGesture] || _zoomInTargetChanged))
```
alex-osm, line 1015: Not clear why `_zoomInTargetChanged` needed. `isTargetChanged` should be enough.

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -1480,6 +1482,9 @@ - (void) zoomInGestureDetected:(UITapGestureRecognizer *)recognizer
     // Get base zoom delta
     float zoomDelta = [self currentZoomInDelta];
     
+    _zoomInTargetChanged = YES;
+    [self storeTargetPosition:recognizer];
```
alex-osm, line 1486: It seems `storeTargetPosition` / `restorePreviousTarget` should be called in `zoomInGestureDetected` method depending on recognizer.state, like in `moveGestureDetected` (`UIGestureRecognizerStateBegan` / `UIGestureRecognizerStateEnded` / `UIGestureRecognizerStateCancelled`)

Sources/Controllers/Map/OAMapViewController.mm
ihoryakovenko: I added _zoomInTargetChanged to distinguish new restorePreviousTarget from the old one. Originally, it was designed to work during one gesture call, but now I save the target in one gesture and restore it in a different gesture call. So, isLastMultiGesture is NO when I call restorePreviousTarget.
As an alternative, I can create a separate restorePreviousTarget for this case.

alex-osm: Now it is clear. Separate restorePreviousTarget not needed.

Sources/Controllers/Map/OAMapViewController.mm
ihoryakovenko: It would be good, but I tested it, and it is not true in this case. moveGestureDetected is called when we touch and remove our finger from the screen and has different states. zoomInGestureDetected is called only once if we double-click and has the UIGestureRecognizerStateChanged state.

PR #4464, opened 2025-03-31
Fix double-rte after exportGpx, sync addToTrack with Java

RZR-UA:


RZR-UA:
https://github.com/osmandapp/web/issues/711#issuecomment-2766013795

PR #4475, opened 2025-04-03
Favorites screen crash after freeze

tigrim:


Sources/Controllers/MyPlaces/OAFavoriteListViewController.mm
```diff
@@ -294,114 +295,131 @@ -(CGFloat) getToolBarHeight
...
+        return;
+    
+    CLLocationDirection newHeading = app.locationServices.lastKnownHeading;
+    CLLocationDirection newDirection =
+    (newLocation.speed >= 1 /* 3.7 km/h */ && newLocation.course >= 0.0f)
+    ? newLocation.course
+    : newHeading;
+    
+    __weak __typeof(self) weakSelf = self;
+    dispatch_barrier_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
```
alex-osm, line 326: it is not okay to use dispatch_barrier_async with any of the global concurrent queues obtained using dispatch_get_global_queue

PR #4501, opened 2025-04-10
Cloud time many digits after dot

DmitrySvetlichny:


Sources/Backup/BackupUiUtils.swift
```diff
@@ -52,7 +52,7 @@ final class BackupUiUtils: NSObject {
         guard time > 0 else {
```
tigrim, line 52: 1)Make getFormattedPassedTime private
2)Use a clear name for the input parameter to indicate it expects seconds

PR #4510, opened 2025-04-11
switch cycle routes defaults to previous selected

Vladyslav-lys:


Sources/QuickAction/Actions/ShowHideCycleRoutesAction.swift
```diff
@@ -40,17 +40,18 @@ final class ShowHideCycleRoutesAction: BaseRouteQuickAction {
...
-        let newValue = isEnabled() ? "false" : "true"
+        let isEnabled = isEnabled()
         let styleSettings = OAMapStyleSettings.sharedInstance()
         let routesParameter = styleSettings?.getParameter(SHOW_CYCLE_ROUTES_ATTR)
         let nodeParameter = styleSettings?.getParameter(CYCLE_NODE_NETWORK_ROUTES_ATTR)
+        let cycleRoutesParameter = OAAppSettings.sharedManager().cycleRoutesParameter.get()
         
         if let nodeParameter {
-            nodeParameter.value = newValue
+            nodeParameter.value = isEnabled ? "false" : cycleRoutesParameter
```
tigrim, line 50: If you use cycleRoutesParameter.get() in the if, then declare it here as well

Vladyslav-lys: fixed

tigrim (approved):

PR #4519, opened 2025-04-16
Fix display c_ icons (c_mx_natural_tree_broadleaved, etc)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/22045

PR #4525, opened 2025-04-17
Show map object tags in original order

Andrewschenko:
Don't change the order of tags of map objects


RZR-UA:
https://github.com/osmandapp/OsmAnd-iOS/pull/4537 this request will be included there

PR #4527, opened 2025-04-18
4225 Suggest to download map in Search 1/2

semenag01:


Sources/Controllers/TargetMenu/Search/OAQuickSearchListItem.h
```diff
@@ -19,7 +19,8 @@ typedef NS_ENUM(NSInteger, EOAQuickSearchListItemType)
     BUTTON,
     SEARCH_MORE,
     EMPTY_SEARCH,
-    ACTION_BUTTON
+    ACTION_BUTTON,
+    RESOURCE_ITEM
```
tigrim, line 23: Is this really necessary? I'm looking at the Android code — they use INDEX_ITEM in [ObjectType.java](https://github.com/osmandapp/OsmAnd/pull/13490/files#diff-12d7e6704152aea8d9c35a66a5f0e22009c5310b2aa1f7aa2ef1ab21a404e129), and everything else is based on that. I think we should have a similar implementation.

Sources/Controllers/TargetMenu/Search/OAQuickSearchResourceItem.h
```diff
@@ -0,0 +1,12 @@
+#import "OAQuickSearchListItem.h"
```
tigrim, line 1: We're writing the new class in Swift

Sources/Controllers/TargetMenu/Search/OAQuickSearchTableController.mm
```diff
@@ -885,11 +886,62 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
                 [cell.titleLabel setText:[item getName]];
             }
             return cell;
+        } else if ([item getType] == RESOURCE_ITEM)
+        {
+            OASimpleTableViewCell *cell;
```
tigrim, line 891: I think it’s better to use cells through a helper. For example, you can create your own class if needed:
`final class SearchDownloadingCellResourceHelper: DownloadingCellResourceHelper  {}`

Sources/Controllers/TargetMenu/Search/OAQuickSearchResourceItem.h
semenag01: To rewrite to Swift, I need to add some imports to Bridging-Header.h. After adding imports, the project does not compile.

Sources/Controllers/TargetMenu/Search/OAQuickSearchTableController.mm
semenag01: I made it consistent with the rest of the code.

PR #4534, opened 2025-04-22
Crash fix 5.0.1

tigrim:


klausr64:
FYI, with build 5019 the app keeps crashing when ending or saving a recording.

PR #4542, opened 2025-04-25
[5.0] Fix lost name in list of rendered objects (within)

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21993

tigrim (approved):

PR #4560, opened 2025-04-29
[5.0] Hide non-translatable tags and lonesome name list

RZR-UA:


RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/21742

Sources/Controllers/TargetMenu/POI/OAPOIViewController.mm
```diff
@@ -342,10 +342,14 @@ - (void) buildRows:(NSMutableArray<OARowInfo *> *)rows
         id value = dic[key];
         if ([value isKindOfClass:[NSDictionary class]]) {
             vl = value[@"name"] ?: @"";
+            id localization = value[@"localization"];
```
tigrim, line 345: Why use id? Isn't "localization" always an NSDictionary?
It's better to write it like this:
```
NSDictionary *localization = value[@"localization"];
if (vl.length > 0 && localization.count == 1)
    continue;
```

RZR-UA: 34d3bf917

tigrim (approved):

PR #4596, opened 2025-05-16
Set label offset for attached marker

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd-Issues/issues/1587

Issue #4414, opened 2025-03-06
Western part of Malaysia is not displayed as part of the relation


Zirochkabila:
### Description

When loading the map of Malaysia, its western part is not displayed as part of the [relation](https://www.openstreetmap.org/relation/2108121) (highlighted in green)
After loading, the western part of Malaysia is available for detailed viewing, navigation, etc., but still not displayed as part of the relation

### Steps to reproduce

1. Go to `2,26966, 102,29244` (Western part) - Map cannot be downloaded - the menu with the download option does not appear
2. Go to `2.44558/112.83487` (Eastern part) - A menu appears with a suggestion to download a map of Malaysia, but the western part of the country is not displayed as Malaysia

### Actual result

<details><summary>Eastern part of Malaysia</summary>
<img width="400" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/ff171c38-e654-42b5-8310-5bb6e686363e">
</details> 

### Expected result

<details><summary>Android</summary>
<img width="400" alt="Screenshot 2023-01-31 at 11 19 41" src="https://github.com/user-attachments/assets/10a61d53-90f4-46a2-bf22-06661e25d3ae">
</details> 

### Your Environment
OsmAnd Version: 4986

yuriiurshuliak:
**Summary:** The western part of Malaysia is not recognized as part of the country's relation in OsmAnd. As a result, there is no option to download it from the app.

The bug has been reproduced. Reproduction Steps:

1. **Go to coordinates:** `2.26966, 102.29244` (Western Malaysia)
   - **Issue:** The map cannot be downloaded; no download option appears.
   
2. **Go to coordinates:** `2.44558, 112.83487` (Eastern Malaysia)
   - **Observation:** A menu appears suggesting to download the Malaysia map.
   - **Issue:** The western region is missing from the map’s relation.

OsmAnd  4.9.5(4991)

<table style="width:100%">
  </tr>
  <tr>
    <th>Western Malaysia</th>
    <th>Eastern Malaysia</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/1ac76c89-cdf4-43d8-9a0d-a3b2cc57f6a0"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/5e2c2015-dcd5-496f-97f5-b9ebf65eda1b"  width="300" height="auto"/> </th>
  </tr>
</table>

[...]

PR #4602, opened 2025-05-21
Used additional poligons data to calculate world region BBOX

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd-iOS/issues/4414

PR #4655, opened 2025-06-12
navbar is removed fixed

Vladyslav-lys:


Vladyslav-lys:
https://github.com/user-attachments/assets/1206acf3-cbfa-4ed2-9273-cbcd06295cfe

Sources/Views/OATargetPointView.mm
```diff
@@ -413,9 +413,9 @@ - (void) updateBottomToolbarFrame:(BOOL)landscape
     }
 }
 
-- (void) showTopToolbar:(BOOL)animated
+- (void) showTopToolbar:(BOOL)animated forceToShowIfTypeFloating:(BOOL)forceToShowIfTypeFloating
```
tigrim, line 416: And what are the calls [self showTopToolbar:YES] and [self showTopToolbar:NO] supposed to do now?

tigrim:
Git branch name using underscores

Sources/Views/OATargetPointView.mm
Vladyslav-lys: these calls are supposed to the same as before

PR #4660, opened 2025-06-13
Update OASettingsItem.mm

Vladyslav-lys:


Sources/Backup/LocalBackup/SettingsItems/OASettingsItem.mm
```diff
@@ -346,6 +347,19 @@ - (BOOL) readFromFile:(NSString *)filePath error:(NSError * _Nullable *)error
     return YES;
 }
 
+//the method reads volumeUnitsChangedManually value from dictionary separetly and remove it
+- (void)readVolumeUnitsChangedManuallyFrom:(NSMutableDictionary<NSString *, NSString *> *)settings
```
tigrim: No, it should be a separate method that registers the settings on which the other settings depend. It should return an NSDictionary

Vladyslav-lys: fixed

Sources/Backup/LocalBackup/SettingsItems/OASettingsItem.mm
```diff
@@ -346,6 +347,25 @@ - (BOOL) readFromFile:(NSString *)filePath error:(NSError * _Nullable *)error
...
+        NSString *value = [mutableSettings objectForKey:key];
+        
+        if (!value)
+            continue;
+        
+        [self.item readPreferenceFromJson:key value:value];
+        [mutableSettings removeObjectForKey:key];
+    }
+    
+    return mutableSettings;
```
tigrim: copy?

Vladyslav-lys: fixed

tigrim (approved):

Vladyslav-lys:
Check if the migration works correctly:
1. Remove the app
2. Switch to master branch, change driving region and unit of volume
3. Switch to the branch with fixes, look at the values in the settings if they're correct

PR #4661, opened 2025-06-13
Navbar is removed

Vladyslav-lys:


Sources/Views/OATargetPointView.mm
```diff
@@ -629,7 +629,7 @@ - (void) prepareForRotation:(UIInterfaceOrientation)toInterfaceOrientation
 {
     if ([self isLandscapeSupported] && [OAUtilities isLandscape:toInterfaceOrientation])
     {
-        [self showTopToolbar:NO];
+        [self showTopToolbar:NO forceToShowIfTypeFloating:NO];
```
tigrim: Create a separate private method for hiding

Vladyslav-lys: fixed

PR #4741, opened 2025-07-21
Implement Map Zoom using MouseWheel (MacOS build)

RZR-UA:


Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -417,7 +419,16 @@ - (void) commonInit
     _grZoomOut.delegate = self;
     _grZoomOut.numberOfTapsRequired = 2;
     _grZoomOut.numberOfTouchesRequired = 2;
-    
+
+    // - MouseWheelScroll gesture (run under MacOS)
+    _grMouseWheelScroll = [[UIPanGestureRecognizer alloc] initWithTarget:self
+                                                                  action:@selector(mouseWheelScrollGuestureDetected:)];
+    _grMouseWheelScroll.delegate = self;
+    _grMouseWheelScroll.allowedScrollTypesMask = 3; // discrete & continuous only
```
tigrim: UIScrollTypeMaskAll

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -629,6 +640,9 @@ - (void) viewDidAppear:(BOOL)animated
     [_mapView addGestureRecognizer:_grElevation];
     [_mapView addGestureRecognizer:_grSymbolContextMenu];
     [_mapView addGestureRecognizer:_grPointContextMenu];
+
+    if (@available(iOS 13.4, *))
```
tigrim: remove (min iOS target 15.0)

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -1527,18 +1554,28 @@ - (void) zoomOutGestureDetected:(UITapGestureRecognizer *)recognizer
...
+    else
+        centerPoint = [recognizer locationInView:self.view]; // MouseWheelScroll
+
+    if (recognizer.numberOfTouches > 0)
     {
-        CGPoint touchPoint = [recognizer locationOfTouch:touchIdx inView:self.view];
-        
-        centerPoint.x += touchPoint.x;
-        centerPoint.y += touchPoint.y;
+        for(NSInteger touchIdx = 1; touchIdx < recognizer.numberOfTouches; touchIdx++)
```
tigrim: for (NSInteger

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -417,7 +419,16 @@ - (void) commonInit
     _grZoomOut.delegate = self;
     _grZoomOut.numberOfTapsRequired = 2;
     _grZoomOut.numberOfTouchesRequired = 2;
-    
+
+    // - MouseWheelScroll gesture (run under MacOS)
+    _grMouseWheelScroll = [[UIPanGestureRecognizer alloc] initWithTarget:self
```
tigrim: If `_grMouseWheelScroll` is used only on macOS, its initialization and addition should be wrapped in` [OAUtilities isiOSAppOnMac]`

Sources/Controllers/Map/OAMapViewController.mm
RZR-UA: fixed

tigrim (approved):

PR #4749, opened 2025-07-24
Navigation widget second UI review fixes

Vladyslav-lys:


Sources/Controllers/Map/Widgets/OANextTurnWidget.mm
```diff
@@ -165,6 +165,13 @@ - (instancetype)initWithHorisontalMini:(BOOL)horisontalMini
     return self;
 }
 
+- (void) layoutSubviews
+{
+    [super layoutSubviews];
+    if ([self isPanelVertical])
```
tigrim: moved` [self isPanelVertical] `into a separate variable

tigrim (approved):

Issue #4559, opened 2025-04-29
Loss of `extensions` after uploading a track to Cloud


Zirochkabila:
### Description

Original [21798](https://github.com/osmandapp/OsmAnd/issues/21798)

1. When sharing tracks between iOS devices, the following data is lost (color, width, movement arrows, start and finish points)
2. The following data (color, width, movement arrows, start and finish points) is lost when sharing folders between iOS devices

### Steps to reproduce

1. iOS > iOS (share, export, Cloud, separate track and track inside a folder with changed appearance)

### Actual result

**A separate track uploaded as a `.gpx`, `.osf` or from the `Cloud` - **has incorrect settings in .gpx****

[Timmi Terrible.gpx.zip](https://github.com/user-attachments/files/19602967/Timmi.Terrible.gpx.zip)

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th>Original</th>
    <th>🔴 .gpx</th>
    <th>🔴 Reset to original from .gpx</th>
    <th>.osf</th>
    <th>Cloud</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/671bcedc-2c01-4eb5-b10e-0e94b3066b5c"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/bfc14785-56e1-42a7-b4a2-0bb73dadde2a"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/727a2d10-15bc-4a22-b587-eea31a6fcbef"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/435a830c-709e-4c83-8226-f5d237c4c2a8"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/e9edddf3-2e41-4f8f-a78f-d8682b43b169"  width="300" height="auto"/> </th>
  </tr>
</table> 
</details> 

**Track inside a folder with a changed appearance**

[sun is up.gpx.zip](https://github.com/user-attachments/files/19603192/sun.is.up.gpx.zip)

<details><summary>Details</summary>
<table style="width:100%">
  <tr>
    <th>Original</th>
    <th>Folder appereance</th>
    <th>Track appereance after folder changed</th>
    <th>🔴 .gpx</th>
    <th>🔴.Reset to original from .gpx</th>
    <th>.osf (folder)</th>
    <th>Cloud</th>
[...]

PR #4784, opened 2025-08-01
Tracks appearance are not exported to info files while uploading to C…

tigrim:
…loud https://github.com/osmandapp/OsmAnd-iOS/issues/4559

### DIFF
--- a/Sources/Helpers/OANetworkUtilities.m
+++ b/Sources/Helpers/OANetworkUtilities.m
@@ -14,6 +14,8 @@
 
 #define BOUNDARY @"CowMooCowMooCowCowCow"
 
+#define DEBUG_NETWORK_OPERATIONS 0  // 1 — on, 0 — off
+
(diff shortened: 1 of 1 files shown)


Sources/Helpers/OANetworkUtilities.m
```diff
@@ -159,10 +184,52 @@ + (void) uploadFile:(NSString *)url fileName:(NSString *)fileName params:(NSDict
     [postData appendData:[[NSString stringWithFormat:@"\r\n--%@--\r\n", BOUNDARY] dataUsingEncoding:NSUTF8StringEncoding]];
     [request setHTTPBody:postData];
 
+#ifdef DEBUG
```
alex-dev-neo: DEBUG is too common for such logs. It is better to introduce new flag in this class like DEBUG_NETWORK_OPERATIONS

PR #4785, opened 2025-08-04
Add ability to resize distance measurement text

Vladyslav-lys:


Sources/Controllers/Map/Layers/OARulerByTapControlLayer.mm
```diff
@@ -205,11 +210,13 @@ - (void) drawLineEnds:(QVector<OsmAnd::PointI>)points
 
 - (void) drawDistanceMarker:(NSString *)distance
 {
+    float textSizeFactor = [OADistanceByTapTextSizeConstant getTextSizeFactor:[_settings.distanceByTapTextSize get:_appMode]];
+    OsmAnd::TextRasterizer::Style captionStyle = self.captionStyle.setSize([self getNormalCaptionSize] * textSizeFactor);
```
alex-dev-neo: Do not modify self.captionStyle

PR #4792, opened 2025-08-05
Fixed markers order and offset for ruler

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd-Issues/issues/1587

PR #4856, opened 2025-09-04
Crash fix 5.2.1

tigrim:


Sources/Controllers/Map/OANetworkRouteDrawable.mm
```diff
@@ -42,9 +42,17 @@ - (instancetype)initWithRouteKey:(OARouteKey *)routeKey
 
 - (UIImage *)getIcon
 {
+    if (!_routeKey)
+        return nil;
+    
+    if (!_env || !_env.mapPresentationEnvironment)
+        return nil;
+    
+    const auto &env = _env.mapPresentationEnvironment;
```
alex-dev-neo: `const auto env = _env.mapPresentationEnvironment;`
and then check for nullptr

Sources/Helpers/OAFavoritesHelper.mm
```diff
@@ -885,9 +885,9 @@ + (void)updateGroupAppearance:(OAFavoriteGroup *)favoriteGroup
     }
 }
 
-+ (NSMutableArray<OAFavoriteGroup *> *) getFavoriteGroups
++ (NSArray<OAFavoriteGroup *> *)getFavoriteGroups
 {
-    return _favoriteGroups;
+    return [_favoriteGroups copy];
```
alex-dev-neo: It is better to not return copy here as well as at `getFavoriteItems`.
`_cachedFavoritePoints` and `_favoriteGroups` should be changed to `NSArray` and new array should be assigned on modifying the array.

PR #4869, opened 2025-09-08
Fix updateCachedCurrentInterfaceOrientation and accessibilityIdentifier

DmitrySvetlichny:


Sources/Controllers/QuickAction/Buttons/MapButtonState.swift
```diff
@@ -27,6 +27,7 @@ open class MapButtonState: NSObject {
         self.positionSize = ButtonPositionSize(id: id)
         self.defaultPositionSize = ButtonPositionSize(id: id)
         super.init()
+        updatePortraitFlag()
```
tigrim: ?

tigrim (approved):

PR #4882, opened 2025-09-12
Buttons layout grid u ireview

DmitrySvetlichny:


Sources/Helpers/MigrationManager.swift
```diff
@@ -481,6 +482,70 @@ final class MigrationManager: NSObject {
...
+                    continue
+                }
+                
+                let portMargin = fabPref.getPortraitFabMargin(mode)
+                let landMargin = fabPref.getLandscapeFabMargin(mode)
+                let portX = CGFloat(portMargin.first?.doubleValue ?? 0)
+                let portY = CGFloat(portMargin.last?.doubleValue ?? 0)
+                let landX = CGFloat(landMargin.first?.doubleValue ?? 0)
+                let landY = CGFloat(landMargin.last?.doubleValue ?? 0)
+                let scenarios: [(need: Bool, x: CGFloat, y: CGFloat, parentW: CGFloat, parentH: CGFloat, pref: OACommonLong)] = [(needPortrait, portX, portY, portraitWidth, portraitHeight, portraitPref), (needLandscape, landX, landY, landscapeWidth, landscapeHeight, landscapePref)]
```
tigrim: It’s better to use a` private struct `for 6 elements instead of a tuple — having that many items in a tuple is not a good practice

PR #4914, opened 2025-09-30
Remove Widgets Visibility

DmitrySvetlichny:


Sources/Controllers/Map/Widgets/Helpers/WidgetsAvailabilityHelper.swift
```diff
@@ -51,11 +51,6 @@ class WidgetsAvailabilityHelper: NSObject {
         regWidgetAvailability(widgetType: .secondNextTurn, appModes: exceptDefault)
         
         // right
-        regWidgetVisibility(widgetType: .intermediateDestination)
```
tigrim, line 54: (https://github.com/osmandapp/Osmand/blob/cfcfc4174e3ae10a219fc0d5928e7bec5f0c9394/OsmAnd/src/net/osmand/plus/settings/backend/WidgetsAvailabilityHelper.java#L106)

PR #4965, opened 2025-10-16
show distance on graph slider in track menu

Vladyslav-lys:


Sources/Helpers/GpxUIHelper.swift
```diff
@@ -250,6 +256,16 @@ class GpxUIHelper: NSObject {
                     }
                 }
             }
+            
+            if let dataSet = chartData?.dataSets.first as? OrderedLineDataSet,
+               showXAxisValue && dataSet.getDataSetAxisType() == .distance {
+                let meters = entry.x * dataSet.getDivX()
+                if let formattedDistance = OAOsmAndFormatter.getFormattedDistance(Float(meters)) {
+                    res.append(NSAttributedString(string: ", \(formattedDistance)",
```
tigrim: It seems the display `","` will be redundant if there is no data before formattedDistance.

Sources/Helpers/GpxUIHelper.swift
```diff
@@ -193,12 +193,18 @@ class GpxUIHelper: NSObject {
...
 
         override init(frame: CGRect) {
             super.init(frame: frame)
 
             setupLayers()
         }
+        
+        convenience init(frame: CGRect, showXAxisValue: Bool) {
+            self.init(frame: frame)
+            self.showXAxisValue = showXAxisValue
```
tigrim, line 206: setupLayers()?

PR #4968, opened 2025-10-17
crash fix

DmitrySvetlichny:


tigrim:
if (imageLink.length > 0)
            [allImageLinks addObject:imageLink];

Issue #4957, opened 2025-10-15
Wrong map is proposed to download in California


dmpr0:
## Description
The "Los Angeles North" map is not suggested for download; instead, the "Los Angeles" map is shown. After downloading the "Los Angeles" map, the download dialog no longer appears for other parts of the region that do not have a map.

### How to reproduce
* Open map at ‎`34.22122° N, 118.99707° W‎`

### Actual result

<table style="width:100%">
  <tr>
    <th>California</th>
    <th>Oregon</th>
    <th>Overview</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/0c74cb3a-16cd-4fba-8a95-9957a298a267"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/add01bdd-000e-4201-8976-1d4c3325708a"  width="300" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/7058ed15-33b3-4411-81a3-78a269262bf7"  width="300" height="auto"/> </td>
  </tr>
</table>  

* If there is no map for California, Los Angeles map will be proposed
* After downloading, download dialog no longer appears 

### Expected result
* Show proper region for download

**Environment**
OsmAnd Version: 5250
iOS version: 26
Device model: iPhone 16 Pro Max

Zirochkabila:
> 5250
> OsmAnd~ 5.2.0#6282m, Released: 2025-10-16


<table style="width:100%">
  <tr>
    <th>Android</th>
    <th>iOS</th>
  </tr>
  <tr>
<th><img src="https://github.com/user-attachments/assets/ef618285-ad19-4003-be08-fa59297ad617"  width="300" height="auto"/> </th>
 <th><img src="https://github.com/user-attachments/assets/72e448b5-4a32-44c8-81e5-5afd8e3329bd"  width="300" height="auto"/> </th>
  </tr>
</table>

PR #4971, opened 2025-10-17
fix wrong map is proposed to download in California

tigrim:


tigrim:
https://github.com/osmandapp/OsmAnd-iOS/issues/4957

PR #5030, opened 2025-11-04
Fix Update recording indicator

DmitrySvetlichny:


Sources/Helpers/SharedLibHelpers/GpxFileExtension.swift
```diff
@@ -56,4 +56,28 @@ extension GpxFile {
...
+        var segs = processedPointsToDisplay ?? []
+        if segs.isEmpty {
+            let seg = TrkSegment()
+            seg.points = NSMutableArray()
+            segs.append(seg)
+            processedPointsToDisplay = segs
+        }
+        
+        let last = segs[segs.count - 1]
+        last.points.add(point)
```
tigrim: it seems better to move it inside the `if let block `

tigrim (approved):

Issue #4894, opened 2025-09-17
Appearance tags are visible in track context menu after track import


dmpr0:
### How to reproduce
* Import track – [Wed 17 Sep 2025.gpx.zip](https://github.com/user-attachments/files/22381068/Wed.17.Sep.2025.gpx.zip)
* Open context menu details

### Actual result
<table style="width:100%">
  <tr>
    <td><img src="https://github.com/user-attachments/assets/892e010f-d4f1-4ae6-989f-babf8c182be2"  width="300" height="auto"/> </td>
  </tr>
</table>   

* Appearance tags visible in context menu

### Expected result
 
* Appearance tags should not be visible

**Environment**
OsmAnd Version: 5235
iOS version: 26
Device model: iPhone 16 Pro Max

vshcherb:
Internal tags should not be visible in poi_types.xml they marked as hidden. 
```
<poi_additional name="shield_bg" tag="shield_bg" type="text" no_edit="true" hidden="true"/>
			<poi_additional name="shield_fg" tag="shield_fg" type="text" no_edit="true" hidden="true"/>
			<poi_additional name="shield_fg_2" tag="shield_fg_2" type="text" no_edit="true" hidden="true"/>
			<poi_additional name="shield_text" tag="shield_text" type="text" no_edit="true" hidden="true"/>
```

RZR-UA:
To Do

- [ ] port and use Android GpxAppearanceInfo.isGpxAppearanceTag (or better move class to Kotlin)
- [ ] implement processing of the `is_join_segments` in Android (if required)

Note: all new GPX appearance tags should be added and processed with isGpxAppearanceTag()

PR #5031, opened 2025-11-04
Gpx appearance info sync

nnngrach:
[issue](https://github.com/osmandapp/OsmAnd-iOS/issues/4894)
[previous pr](https://github.com/osmandapp/OsmAnd-iOS/pull/4993)

- Rewrited OAGpxAppearanceInfo.mm -> GpxAppearanceInfo.swift
- Delited old OAGpxAppearanceInfo.mm
- Synced GpxAppearanceInfo.swift with android

Tested with settings export with gpx.
Exported gpx looks the same:

[old Wed_17_Sep_2025.gpx.zip](https://github.com/user-attachments/files/22381068/Wed.17.Sep.2025.gpx.zip)
[new Wed_17_Sep_2025.gpx.zip](https://github.com/user-attachments/files/23354008/new.Wed_17_Sep_2025.gpx.zip)

Screenshot 2025-11-05 at 12 46 56

### DIFF
--- a/Sources/GPX/GpxAppearanceInfo.swift
+++ b/Sources/GPX/GpxAppearanceInfo.swift
@@ -141,13 +141,13 @@ final class GpxAppearanceInfo: NSObject {
         }
         
         if let value = json[TAG_SHOW_ARROWS] {
-            gpxAppearanceInfo.showArrows = (value as? Bool) ?? (value as? String).map { ($0 as NSString).boolValue } ?? false
+            gpxAppearanceInfo.showArrows = boolValue(from: value)
         }
         if let value = json[TAG_START_FINISH] {
-            gpxAppearanceInfo.showStartFinish = (value as? Bool) ?? (value as? String).map { ($0 as NSString).boolValue } ?? false
+            gpxAppearanceInfo.showStartFinish = boolValue(from: value)
         }
         if let value = json[TAG_IS_JOIN_SEGMENTS] {
-            gpxAppearanceInfo.isJoinSegments = (value as? Bool) ?? (value as? String).map { ($0 as NSString).boolValue } ?? false
+            gpxAppearanceInfo.isJoinSegments = boolValue(from: value)
         }
 
         gpxAppearanceInfo.splitType = OAGPXDatabase.splitType(byName: json[TAG_SPLIT_TYPE] as? String ?? "")
@@ -259,5 +259,11 @@ final class GpxAppearanceInfo: NSObject {
         if !value.isNaN {
             json[name] = value
         }
+    }
+    
+    private static func boolValue(from any: Any?) -> Bool {
+        if let bool = any as? Bool { return bool }
+        if let string = any as? String { return (string as NSString).boolValue }
+        return false
     }
 }


Sources/GPX/GpxAppearanceInfo.swift
```diff
@@ -61,7 +61,193 @@ final class GpxAppearanceInfo: NSObject {
...
+    
+    // to run from obj-c
+    func toJson(_ json: [String: Any]) -> [String: Any] {
+        var newJson = json
+        toJson(&newJson)
+        return newJson
+    }
+    
+    private static func writeParam(_ json: inout [String: Any], name: String, value: Any?) {
+        guard let value = value else { return }
```
tigrim: use short syntax

Sources/GPX/GpxAppearanceInfo.swift
```diff
@@ -61,7 +61,193 @@ final class GpxAppearanceInfo: NSObject {
...
+        
+        if let v = value as? Int, v != 0 {
```
tigrim: use switch

tigrim: switch value {
    case let v as Int where v != 0:
        json[name] = "\(v)"
   case let v as Double where !v.isZero && !v.isNaN:
        json[name] = "\(v)"
    ....
    }

Sources/GPX/GpxAppearanceInfo.swift
```diff
@@ -61,7 +61,193 @@ final class GpxAppearanceInfo: NSObject {
...
+        Self.writeParam(&json, name: Self.TAG_LINE_3D_VISUALIZATION_POSITION_TYPE, value: OAGPXDatabase.lineVisualizationPositionTypeName(for: trackLinePositionType))
+        
+        Self.writeParam(&json, name: Self.TAG_VERTICAL_EXAGGERATION_SCALE, value: verticalExaggeration)
+        Self.writeParam(&json, name: Self.TAG_ELEVATION_METERS, value: elevationMeters)
+        
+        Self.writeParam(&json, name: Self.TAG_TIME_SPAN, value: timeSpan)
+        Self.writeParam(&json, name: Self.TAG_WPT_POINTS, value: wptPoints)
+        Self.writeParam(&json, name: Self.TAG_TOTAL_DISTANCE, value: totalDistance)
+        
+        Self.writeValidDouble(&json, name: Self.TAG_SMOOTHING_THRESHOLD, value: smoothingThreshold)
```
tigrim, line 218: Why don’t we use writeParam?

Sources/GPX/GpxAppearanceInfo.swift
```diff
@@ -61,7 +61,193 @@ final class GpxAppearanceInfo: NSObject {
         TAG_IS_JOIN_SEGMENTS
     ])
     
+    var width: String?
+    var coloringType: String?
+    var gradientPaletteName: String?
+    var color: Int = 0
+    var splitType: EOAGpxSplitType = .none
+    var splitInterval: Double = 0.0
+    var showArrows: Bool = false
```
tigrim: var showArrows = false

Sources/GPX/GpxAppearanceInfo.swift
DmitrySvetlichny: It matches Android. writeValidDouble writes native numbers and doesn't discard leading zeros.

Sources/GPX/GpxAppearanceInfo.swift
```diff
@@ -61,7 +61,203 @@ final class GpxAppearanceInfo: NSObject {
...
+            } else if let string = color as? String {
+                gpxAppearanceInfo.color = Int(UIColor.toNumber(from: string))
+            }
+        }
+        
+        if let value = json[TAG_WIDTH] as? String {
+            gpxAppearanceInfo.width = value
+        }
+        
+        if let value = json[TAG_SHOW_ARROWS] {
```
tigrim, line 143: It’s necessary to extract the repeated code (Bool) into a separate function

tigrim: Example:
```
private static func boolValue(from any: Any?) -> Bool {
    if let b = any as? Bool { return b }
    if let s = any as? String { return (s as NSString).boolValue }
    return false
}
```

tigrim (approved):

PR #5046, opened 2025-11-10
Crash fixes 5.2.2 73

tigrim:


Sources/Models/OATableSectionData.m
```diff
@@ -103,7 +103,9 @@ - (void)removeRowAtIndex:(NSInteger)index
 - (NSUInteger)rowCount
 {
     NSUInteger res = 0;
-    for (OATableRowData *row in _rowData)
+    NSArray *rowDataCopy = [_rowData copy];
```
alex-dev-neo, line 106: why copy? it should always be in UI thread

PR #5053, opened 2025-11-12
Improve UI max axle load max weight full load UI review fixed

Vladyslav-lys:


Sources/Controllers/Settings/NavigationSettings/OAVehicleParametersViewController.mm
```diff
@@ -170,6 +175,11 @@ - (void)generateData
     _data = [NSArray arrayWithArray:tableData];
 }
 
+- (NSString *)parameterValue:(NSString *)value paramId:(NSString *)paramId cutUnit:(BOOL)cutUnit
+{
+    return [NSString stringWithFormat:@"%@ %@", cutUnit ? [value substringToIndex:value.length - 1] : value, OALocalizedString([RouteParamVehicleHelper isWeightParameter:paramId] ? @"metric_ton" : @"m")];
```
tigrim: [value substringToIndex:value.length - 1] — potential crash

PR #5056, opened 2025-11-12
updated three dots image

Vladyslav-lys:


Resources/Images.xcassets/Icons/three_dots.imageset/Contents.json
```diff
@@ -0,0 +1,12 @@
+{
+  "images" : [
+    {
+      "filename" : "three_dots.png",
```
tigrim: I don’t see that the @2x and @3x icons were transferred.

tigrim (approved):

PR #5097, opened 2025-12-03
Speed limit animation

tigrim:


Sources/Router/OARoutingHelper.mm
```diff
@@ -487,7 +487,7 @@ - (int) calculateCurrentRoute:(CLLocation *)currentLocation posTolerance:(float)
                     double diffToNext = ABS(degreesDiff(bearingMotion, bearingRouteNext));
                     if (diff > diffToNext)
                     {
-                        NSLog(@"Processed point bearing deltas : %f %f", diff, diffToNext);
```
tigrim, line 490: This produces excessive console logs. Temporarily commented out

PR #5098, opened 2025-12-03
Unite search results by osmid / wikidata

ivanPyrohivskyi:
To issue https://github.com/osmandapp/OsmAnd/issues/23892
Java code [uniteSearchResultsByOsmIdOrWikidata ](https://github.com/osmandapp/OsmAnd/blob/55724c0966fff6e4045210c4e1906cf9b7c7586b/OsmAnd-java/src/main/java/net/osmand/search/SearchUICore.java#L337)

Combine by wikidata ([Le Phare](https://www.openstreetmap.org/way/108358489)):
 

Combine by osm_id ([Fuel and liquid gas store](https://www.openstreetmap.org/way/451454731)):

### DIFF
--- a/Sources/Search/OASearchUICore.mm
+++ b/Sources/Search/OASearchUICore.mm
@@ -372,13 +372,8 @@ const static NSArray<NSNumber *> *compareStepValues = @[@(EOATopVisible),
         {
             OAPOI *that = (OAPOI *)sr.object;
             int64_t osmId = [that getOsmId];
-            NSNumber * osmIdObj = osmId == -1 ? nil : @(osmId);
+            NSNumber * osmIdObj = osmId < 0 ? nil : @(osmId);
             NSString *wikidata = [that getWikidata];
-
-            if (osmIdObj != nil && [osmIdObj longLongValue] < 0)
-            {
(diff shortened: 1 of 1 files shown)


Sources/Search/OASearchUICore.mm
```diff
@@ -357,6 +360,127 @@ - (OASearchResultCollection *) addSearchResults:(NSArray<OASearchResult *> *)sr
...
+    NSMutableDictionary<NSNumber *, NSNumber *> *osmIdMap = [NSMutableDictionary dictionary];
+    NSMutableDictionary<NSString *, NSNumber *> *wikidataMap = [NSMutableDictionary dictionary];
+
+    for (OASearchResult *sr in input)
+    {
+        if ([sr.object isKindOfClass:OAPOI.class])
+        {
+            OAPOI *that = (OAPOI *)sr.object;
+            int64_t osmId = [that getOsmId];
+            NSNumber * osmIdObj = osmId == -1 ? nil : @(osmId);
```
alex-dev-neo: why not compare < 0 here?

ivanPyrohivskyi: -1 is mean null on Java.
< 0 - can be object from basemap

PR #5121, opened 2025-12-19
Popular places images preview

tigrim:


Sources/Core/OACoreResourcesAmenityIconProvider.mm
```diff
@@ -79,6 +79,15 @@
             if (!type)
                 continue;
             
+            QHash<QString, QString> decodedValues = amenity->getDecodedValuesHash();
```
alex-dev-neo: Will it slowdown getIcon method?

Sources/POI/ExplorePlacesOnlineProvider.swift
```diff
@@ -0,0 +1,309 @@
...
+        }
+
+        let loadAll = Int(zoom) == Self.maxLevelZoomCache &&
+                      (abs(Double(maxTileX - minTileX)) <= Self.loadAllTinyRect)
+
+        var filteredAmenities: [OAPOI] = []
+        var uniqueIds = Set<Int64>()
+        let languages = getPreferredLangs()
+
+        for tx in minTileX...maxTileX {
```
alex-dev-neo, line 108: There was a fix in android code (https://github.com/osmandapp/OsmAnd/pull/24488/changes) Probably it needs to be transferred here.

Sources/Core/OACoreResourcesAmenityIconProvider.mm
tigrim: I haven’t performed any measurements. Visually, though, it seems to render in about the same time. On Android, this works via getIconNameForAmenity

Sources/POI/ExplorePlacesOnlineProvider.swift
tigrim: done

PR #5140, opened 2026-01-07
Fix "Search more" button in poi context menu show empty list. Rework Search field phrase trimming to one last word.

nnngrach:
[current issue](https://github.com/osmandapp/OsmAnd/issues/20506#issuecomment-3631857751)

[related issue](https://github.com/osmandapp/OsmAnd/issues/22089#issuecomment-3233847418)

before / after (photo):

Screenshot 2026-01-08 at 01 25 35    Screenshot 2026-01-08 at 01 25 56

before (video):

https://github.com/user-attachments/assets/7c557de6-2778-4004-9608-3dc0cd3e5b0f

after (video):

https://github.com/user-attachments/assets/f064ad1e-70c0-4a79-a90b-43502e3656ef

---

Search more button - test point:

https://www.openstreetmap.org/way/1294840099#map=20/53.8693260/-1.7048201

Previous search field resetting [bug](https://github.com/osmandapp/OsmAnd-iOS/pull/4907) -  test point:

https://www.openstreetmap.org/node/1539266304#map=21/49.4522200/11.0695000

phrase: "turm de sinne"

### DIFF
--- a/Sources/Controllers/Panels/OAMapPanelViewController.mm
+++ b/Sources/Controllers/Panels/OAMapPanelViewController.mm
@@ -1214,22 +1214,24 @@ typedef enum
 
         if ([object isKindOfClass:[OAPOICategory class]])
         {
-            if (NSStringIsEmpty(objectLocalizedName))
-            {
-                objectLocalizedName = ((OAPOICategory *) object).nameLocalized;
-            }
+            OAPOICategory *c = ((OAPOICategory *) object);
+            objectLocalizedName = c.nameLocalized;
             phrase = [searchUICore resetPhrase:[NSString stringWithFormat:@"%@ ", objectLocalizedName]];
         }
         else if ([object isKindOfClass:[OAPOIUIFilter class]])
         {
             OAPOIUIFilter *filter = (OAPOIUIFilter *) object;
             filterByName = [filter.filterId isEqualToString:BY_NAME_FILTER_ID] || [filter.filterId hasPrefix:topIndexBrandPrefix];
-            if (NSStringIsEmpty(objectLocalizedName))
-            {
-                objectLocalizedName = filterByName ? filter.filterByName : filter.name;
-            }
+            objectLocalizedName = filterByName ? filter.filterByName : filter.name;
             phrase = [searchUICore resetPhrase];
         }
+        
+        if (!NSStringIsEmpty(searchQuery))
+        {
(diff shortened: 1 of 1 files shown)


nnngrach:
More context of this fix. 

---

There was old bug with search field resetting. Here is old task of that bug:
https://github.com/osmandapp/OsmAnd/issues/22089#issuecomment-3233847418

Short description:

- Open search. 
- Type "turm de sinne".
- Press "Show on map" button. (Search screen closes, map opens).
- Press Back button. (map screen closes, Search screen reopens).
- Search filed text now shows only last word: "sinne". Bug. Search results list is dfifferent.

Screenshot 2026-01-08 at 19 02 53

---

There was a fix of this bug. 

https://github.com/osmandapp/OsmAnd-iOS/pull/4907

And it causes a regression of "Show more" button in context menu (current issue).  Because of it search screen sometimes creates with empty Search field. Instead of filled with search name or category ("Shop" in our rxample).

Screenshot 2026-01-08 at 19 32 44


---

This code blocks in ios and in android:

https://github.com/osmandapp/OsmAnd-iOS/blob/master/Sources/Controllers/Panels/OAMapPanelViewController.mm#L1215

https://github.com/osmandapp/Osmand/blob/3fd171dbefd070808433e8983a3fb544613b6c9d/OsmAnd/src/net/osmand/plus/search/dialogs/QuickSearchDialogFragment.java#L1925

This two identical code blocks (right) in ios was united in one block (left):

Screenshot 2026-01-08 at 18 52 00

By the way, this is a similar code. 

Screenshot 2026-01-08 at 19 16 31

---

We're most interested in selected part. Especially, in the Search field text initialisation line: `objectLocalizedName = filter.getName();`

But there is a difference. 

In android whole this function called only one time: on Search screen init. When search screen closes (show on map button) and then restored from map (back button), look like it don't recreates from scratch. Just restores. 

In ios after "Show on map" button pressed, it kills Search screen. Shows map. And on Back button pressed, it create a new Search screen instance. Just with different init parameters. So this function call two time. On first screen creation. And on recreation after "Show on map". 

---

So. Fix in this pr reverts some code prom previous fix. It added a new Bool function's parameter. And new little isolated code block, that should be called only on screen reopening and from never more. (to not break any previous logic).

This new code still fixes old bug: "turm de sinne" - > "sinne".
And don't create current "Search more" bug anymore.

PR #5163, opened 2026-01-21
Poi share link UI review

DmitrySvetlichny:


Sources/POI/DeepLinkParser.swift
```diff
@@ -370,6 +370,12 @@ final class DeepLinkParser: NSObject {
         return OAAmenitySearcher.sharedInstance().searchDetailedObject(with: request)
     }
     
+    private func normalizeQueryValue(_ rawValue: String?) -> String? {
+        guard let rawValue, !rawValue.isEmpty else { return nil }
+        let normalized = (rawValue.replacingOccurrences(of: "+", with: " ").removingPercentEncoding ?? rawValue).trimmingCharacters(in: .whitespacesAndNewlines)
+        return normalized.isEmpty ? nil : normalized
```
tigrim: Why not just return normalized?

Sources/Helpers/OAShareMenuActivity.mm
```diff
@@ -65,19 +65,26 @@ - (UIImage *)activityImage
...
         case OAShareMenuActivityCopyPOIName:
-            return [[UIImage systemImageNamed:@"document.on.document"] imageWithRenderingMode:UIImageRenderingModeAlwaysTemplate];
+            return [self sfSymbol:@"document.on.document"];
     }
     return nil;
 }
 
+- (UIImage *)sfSymbol:(NSString *)name
+{
+    UIImageSymbolConfiguration *config = [UIImageSymbolConfiguration configurationWithPointSize:23 weight:UIImageSymbolWeightRegular scale:UIImageSymbolScaleLarge];
```
tigrim: Why 23? Is there a way to use the system default size instead? For example, `[UIImageSymbolConfiguration configurationWithScale:UIImageSymbolScaleLarge]`

Issue #5126, opened 2025-12-22
Add MSAA smoothing for CarPlay


dmpr0:
## ToDo
Android has the Enable MSAA parameter for smoothing the 3D model of the location icon. On iOS, it is probably not needed due to the higher resolution, but it can be enabled for CarPlay.

<table style="width:100%">
  <tr>
    <th colspan=2>CarPlay</th>
  </tr>
  <tr>
    <td colspan=2><img src="https://github.com/user-attachments/assets/563058aa-182b-447e-921b-0af98123f448"  width="auto" height="auto"/> </td>
  </tr>
  <tr>
    <th>iPhone: Bearing</th>
    <th>iPhone: Moving</th>
  </tr>
  <tr>
    <td><img src="https://github.com/user-attachments/assets/fd4ef0ce-d2a8-4e5c-b422-68cfca901e90"  width="auto" height="auto"/> </td>
    <td><img src="https://github.com/user-attachments/assets/a66a0d4a-7075-463e-aef1-e887436e44bb"  width="auto" height="auto"/> </td>
  </tr>
</table>  

We can add **Enable MSAA** with parameters: Off / iPhone / CarPlay / Both

PR #5164, opened 2026-01-22
Supported MSAA

ihoryakovenko:


ihoryakovenko:
https://github.com/osmandapp/OsmAnd-iOS/issues/5126

PR #5169, opened 2026-01-26
Update colour component customise route line

DmitrySvetlichny:


Sources/Controllers/RouteInfoMenu/OARouteLineAppearanceHudViewController.mm
```diff
@@ -1285,15 +1265,29 @@ - (void)updateData:(OAGPXBaseTableData *)tableData
...
+            item = ci;
+            break;
+        }
+    }
     
-    if ([customColorDay isEqual:defaultColorDay])
-        [_previewRouteLineInfo setCustomColor:_availableColors.firstObject.intValue nightMode:NO];
-    if ([customColorNight isEqual:defaultColorNight])
-        [_previewRouteLineInfo setCustomColor:_availableColors.firstObject.intValue nightMode:YES];
+    _selectedColorItem = item ?: [_appearanceCollection getColorItemWithValue:(int)currentValue] ?: [_appearanceCollection getDefaultLineColorItem];
```
tigrim: using two ternary operators in the same logic is a bad practice, so we avoid it

Sources/Controllers/RouteInfoMenu/OARouteLineAppearanceHudViewController.mm
```diff
@@ -1546,144 +1504,129 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
...
-                            forControlEvents:UIControlEventValueChanged];
-
-            NSMutableDictionary *values = [cellData.values mutableCopy];
-            values[@"cell_value"] = cell;
-            [cellData setData:values];
         }
+        
+        NSInteger selectedIndex = 0;
+        if ([cellData.key isEqualToString:@"color_day_night_value"])
+            selectedIndex = [arrayValue indexOfObject:_selectedDayNightMode];
```
tigrim: Using indexOfObject here is unsafe

tigrim: NSNotFound result

Sources/Controllers/RouteInfoMenu/OARouteLineAppearanceHudViewController.mm
```diff
@@ -311,7 +311,13 @@ - (void)setOldValues
 
 - (void)updateAllValues
 {
+    _nightMode = _settings.nightMode;
+    _selectedDayNightMode = _nightMode ? kColorNightMode : kColorDayMode;
```
tigrim: I think u can use the localized strings directly without defining macros like kColorDayMode and kColorNightMode.

PR #5173, opened 2026-01-27
Temporary files hidden from the search

DmitrySvetlichny:


Sources/Search/OAQuickSearchHelper.mm
```diff
@@ -377,6 +379,10 @@ - (BOOL) search:(OASearchPhrase *)phrase resultMatcher:(OASearchResultMatcher *)
 {
     for (OASGpxDataItem *dataItem in OAGPXDatabase.sharedDb.getDataItems)
     {
+        // Exclude temporary GPX tracks from search.
+        if ([self isTemporaryGpxItem:dataItem])
```
tigrim: use common method [self shouldIgnoreGpxItem:dataItem]

PR #5179, opened 2026-02-02
Car play finish navigation template

DmitrySvetlichny:


Sources/Router/OARoutingHelper.mm
```diff
@@ -175,8 +179,59 @@ - (BOOL) isPauseNavigation
     return _isPauseNavigation;
 }
 
+- (void)onCarNavigationStart
```
tigrim: Misleading name:
– Car is ambiguous, use CarPlay
– the method performs resume, not start

Sources/Router/OARoutingHelper.mm
```diff
@@ -175,8 +179,59 @@ - (BOOL) isPauseNavigation
...
+{
+    if (_isPausedOnCarDisconnect && [self isPauseNavigation])
+    {
+        _isPausedOnCarDisconnect = NO;
```
tigrim: _isPausedDueToCarPlayDisconnected

Sources/Router/OARoutingHelper.mm
```diff
@@ -175,8 +179,59 @@ - (BOOL) isPauseNavigation
     return _isPauseNavigation;
 }
 
+- (void)onCarNavigationStart
+{
+    if (_isPausedOnCarDisconnect && [self isPauseNavigation])
```
tigrim: _isPauseNavigation

Sources/Router/OARoutingHelper.mm
tigrim: I still don’t understand why this is needed

Sources/Router/OARoutingHelper.mm
DmitrySvetlichny: The method name was transferred from the Android code.

Sources/Router/OARoutingHelper.mm
DmitrySvetlichny: Used to auto-resume navigation only when the pause was caused by CarPlay disconnect.
Prevents resuming navigation that was paused manually by the user.
The code was synchronized with android

Sources/Router/OARoutingHelper.mm
tigrim: Ok, it needs to be made readable

Sources/Router/OARoutingHelper.mm
tigrim: isPausedOnAADisconnect is used in Android

DmitrySvetlichny: https://github.com/osmandapp/OsmAnd/blob/master/OsmAnd/src/net/osmand/plus/routing/RoutingHelper.java#L161

Sources/Router/OARoutingHelper.mm
```diff
@@ -175,8 +179,59 @@ - (BOOL) isPauseNavigation
...
+- (void)onCarPlayNavigationSessionRestored
+{
+    if (_isPausedDueToCarPlayDisconnect && _isPauseNavigation)
+    {
+        _isPausedDueToCarPlayDisconnect = NO;
+        [self resumeNavigation];
+    }
+}
+
+- (void)onCarNavigationSessionChanged
```
tigrim: up

PR #5192, opened 2026-02-06
address in what's here menu shown

Vladyslav-lys:


Sources/Controllers/Map/Layers/OAContextMenuLayer.mm
```diff
@@ -542,6 +542,7 @@ - (void) showContextMenuForSelectedObjects:(CLLocation *)touchPointLatLon select
         if (provider)
         {
             OATargetPoint *targetPoint = [provider getTargetPoint:selectedObject.object];
+            [targetPoint initAdderssIfNeeded];
```
tigrim: Not the most appropriate location for this logic. Where is the equivalent code on Android?

tigrim (approved):

PR #5203, opened 2026-02-13
fix in getName transliterate as Android

DmitrySvetlichny:


Sources/POI/OAPOI.mm
```diff
@@ -273,12 +273,12 @@ - (NSString *)getName:(NSString *)lang transliterate:(BOOL)transliterate
             nm = self.localizedNames[@"en"];
             if (!nm || nm.length == 0)
                 nm = OsmAnd::ICU::transliterateToLatin(QString::fromNSString(self.name)).toNSString();
-            return nm;
+            return nm && nm.length > 0 ? nm : self.name;
```
tigrim: return nm.length > 0 ? nm : self.name;

Sources/POI/OAPOI.mm
```diff
@@ -273,12 +273,12 @@ - (NSString *)getName:(NSString *)lang transliterate:(BOOL)transliterate
...
         }
         nm = self.localizedNames[lang];
         if (transliterate)
             nm = OsmAnd::ICU::transliterateToLatin(QString::fromNSString(nm)).toNSString();
-        return nm;
+        return nm && nm.length > 0 ? nm : self.name;
```
tigrim: up

PR #5214, opened 2026-02-23
UI in cell fix

DmitrySvetlichny:


Sources/Controllers/DashboardOnMap/MapSettings/MapSettingsCoordinatesGridScreen.swift
```diff
@@ -175,26 +175,25 @@ final class MapSettingsCoordinatesGridScreen: NSObject, OAMapSettingsScreen {
...
-            config.cornerStyle = .capsule
-            config.contentInsets = NSDirectionalEdgeInsets(top: 3.1, leading: 4, bottom: 3.1, trailing: 4)
-            config.imagePadding = 4
-            config.imagePlacement = .trailing
-            cell.rightButton.configuration = config
-            cell.rightButton.setContentHuggingPriority(.required, for: .horizontal)
-            cell.rightButton.setContentCompressionResistancePriority(.required, for: .horizontal)
+            let isRTL = cell.contentView.isDirectionRTL()
+            let arrow = isRTL ? "arrow.left" : "arrow.right"
+            var config = UIButton.Configuration.plain()
```
tigrim: I’ve seen this configuration before. Let’s make it shared

Sources/SwiftExtensions/UIButton+Extension.swift
```diff
@@ -22,4 +22,9 @@ extension UIButton {
             }
         }
     }
+
+    @objc(applyPurchasePlanButtonConfigurationWithTitle:)
+    func applyPurchasePlanButtonConfiguration(title: String) {
```
tigrim: ?

Sources/Controllers/DashboardOnMap/MapSettings/MapSettingsCoordinatesGridScreen.swift
tigrim: use @objc final class ButtonConfigurationHelper: NSObject {}

PR #5223, opened 2026-02-25
full screen for navigation and direction from

Vladyslav-lys:


Sources/Controllers/Map/OAMapActions.mm
```diff
@@ -60,13 +60,35 @@ - (instancetype)init
...
-- (void) enterRoutePlanningMode:(CLLocation *)from fromName:(OAPointDescription *)fromName
+- (void)enterRoutePlanningMode:(CLLocation *)from fromName:(OAPointDescription *)fromName
 {
     [self enterRoutePlanningModeGivenGpx:nil
                                     from:from
                                 fromName:fromName
           useIntermediatePointsByDefault:YES
-                              showDialog:YES];
+                              showDialog:YES
+                                fullMenu:YES];
```
tigrim: What is fullMenu? How is it on Android?

PR #5237, opened 2026-03-11
Add proBannerButtonConfigurationWithImageName

DmitrySvetlichny:


Sources/SwiftExtensions/UIButtonConfiguration+Extension.swift
```diff
@@ -21,4 +21,16 @@ extension UIButton.Configuration {
         config.background.backgroundColor = .buttonBgColorTertiary
         return config
     }
+    
+    static func proBannerButtonConfiguration(imageName: String) -> UIButton.Configuration {
```
tigrim: It should be migrated to use UIImage instead of imageName. For ButtonConfigurationHelper, you can keep imageName

PR #5239, opened 2026-03-12
middle state after setting destination

Vladyslav-lys:


Sources/Controllers/RouteInfoMenu/OARouteInfoView.mm
```diff
@@ -1124,7 +1125,10 @@ - (void)show:(BOOL)animated fullScreen:(BOOL)fullScreen onComplete:(void (^)(voi
     _optionsMenuSelected = NO;
     [_appModeView setupModeButtons];
     [_tableView setContentOffset:CGPointZero];
-    _currentState = fullScreen ? EOARouteInfoMenuStateFullScreen : _currentState;
+    EOARouteInfoMenuState menuState = _currentState;
+    if (!fullScreen && _currentState == EOARouteInfoMenuStateFullScreen)
```
tigrim, line 1129: Doesn’t this break the layout on Mac?

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -1658,7 +1658,7 @@ - (BOOL) processTargetPoint:(OATargetPoint *)targetPoint
             }
 
             [self hideTargetPointMenu];
-            [self showRouteInfo:NO];
+            [self showRouteInfo:(_activeTargetType == OATargetRouteStartSelection || _activeTargetType == OATargetRouteFinishSelection) && !_app.data.pointToNavigate];
```
tigrim: Unclear logic.    OATargetRouteIntermediate,
    OATargetRouteIntermediateSelection,

Sources/Controllers/RouteInfoMenu/OARouteInfoView.mm
Vladyslav-lys: Yes, it works correctly

PR #5260, opened 2026-03-26
Unblock HH+A* to run simultaneously with MissingMapsWarning (iOS part)

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/23853

### DIFF
--- a/Sources/Helpers/OAAppSettings.h
+++ b/Sources/Helpers/OAAppSettings.h
@@ -1532,7 +1532,6 @@ typedef NS_ENUM(NSInteger, EOAWikiDataSourceType)
 @property (nonatomic) OACommonLong *userIosIdExpiredTime;
 @property (nonatomic) OACommonBoolean *useOldRouting;
 @property (assign, nonatomic) BOOL ignoreMissingMaps;
-@property (assign, nonatomic) BOOL stopOnMissingMaps;
 
 @property (nonatomic) OACommonBoolean *wikiShowImagePreviews;
 
--- a/Sources/Router/OARouteProvider.mm
+++ b/Sources/Router/OARouteProvider.mm
@@ -1071,7 +1071,7 @@
     router->setUseFastRecalculation(settings.useFastRecalculation);
 
     router->CALCULATE_MISSING_MAPS = !settings.ignoreMissingMaps;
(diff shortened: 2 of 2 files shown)


Sources/Helpers/OAAppSettings.h
```diff
@@ -1531,11 +1529,10 @@ typedef NS_ENUM(NSInteger, EOAWikiDataSourceType)
 @property (nonatomic) OACommonLong *userIosIdExpiredTime;
 @property (nonatomic) OACommonBoolean *useOldRouting;
 @property (assign, nonatomic) BOOL ignoreMissingMaps;
+@property (assign, nonatomic) BOOL stopOnMissingMaps;
```
alex-dev-neo: It never assigns

Sources/Helpers/MissingMapsCalculator/MissingMapsCalculator.mm
```diff
@@ -223,47 +270,66 @@ - (BOOL)checkIfThereAreMissingMaps:(std::shared_ptr<RoutingContext>)ctx
...
+        if (ctx->progress != nullptr)
+        {
+            ctx->progress->missingMapsCalculationResult = nullptr;
+            ctx->progress->resetFastRoutingStatus();
+        }
         return NO;
     }
-    self.missingMaps = [self convert:[mapsToDownload copy]];
-    self.mapsToUpdate = [self convert:[mapsToUpdate copy]];
-    self.potentiallyUsedMaps = [self convert:[usedMaps copy]];
```
alex-dev-neo, line 230: Looks like these three fields never assigns. Should be dropped completely?

Sources/Helpers/OAAppSettings.h
RZR-UA: 0c9586410

Sources/Helpers/MissingMapsCalculator/MissingMapsCalculator.mm
RZR-UA:
``` 
- @property (nonatomic) NSArray<OAWorldRegion *> *potentiallyUsedMaps;
- @property (nonatomic) NSArray<OAWorldRegion *> * missingMaps;
- @property (nonatomic) NSArray<OAWorldRegion *> * mapsToUpdate;
```

They are already dropped from MissingMapsCalculator.

PR #5283, opened 2026-04-06
Marketing deep links testing

DmitrySvetlichny:


Sources/POI/DeepLinkAppRouter.swift
```diff
@@ -305,13 +326,26 @@ final class DeepLinkAppRouter: NSObject {
     }
     
     @discardableResult private func dismissAndPopToRoot() -> UINavigationController? {
+        guard let nav = prepareRootNavigationTransition() else { return nil }
```
tigrim, line 316: Why are there two methods if they can be merged into one?

Sources/POI/DeepLinkAppRouter.swift
```diff
@@ -289,7 +292,25 @@ final class DeepLinkAppRouter: NSObject {
     }
     
     func openNavigationScreen(appMode: OAApplicationMode) {
+        // Opening Route Info immediately mutates routing state and triggers route recalculation,so deeplinks wait until the app is active to match the regular map button flow.
+        guard UIApplication.shared.applicationState == .active else {
```
tigrim: ? I don’t think this is a good approach

DmitrySvetlichny: This is intentionally local because the issue was specific to Route Info opening too early during app activation. This screen immediately changes routing state and triggers route recalculation, so delaying only this deeplink until the app becomes active keeps the fix minimal and avoids changing the global deeplink pipeline.

tigrim: We need to investigate the issue. It might require reworking the mechanism globally

tigrim (approved):

tigrim (approved):

PR #5307, opened 2026-04-10
Fix double CarPlay preview in newRouteIsCalculated() (LLM)

RZR-UA:
https://github.com/osmandapp/OsmAnd/issues/23853


RZR-UA:
Will be replaced with better fix by @tigrim

PR #5342, opened 2026-04-22
pause spoken audio when voice used, pause spoken audio switch added

### LINKED ISSUE #5134: Voice prompt pause music/podcast with CarPlay
It would be nice to have the possibility to pause music or podcasts when there is voice prompt navigation with CarPlay.
This feature seems to be available on android but not on iOS.

Vladyslav-lys:


Resources/Localizations/en.lproj/Localizable.strings
tigrim, line 1: use
[_audioSession setActive:NO
            withOptions:AVAudioSessionSetActiveOptionNotifyOthersOnDeactivation
                  error:nil];
 for deactivation

PR #5360, opened 2026-04-28
hide public transport headers if needed

### LINKED ISSUE #5249: Wrong headers above Public transport routes
### How to reproduce
* Build a **Public transport** route.

### Actual result

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

* The group headers from the first state of **Navigation UI** is still visible after routes are calculated: **Previous** route, **Map markers**.

### Expected result
* Hide group headers when routes are visible

**Environment**
OsmAnd Version: 5350
iOS version: 26.4

Vladyslav-lys:


Sources/Controllers/RouteInfoMenu/OARouteInfoView.mm
```diff
@@ -2136,17 +2136,27 @@ - (NSString *) getTitleForSection:(NSInteger) section
         return nil;
     
     if (section == _prevRouteSection)
-        return OALocalizedString(@"previous_route");
+        return [self hidePublicTransportHeadersFor:section] ? nil : OALocalizedString(@"previous_route");
```
tigrim: Something seems off here. Why does each cell have its own section instead of having one section with multiple cells?

tigrim (approved):

PR #5376, opened 2026-05-04
Profile units used in other profile settings

DmitrySvetlichny:


Sources/Controllers/Settings/NavigationSettings/OARouteParametersViewController.mm
```diff
@@ -485,9 +483,7 @@ - (void) setupSelectRouteRecalcDistance:(NSMutableArray *)tableSection
 {
     double recalcDist = [_settings.routeRecalculationDistance get:self.appMode];
     recalcDist = recalcDist == 0 ? [OARoutingHelper getDefaultAllowedDeviation:self.appMode posTolerance:[OARoutingHelper getPosTolerance:0]] : recalcDist;
-    NSString *descr = recalcDist == -1
-            ? OALocalizedString(@"rendering_value_disabled_name")
-    : [OAOsmAndFormatter getFormattedDistance:recalcDist withParams:[OsmAndFormatterParams noTrailingZeros]];
+    NSString *descr = recalcDist == -1 ? OALocalizedString(@"rendering_value_disabled_name") : [OAOsmAndFormatter getFormattedDistance:recalcDist  mode:self.appMode withParams:[OsmAndFormatterParams noTrailingZeros]];
```
tigrim: Remove the extra space

Sources/Controllers/Settings/NavigationSettings/DetailedTrackGuidanceViewController.swift
```diff
@@ -157,7 +157,7 @@ final class DetailedTrackGuidanceViewController: OABaseSettingsViewController {
             if newDistanceThreshold != distanceThreshold {
                 distanceThreshold = newDistanceThreshold
                 isSettingsChanged = true
-                cell.valueLabel.text = OAOsmAndFormatter.getFormattedDistance(Float(distanceThreshold))
+                cell.valueLabel.text = OAOsmAndFormatter.getFormattedDistance(Float(distanceThreshold), mode: appMode, with: nil)
```
tigrim: with: nil ?

DmitrySvetlichny: Yes, nil is intentional here. It preserves the previous default formatter params; the only change is passing appMode so distance units are formatted for the edited profile.

tigrim: Non-obvious parameter in Swift (what does “with” actually mean?)

tigrim: Refactor it to remove the need to pass with: nil

tigrim (approved):

PR #5411, opened 2026-05-21
Align stroke thickness on turn arrows #4787

### LINKED ISSUE #4787: Align stroke thickness on turn arrows
### Actual result

Image

* Turn and Lane arrows has different stroke thickness

### Expected result

* All turn arrows should have the same stroke thickness
- [x] CarPlay
- [x] Device

**Environment**
OsmAnd Version: 5150
iOS version: 26

vitaliy-sova-ios:


Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -852,9 +852,7 @@ - (void)onLocationUpdate
                     [drawable setTurnImminent:nextNextDirInfo.imminent
                             deviatedFromRoute:deviatedFromRoute];
                     drawable.textFont = [UIFont scaledSystemFontOfSize:16 weight:UIFontWeightSemibold];
-                    CGFloat size = MAX(drawable.pathForTurn.bounds.origin.x + drawable.pathForTurn.bounds.size.width,
-                                       drawable.pathForTurn.bounds.origin.y + drawable.pathForTurn.bounds.size.height);
-                    drawable.frame = CGRectMake(0, 0, size, size);
+                    drawable.frame = CGRectMake(0, 0, UIScreen.mainScreen.scale * 16, UIScreen.mainScreen.scale * 16);
```
tigrim: what does 16 mean, and why is UIScreen.mainScreen.scale needed in CarPlay?

vitaliy-sova-ios: 16 is the turn icon size in points, which is set just below in line 882.

UIScreen.mainScreen.scale is used to render the image at the device's native pixel resolution before downscaling to 16pt. Without this, CarPlay screens with different zoom levels would display a blurry image or a disproportionately thin border.

tigrim: 1) Move 16 to a constant.
2) Why are you using UIScreen.mainScreen.scale in CarPlay? You should use the scale of the CarPlay screen instead - they can be different.

vitaliy-sova-ios: Before
before1
before2






After
after1
after2
after3
after4

tigrim (approved):

PR #5419, opened 2026-05-25
Fix incorrect map centering after double-tap zoom gesture

### LINKED ISSUE #5365: Incorrect map centering after double-tap gesture and tapping My Location button
## Description
Found in https://github.com/osmandapp/OsmAnd-iOS/issues/5158

### How to reproduce
* Perform a double-tap with a move up or down gesture in the right corner of the screen.
* Afterward, tap the **My Location** button.

### Actual result
<table style="width:100%">
  <tr>
    <th>Video</th>
    <th>Screenshot</th>
  </tr>
  <tr>
    <td><video src="https://github.com/user-attachments/assets/37bf82ec-5454-4a31-8838-c96d242c48c6"  width="300" height="auto"/> </td>
    <td> </td>
  </tr>
</table>   

* The map moves to the current location, but it is centered on the section of the screen where the double-tap gesture was performed.

### Expected result

* The map moves to the current location and is centered correctly in the middle or bottom part of the screen depend on **Location position on the screen option**

**Environment**
OsmAnd Version: OsmAnd Maps 5371, released 29.04.2026
iOS version: 26.5

aleksandr-tata:
isLastMultiGesture didn't account for _zoomingByTapGesture, so restorePreviousTarget never ran after the double-tap + drag gesture. This left fixedPixel pinned to the touch point, causing My Location to center the map there instead of the viewport center.

### DIFF
--- a/Sources/Controllers/Map/OAMapViewController.mm
+++ b/Sources/Controllers/Map/OAMapViewController.mm
@@ -1056,10 +1056,11 @@ static char kMapSourceUpdateQueueKey;
 
 - (BOOL) isLastMultiGesture
 {
-    return (_movingByGesture && !_zoomingByGesture && !_rotatingByGesture && !_zoomingByTapGesture)
-        || (!_movingByGesture && _zoomingByGesture && !_rotatingByGesture && !_zoomingByTapGesture)
-        || (!_movingByGesture && !_zoomingByGesture && _rotatingByGesture && !_zoomingByTapGesture)
-        || (!_movingByGesture && !_zoomingByGesture && !_rotatingByGesture && _zoomingByTapGesture);
+    NSInteger activeGestures = (_movingByGesture ? 1 : 0)
+        + (_zoomingByGesture ? 1 : 0)
+        + (_rotatingByGesture ? 1 : 0)
+        + (_zoomingByTapGesture ? 1 : 0);
+    return activeGestures == 1;
 }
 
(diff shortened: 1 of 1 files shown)


Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -1056,9 +1056,10 @@ - (BOOL) isTargetChanged
 
 - (BOOL) isLastMultiGesture
 {
-    return (_movingByGesture && !_zoomingByGesture && !_rotatingByGesture)
-    	|| (!_movingByGesture && _zoomingByGesture && !_rotatingByGesture)
-    	|| (!_movingByGesture && !_zoomingByGesture && _rotatingByGesture);
+    return (_movingByGesture && !_zoomingByGesture && !_rotatingByGesture && !_zoomingByTapGesture)
+        || (!_movingByGesture && _zoomingByGesture && !_rotatingByGesture && !_zoomingByTapGesture)
+        || (!_movingByGesture && !_zoomingByGesture && _rotatingByGesture && !_zoomingByTapGesture)
+        || (!_movingByGesture && !_zoomingByGesture && !_rotatingByGesture && _zoomingByTapGesture);
```
tigrim: The method looks overly complex for checking a single condition that exactly one gesture state is active. It could be simplified by counting active gesture flags e.g. activeGestures == 1

tigrim (approved):

PR #5425, opened 2026-05-27
Fixes stuttering/jerky behavior during GPS navigation (issue #5355)

### LINKED ISSUE #5355: Stuttering / jerky map movement during navigation
### Description

During navigation or navigation simulation, the map movement occasionally becomes jerky ("stutters" or "jumps"). This occurs relatively rarely - approximately twice per minute—and does not appear to follow a consistent pattern.

The issue persists regardless of visual settings such as "Globe view" (Round Earth) or 3D mode. It feels like a brief loss of frame rate or a synchronization hiccup between the location updates and map rendering.

### Steps to reproduce

Start a navigation or a simulation along any route > Observe the map movement for at least 1-2 minutes > Pay attention to the smoothness of the movement of the location icon

### Actual result

The movement is generally smooth, but periodically (about twice a minute) the map "jumps" or stutters, breaking the visual fluidity of the navigation

<video src="https://github.com/user-attachments/assets/9a55786b-b1bb-4eb2-8bbf-a9273c4c7554"></video>
</details> 

### Expected result

- [ ] The location icon should move fluidly and consistently at a high frame rate without any random jerks or "skips" during the entire navigation session

### Your Environment
OsmAnd Version: OsmAnd Maps 5368, released 23.04.26
Android/iOS version: iOS 26.5
Device model: iPhone 15

aleksandr-tata:
OAMyPositionLayer: Previously every location update unconditionally cancelled all marker animations. When a compass heading update arrived mid-position-animation, it interrupted the smooth arrow movement, causing a visible jerk. Now, if a heading-only update (animationDuration <= 0) arrives while a position animation is already running, only the azimuth animation is cancelled and restarted — the position animation continues uninterrupted.

OAMapViewTrackingUtilities: Added NAV_ANIMATION_TIME (1s) as a minimum for movingTime in navigation mode. Previously movingTime = timeDiff could be 0 on the first GPS update (no previous location) or shorter than 1s on fast GPS intervals, causing the camera to snap rather than animate smoothly.

### DIFF
--- a/Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
+++ b/Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
@@ -248,56 +248,40 @@ typedef enum {
             [OARootViewController.instance.mapPanel.mapViewController.mapLayers.myPositionLayer setMyLocationCircleRadius:(horizontalAccuracy)];
         });
 
-        BOOL hasOngoingPositionAnim = animationDuration <= 0
-            && _mapView.mapMarkersAnimator->getCurrentAnimation(marker, OsmAnd::Animator::AnimatedValue::Target) != nullptr;
+        BOOL isAnimatedMove = animationDuration > 0;
+        BOOL targetAnimRunning = _mapView.mapMarkersAnimator->getCurrentAnimation(marker, OsmAnd::Animator::AnimatedValue::Target) != nullptr;
+        BOOL keepRunningTargetAnim = !isAnimatedMove && targetAnimRunning;
+        BOOL shouldAnimateAzimuth = isAnimatedMove || keepRunningTargetAnim;
 
-        if (!hasOngoingPositionAnim)
-            _mapView.mapMarkersAnimator->cancelAnimations(marker);
-        else
-            _mapView.mapMarkersAnimator->cancelCurrentAnimation(marker, OsmAnd::Animator::AnimatedValue::Azimuth);
+        double bearingDir = OsmAnd::Utilities::normalizedAngleDegrees(bearing);
+        double headingDir = OsmAnd::Utilities::normalizedAngleDegrees(heading);
 
-        if (animationDuration > 0)
+        if (isAnimatedMove)
         {
+            _mapView.mapMarkersAnimator->cancelCurrentAnimation(marker, OsmAnd::Animator::AnimatedValue::Target);
             _mapView.mapMarkersAnimator->animatePositionTo(marker, target31, animationDuration,  OsmAnd::Animator::TimingFunction::Linear);
-            if (marker->model3D != nullptr)
-            {
-                _mapView.mapMarkersAnimator->animateModel3DDirectionTo(marker, OsmAnd::Utilities::normalizedAngleDegrees(bearing), kRotateAnimationTime, OsmAnd::Animator::TimingFunction::Linear);
-                [_mapView setMyLocationSectorDirection:(OsmAnd::Utilities::normalizedAngleDegrees(heading))];
-            }
-            else
-            {
-                if (iconKey)
-                    _mapView.mapMarkersAnimator->animateDirectionTo(marker, iconKey, OsmAnd::Utilities::normalizedAngleDegrees(bearing), kRotateAnimationTime,  OsmAnd::Animator::TimingFunction::Linear);
-            }
         }
-        else if (!hasOngoingPositionAnim)
+        else if (!keepRunningTargetAnim)
         {
             marker->setPosition(target31);
             [_mapView setMyLocationCirclePosition:(target31)];
-
-            if (marker->model3D != nullptr)
-            {
-                marker->setModel3DDirection(OsmAnd::Utilities::normalizedAngleDegrees(bearing));
-                [_mapView setMyLocationSectorDirection:(OsmAnd::Utilities::normalizedAngleDegrees(heading))];
-            }
-            else
-            {
-                if (iconKey)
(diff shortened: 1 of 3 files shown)


aleksandr-tata:
https://github.com/user-attachments/assets/441b5a21-c6b1-44e9-bede-c9deb4af8060

https://github.com/user-attachments/assets/1640e67f-8663-4559-bbec-b3208e0d0e9a

aleksandr-tata:
https://github.com/user-attachments/assets/bd639747-cbb5-4545-87e3-6a081b63be0c

The same navigation route on the Android device and simulation at 30 km/h speed

Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
```diff
@@ -248,7 +248,14 @@ - (void) updateLocation:(OsmAnd::PointI)target31 animationDuration:(float)animat
...
         });
 
-        _mapView.mapMarkersAnimator->cancelAnimations(marker);
+        BOOL hasOngoingPositionAnim = animationDuration <= 0
+            && _mapView.mapMarkersAnimator->getCurrentAnimation(marker, OsmAnd::Animator::AnimatedValue::Target) != nullptr;
+
+        if (!hasOngoingPositionAnim)
+            _mapView.mapMarkersAnimator->cancelAnimations(marker);
+        else
+            _mapView.mapMarkersAnimator->cancelCurrentAnimation(marker, OsmAnd::Animator::AnimatedValue::Azimuth);
```
alex-dev-neo: It is not clear why Azimuth animation cancelled here

Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
```diff
@@ -279,6 +286,19 @@ - (void) updateLocation:(OsmAnd::PointI)target31 animationDuration:(float)animat
                     marker->setOnMapSurfaceIconDirection(iconKey, OsmAnd::Utilities::normalizedAngleDegrees(bearing));
             }
         }
+        else
+        {
+            if (marker->model3D != nullptr)
```
alex-dev-neo: Why new directionTo animation starts with old ongoing position animation?

Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
```diff
@@ -248,7 +248,14 @@ - (void) updateLocation:(OsmAnd::PointI)target31 animationDuration:(float)animat
             [OARootViewController.instance.mapPanel.mapViewController.mapLayers.myPositionLayer setMyLocationCircleRadius:(horizontalAccuracy)];
         });
 
-        _mapView.mapMarkersAnimator->cancelAnimations(marker);
+        BOOL hasOngoingPositionAnim = animationDuration <= 0
```
alex-dev-neo: What if location is already significantly changed at this point?

Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
aleksandr-tata: We always start a new azimuth animation below (to the updated compass direction), so the current one must be cancelled first to avoid stacking multiple animations on the same value. In the !hasOngoingPositionAnim branch this is done implicitly by cancelAnimations which cancels everything. In the hasOngoingPositionAnim branch we cancel only azimuth explicitly — to replace it with the new direction while leaving the position animation untouched.

Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
aleksandr-tata: Target and Azimuth are independent animated values on the marker and are designed to run concurrently. The arrow should simultaneously move toward the new GPS position and rotate to face the new compass direction — this is the intended behavior. This is the same pattern Android uses in PointLocationLayer: updateMarkerPosition and updateMarkerBearing are completely independent and always run in parallel without interfering with each other.

Sources/Controllers/Map/Layers/OAMyPositionLayer.mm
aleksandr-tata: This branch is only reached when animationDuration <= 0, which exclusively means a compass/heading update — the GPS position has not changed in this call at all. The ongoing position animation was started by the previous GPS update and is still animating toward the correct latest position. When the next GPS update arrives, updateLocation will be called with animationDuration > 0 and will cancel and restart the position animation as usual.

alex-dev-neo: 1. Do not copy-paste AI answers in response to PR questions
2. "when animationDuration <= 0, which exclusively means a compass/heading update" - Did you figure this out yourself or just trust the AI's answer?

aleksandr-tata: I agree, "exclusively" is not correct. animationDuration can also be 0 when "Animate own position" option is off, or when more than 5 seconds pass between location updates — so animationDuration=0 doesn't only mean a heading update.

The actual problem: the marker animates smoothly from one location to the next, and cancelAnimations stops every animation. When only the heading changes, the coordinate stays the same (animationDuration=0), and the old code still ran cancelAnimations and then setPosition. That cancelled the running position animation and jumped the marker straight to its end point — the visible jump. In the logs this happens constantly: animationDuration=0 while a Target (position) animation is still running, because heading updates keep arriving between location updates during navigation.

Why the fix is safe: we skip setPosition only when a position animation is already running. It's already moving the marker toward the latest point, so we let it finish instead of moving the marker there directly. A real position change always comes with animationDuration > 0, which restarts the animation — so a real move is never skipped.

PR #5453, opened 2026-06-05
fix testAmPm test (OAExternalTimeFormatter localeUsingLatinDigits)

tigrim:


RZR-UA:
"ar" is underspecified.

_It fixes only the language, not the region or numbering system. Apple/ICU is then free to choose defaults for Arabic formatting. Those defaults can differ by OS version, platform build, ICU/CLDR data, and sometimes locale preference handling._

So, the complication requires another approach to fix.

PR #5456, opened 2026-06-08
Fix Disable Context menu or What's Here menu opening while gesture is performing

### LINKED ISSUE #5365: Incorrect map centering after double-tap gesture and tapping My Location button

aleksandr-tata:
Fix Disable Context menu or What's Here menu opening while gesture is performing

### DIFF
--- a/Sources/Controllers/Map/OAMapViewController.mm
+++ b/Sources/Controllers/Map/OAMapViewController.mm
@@ -1063,6 +1063,11 @@ static char kMapSourceUpdateQueueKey;
     return activeGestures == 1;
 }
 
+- (BOOL) isMapGestureInProgress
+{
+    return _movingByGesture || _zoomingByGesture || _rotatingByGesture || _zoomingByTapGesture;
+}
+
 - (void) storeTargetPosition:(UIGestureRecognizer *)recognizer scheduleRestore:(BOOL)scheduleRestore
 {
     if (![self isTargetChanged])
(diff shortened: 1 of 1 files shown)


Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -1063,6 +1063,11 @@ - (BOOL) isLastMultiGesture
     return activeGestures == 1;
 }
 
+- (BOOL) isMapGestureInProgress
```
tigrim: I think this approach would be more reliable.
```
 - (BOOL)isGestureRecognizerActive:(UIGestureRecognizer *)recognizer
{
    return recognizer.state == UIGestureRecognizerStateBegan
        || recognizer.state == UIGestureRecognizerStateChanged
        || recognizer.state == UIGestureRecognizerStateEnded;
}

- (BOOL)isMapGestureInProgress
{
    return _movingByGesture
        || _zoomingByGesture
        || _rotatingByGesture
        || _zoomingByTapGesture
        || [self isGestureRecognizerActive:_grMove]
        || [self isGestureRecognizerActive:_grZoom]
        || [self isGestureRecognizerActive:_grRotate]
        || [self isGestureRecognizerActive:_grZoomDoubleTap]
        || [self isGestureRecognizerActive:_grElevation];
}
```

tigrim: remove space

aleksandr-tata: fixed

aleksandr-tata: Fixed

tigrim (approved):

PR #5478, opened 2026-06-19
Change default appearance track folders

### LINKED ISSUE #5080: Change default appearance for Track folders
## Description 
We need to add ability to set Default appearance for Track folders. Using the same UI is using for Change appearance in selection mode.

## Track

- [ ] Add action Change default appearance for folder context menu
- - [ ] Use icon –  

## Default appearance

DmitrySvetlichny:


tigrim (commented):
### 1. Backup should preserve `nil` values (tech debt)

**`GpxAppearanceInfo.init(dataItem:)`**

Android stores raw appearance parameters as nullable values. The iOS implementation replaces missing values with `0`, `false`, or `.none`, which causes them to become explicit overrides after restore and prevents folder defaults from being applied.

**Fix:** Make the corresponding fields optional, omit absent values from the JSON, and do not apply them during restore.

### 2. Do not override appearance after move

**`OAGPXUIHelper.copyGPXToNewFolder`**

`GpxDbHelper.rename()` already applies folder defaults using `updateExistingValues: false`. The subsequent call to `applyFolderDefaultAppearance()` overwrites individual appearance settings, diverging from the Android/shared contract.

**Fix:** Remove the additional call after `rename()` and rely on the shared implementation.

### 3. Wait for all DB callbacks to complete

**`ChangeTracksAppearanceTask`**

`execute()` invokes completion immediately after `doInBackground()`, even though `getItem(...callback:)` may finish later. As a result, the screen can be dismissed and the map refreshed before all database updates are completed.

**Fix:** Track pending callbacks using `async/await` or a `DispatchGroup`, and call completion only after the last `updateDataItem()` finishes.

### 4. Verify split cache invalidation

**`saveFolderDefaultAppearance` / `ChangeTracksAppearanceTask`**

Android calls `SelectedGpxFile.resetSplitProcessed()` whenever split-related settings change. On iOS, only a map refresh notification is sent.

**Fix:** Verify that the notification reliably triggers split data recalculation. If not, add an explicit reset for affected selected tracks.

### 5. Remove redundant DB updates

**`applyFolderDefaultAppearance()` and callers**

The shared `add()` and `rename()` implementations already apply defaults and persist the item. Some current flows perform two or even three consecutive database writes for the same operation.

**Fix:** Keep default application and persistence in a single layer, preferably inside the shared `GpxDbHelper` implementation.

Sources/Helpers/OAGPXUIHelper.mm
```diff
@@ -600,6 +600,7 @@ - (void)copyGPXToNewFolder:(NSString *)newFolderName
                     OASGpxDataItem *gpx = [[OAGPXDatabase sharedDb] getGPXItem:newStoringFullPath];
                     if (gpx)
                     {
+                        [gpx applyFolderDefaultAppearance];
```
tigrim: `GpxDbHelper.rename()` already applies the destination folder defaults with `updateExistingValues: false`, preserving the track’s individual appearance settings. Calling `applyFolderDefaultAppearance()` here afterwards overwrites those settings and diverges from the Android/shared behavior. Please remove this extra call and rely on the shared rename logic.

Sources/Controllers/MyPlaces/ChangeTracksAppearanceTask.swift
```diff
@@ -87,7 +87,7 @@ final class ChangeTracksAppearanceTask: NSObject {
         for parameter in GpxParameter.companion.getAppearanceParameters() {
             if data.shouldResetParameter(parameter), let gpxFile = gpxFile {
                 item.readGpxAppearanceParameter(gpxFile: gpxFile, parameter: parameter)
-            } else if let value: Any = data.getParameter(for: parameter) {
+            } else if let value = data.rawParameter(for: parameter) {
```
tigrim, line 80: `getItem(file:callback:)` can complete asynchronously, but `execute()` invokes its completion as soon as `doInBackground()` returns. The new folder flow may therefore dismiss the screen and refresh the map while some callbacks are still updating the database. Please track pending reads (for example with structured concurrency or a `DispatchGroup`) and invoke completion only after the final `updateDataItem()` has finished. Also avoid updating the same cached item both synchronously and again from the callback.

Sources/Controllers/MyPlaces/TracksChangeAppearanceViewController.swift
```diff
@@ -423,52 +478,149 @@ final class TracksChangeAppearanceViewController: OABaseNavbarViewController {
...
+                dirItem.setParameter(parameter: parameter, value: nil)
+            } else {
+                dirItem.setParameter(parameter: parameter, value: folderDefaultValue(for: parameter))
+            }
         }
         
-        return data
+        GpxDbHelper.shared.updateDataItem(item: dirItem)
+        if updateExisting {
+            let task = ChangeTracksAppearanceTask(data: data, items: tracks) { [weak self] in
```
tigrim, line 543: Android explicitly calls `SelectedGpxFile.resetSplitProcessed()` when split-related settings change, including the “apply only to new” path. Here iOS only posts a map refresh notification after saving. Please verify that this notification invalidates all cached split data; otherwise reset the split state for the affected selected tracks before refreshing the map.

Sources/Helpers/SharedLibHelpers/GpxDataItemExtension.swift
```diff
@@ -313,13 +330,26 @@ extension GpxDataItem {
...
+        guard let parentFile = file.getParentFile() else { return }
+        let dirItem = GpxDbHelper.shared.getGpxDirItem(file: parentFile)
+        for parameter in GpxParameter.companion.getAppearanceParameters() {
+            let defaultValue = dirItem.getParameter(parameter: parameter)
+            if defaultValue != nil {
+                setParameter(parameter: parameter, value: defaultValue)
+            }
+        }
+
+        GpxDbHelper.shared.updateDataItem(item: self)
```
tigrim: The shared `GpxDbHelper.add()` and `rename()` paths already apply folder defaults and persist the item. Persisting again inside this helper, followed by another update in some callers (for example route saving), results in two or three consecutive writes for one operation. Please keep default application and persistence in a single layer, preferably the shared `GpxDbHelper`, and remove the redundant updates from iOS callers.

DmitrySvetlichny:
Verified. iOS does not have a SelectedGpxFile.resetSplitProcessed() equivalent. The map refresh notification triggers GPX layer rebuild and clears split labels/cache, so no additional invalidation is needed unless manual testing shows stale split labels.

Sources/Controllers/MyPlaces/TracksChangeAppearanceViewController.swift
DmitrySvetlichny: Verified. iOS does not have a SelectedGpxFile.resetSplitProcessed() equivalent. Split labels are cached in OAGPXLayer, and updateGpxTracksOnMapObservable triggers GPX layer refresh/rebuild. The refresh path reaches refreshStartFinishPoints(), which clears split labels before recalculation, so no extra reset is needed here.

PR #5497, opened 2026-06-30
Wiki UI review fixes

vitaliy-sova-ios:


Sources/Views/WikipediaContextMenuView.swift
```diff
@@ -126,6 +132,9 @@ final class WikipediaContextMenuView: UIView {
 
     private func setupActions() {
         actionButton.addTarget(self, action: #selector(didTapActionButton), for: .touchUpInside)
+        actionButton.addTarget(self, action: #selector(didTapDownActionButton), for: .touchDown)
+        actionButton.addTarget(self, action: #selector(resetButtonState), for: .touchUpOutside)
+        actionButton.addTarget(self, action: #selector(resetButtonState), for: .touchCancel)
```
tigrim: It's not clear why we're overriding the native behavior

Sources/Controllers/TargetMenu/OATargetInfoViewController.mm
```diff
@@ -1563,6 +1563,8 @@ - (UITableViewCell *) tableView:(UITableView *)tableView cellForRowAtIndexPath:(
         };
         
         info.height = [cell calculateHeightIn:tableView.frame.size.width];
+        
+        cell.backgroundColor = _contentColor;
```
tigrim, line 1567: Did it not affect the cell with the coordinates?

tigrim: ?

Sources/Views/WikipediaContextMenuView.swift
tigrim: actionButton.addTarget(self, action: #selector(resetButtonState), for:  .touchUpOutside, .touchCancel])

Sources/Views/WikipediaContextMenuView.swift
```diff
@@ -139,6 +148,27 @@ final class WikipediaContextMenuView: UIView {
...
+        config?.background.backgroundColor = .buttonBgColorTap
+        config?.baseForegroundColor = .buttonTextColorPrimary
+        
+        UIView.performWithoutAnimation {
+            actionButton.configuration = config
+        }
+    }
+    
+    @objc private func resetButtonState() {
+        var config = self.actionButton.configuration
```
tigrim: self?

tigrim (approved):

PR #5505, opened 2026-07-03
Import tracks UI & search for the nearest points (KDTree)

### LINKED ISSUE #3182: iOS tracks imported from GPX seen as unique track
### Description

On iOS when importing a GPX with different tracks on it the behaviour is very different from Android  

1. on iOS I can’t choose witch track I want to import from the GPX 
2. On iOS I see a single track with different segments but I can’t modify the appearance of each segment. It modifies the appearance of all the track
3. On iOS I can’t make visible only one of the track of the GPX

I attach the screenshot of iOS and what would be expected in Android 

### Steps to reproduce

Import the attached GPX with iOS device
Import attached GPX with Android device and check the different behaviour. 

### Actual result

On iOS the GPX is seen as single track with segments. 
on Android (as I would expect) is seen as different tracks. 

<table style="width:100%">
  <tr>
<th> </td>
  </tr>
</table>

On iOS I can’t modify appearance of each segment as I can do in Android. It modifies the whole track. 

<table style="width:100%">
  <tr>
<th> </td>
  </tr>
</table>

### Expected result

I would expect the same behaviour as Android as is the best user experience 

<table style="width:100%">
  <tr>
<th> </td>
  </tr>
</table>

**WARNING** Crash-Logs **MAY** contain information you deem sensitive.
Review this **CAREFULLY** before posting your issue!

```
OsmAnd Version: 4.5
Android/iOS version: IOS 17
Device model: iPhone 13
Crash-Logs: ?
```

vitaliy-sova-ios:


Sources/Controllers/MyPlaces/Import/Tasks/TrackPointsKDIndex.swift
```diff
@@ -0,0 +1,168 @@
...
+            let p = points[i]
+            minLat = min(minLat, p.lat)
+            maxLat = max(maxLat, p.lat)
+            minLon = min(minLon, p.lon)
+            maxLon = max(maxLon, p.lon)
+        }
+        return GeoBBox(minLat: minLat, maxLat: maxLat, minLon: minLon, maxLon: maxLon)
+    }
+
+    func minHaversineDistance(lat: Double, lon: Double) -> Double {
```
tigrim: Consider reusing the existing project distance API instead of calling OAMapUtils.getDistance directly, but keep the allocation-free coordinate-based path for KD-tree hot loops. If haversineDistanceInMetersTo requires creating CLLocation objects per candidate point, it may erase part of the KD-tree performance win

Sources/Controllers/MyPlaces/Import/Tasks/TrackPointsKDIndex.swift
```diff
@@ -0,0 +1,168 @@
...
+    }
+    
+    func nearestTrackIndex(lat: Double, lon: Double) -> Int? {
+        guard let root else { return nil }
+
+        var bestDist = Double.greatestFiniteMagnitude
+        var bestTrackIndex = -1
+        var bestPointIndex = Int.max
+
+        @inline(__always)
```
tigrim: @inline(__always) looks unnecessary here. These are small local helpers and the optimizer should inline them when profitable. Unless profiling showed a measurable win, prefer removing the attribute to avoid forcing the optimizer and making the code look more performance-fragile than it is

Sources/Controllers/MyPlaces/Import/Tasks/TrackPointsKDIndex.swift
```diff
@@ -0,0 +1,168 @@
...
+        func isBetter(_ d: Double, _ t: Int, _ p: Int) -> Bool {
+            if d < bestDist { return true }
+            if d > bestDist { return false }
+            if t < bestTrackIndex { return true }
+            if t > bestTrackIndex { return false }
+            return p < bestPointIndex
+        }
+
+        @inline(__always)
```
tigrim: up

Sources/Controllers/MyPlaces/Import/Tasks/TrackPointsKDIndex.swift
tigrim: ?

tigrim (commented):
fix merge conflicts

Sources/Controllers/MyPlaces/Import/Tasks/TrackPointsKDIndex.swift
```diff
@@ -0,0 +1,166 @@
...
+            maxLat = max(maxLat, p.lat)
+            minLon = min(minLon, p.lon)
+            maxLon = max(maxLon, p.lon)
+        }
+        return GeoBBox(minLat: minLat, maxLat: maxLat, minLon: minLon, maxLon: maxLon)
+    }
+
+    func minHaversineDistance(lat: Double, lon: Double) -> Double {
+        let clampedLat = min(max(lat, minLat), maxLat)
+        let clampedLon = min(max(lon, minLon), maxLon)
```
alex-dev-neo: Will it work correctly with longitude ±180° ?

Sources/Controllers/MyPlaces/Import/Tasks/TrackPointsKDIndex.swift
```diff
@@ -0,0 +1,166 @@
...
+            }
+        }
+
+        struct StackItem {
+            let node: KDNode
+        }
+
+        var stack: [KDNode] = [root]
+
+        while let node = stack.popLast() {
```
alex-dev-neo, line 180: Should it be cancellable?

PR #5516, opened 2026-07-07
UI fixes & fixed glitch regular map

vitaliy-sova-ios:


Sources/Plugins/Astronomy/StarMapViewController.swift
```diff
@@ -860,7 +860,13 @@ final class StarMapViewController: UIViewController, StarViewDelegate {
     }
 
     private func regularMapHeight() -> CGFloat {
-        view.bounds.width > view.bounds.height ? Layout.regularMapHeightLandscape : Layout.regularMapHeight
+        let isLandscape = OAUtilities.isLandscape()
+        
+        if OAUtilities.isIPad() {
+            let fraction: CGFloat = isLandscape ? 0.30 : 0.33
```
tigrim: What do these constants affect?

Sources/Plugins/Astronomy/StarMapViewController.swift
```diff
@@ -905,14 +911,20 @@ final class StarMapViewController: UIViewController, StarViewDelegate {
...
-    private func layoutRegularMapRenderer() {
+    private func layoutRegularMapRenderer(forceResize: Bool = false) {
         guard regularMapVisible,
               let mapView = currentMapViewController()?.view,
               mapView.superview === regularMapContainer else {
             return
         }
-        mapView.frame = regularMapContainer.bounds
-        mapView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        let target = regularMapContainer.bounds
```
tigrim: let bounds?

Sources/Plugins/Astronomy/StarMapViewController.swift
vitaliy-sova-ios: These constants determine what proportion of the screen's height a standard map occupies on an iPad.

tigrim (approved):

PR #5522, opened 2026-07-08
Fixed issues for Organized By feature

aleksandr-tata:
## Fixes: Organize by / smart folder screen

- **Activity icon size** — map (`mx_`) activity icons rendered edge-to-edge (no padding), looking larger than standard `ic_custom_*`. Added shared `UIImage.activityIcon(_:fallback:)` (pads glyph to 24pt in 30pt frame) and used it in all 4 call sites.
- **Navbar title `%@ %@`** — group screen used raw `getName()` (range template). Now uses `organizedGroupTitle(_:)`, same as row titles.
- **Row separator** — organized-group rows now use the 16pt right inset (were going edge-to-edge).
- **Refresh inside a group** — `onLoadFinished` unconditionally reset `organizedGroup = nil`, bouncing to the group list. Now re-resolves the group by id via `getSubgroupById` (as Android does), with title refresh.
- **Wrong footer stats in a group** — `getFolderAnalysis()` returns a cached (stale) analysis; footer also wasn't recomputed on enter. Now builds a fresh `TrackFolderAnalysis` (like Android) and `enter`/`leaveOrganizedGroup` call `updateData()`.

### DIFF
--- a/Sources/Controllers/MyPlaces/TracksViewController.swift
+++ b/Sources/Controllers/MyPlaces/TracksViewController.swift
@@ -547,7 +547,7 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
         row.title = organizedGroupTitle(organizedGroup)
         row.descr = formattedTracksCount(organizedGroup.getTrackItems().count)
         if organizedGroup.getType() == .activity {
-            row.icon = UIImage.activityIcon(organizedGroup.getIconName(), fallback: .templateImageNamed("ic_custom_activity"))
+            row.icon = UIImage.routeActivityIcon(organizedGroup.getIconName(), fallback: .templateImageNamed("ic_custom_activity"))
         } else {
             row.iconName = organizedGroup.getType().iconName
         }
--- a/Sources/Controllers/MyPlaces/TracksViewController.swift
+++ b/Sources/Controllers/MyPlaces/TracksViewController.swift
@@ -59,7 +59,7 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
     private let isVisibleKey = "isVisibleKey"
     private let isFullWidthSeparatorKey = "isFullWidthSeparatorKey"
     private let trackSortDescrKey = "trackSortDescrKey"
-    
+
     private var tableData = OATableDataModel()
     private var asyncLoader: TrackFolderLoaderTask?
     
@@ -171,8 +171,11 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
     private func onLoadFinished(folder: TrackFolder) {
         rootFolder = folder
         currentFolder = getTrackFolderByPath(currentFolderPath) ?? rootFolder
-        organizedGroup = nil
+        if let groupId = organizedGroup?.getId() {
+            organizedGroup = smartFolder?.getSubgroupById(subgroupId: groupId) as? OrganizedTracksGroup
+        }
         onRefreshEnd()
+        updateNavigationBarTitle()
         updateSearchResultsWithFilteredTracks()
         updateData()
     }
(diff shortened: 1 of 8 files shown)


Sources/Controllers/MyPlaces/TracksViewController.swift
```diff
@@ -898,14 +901,14 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
         let folderAnalysis: TrackFolderAnalysis
         if isSmartFolder {
             if let organizedGroup {
-                folderAnalysis = organizedGroup.getFolderAnalysis()
```
tigrim, line 901: getTotalTracksStatistics() now creates a new TrackFolderAnalysis every time for smart folders, organized groups, and regular folders. Previously, getFolderAnalysis() reused a cached analysis. This looks like a performance regression

Sources/Helpers/Extensions/Extensions.swift
```diff
@@ -70,6 +72,21 @@ extension UIImage {
...
+        guard padding > 0 else { return self }
+        let paddedSize = CGSize(width: size.width + padding * 2, height: size.height + padding * 2)
+        let renderer = UIGraphicsImageRenderer(size: paddedSize)
+        let padded = renderer.image { _ in
+            draw(in: CGRect(x: padding, y: padding, width: size.width, height: size.height))
+        }
+        return padded.withRenderingMode(renderingMode)
+    }
+
+    @objc static func activityIcon(_ iconName: String?, fallback: UIImage?) -> UIImage? {
```
tigrim: If this is needed for the task, move it to a separate extension and rename it to routeActivityIcon

Sources/Helpers/Extensions/Extensions.swift
```diff
@@ -70,6 +72,21 @@ extension UIImage {
     @objc func resizedTemplateImage(with size: CGFloat) -> UIImage? {
         OAUtilities.resize(self, newSize: CGSize(width: size, height: size))?.withRenderingMode(.alwaysTemplate)
     }
+
+    func withPadding(_ padding: CGFloat) -> UIImage {
```
tigrim: private?

Sources/Helpers/Extensions/Extensions.swift
```diff
@@ -70,6 +72,21 @@ extension UIImage {
     @objc func resizedTemplateImage(with size: CGFloat) -> UIImage? {
         OAUtilities.resize(self, newSize: CGSize(width: size, height: size))?.withRenderingMode(.alwaysTemplate)
     }
+
```
tigrim: Maybe we can simply make the UIImageView smaller instead of redrawing the image?

Sources/Helpers/Extensions/Extensions.swift
```diff
@@ -70,6 +72,21 @@ extension UIImage {
...
+        let renderer = UIGraphicsImageRenderer(size: paddedSize)
+        let padded = renderer.image { _ in
+            draw(in: CGRect(x: padding, y: padding, width: size.width, height: size.height))
+        }
+        return padded.withRenderingMode(renderingMode)
+    }
+
+    @objc static func activityIcon(_ iconName: String?, fallback: UIImage?) -> UIImage? {
+        guard let iconName, let mapIcon = mapSvgImageNamed("mx_\(iconName)") else { return fallback }
+        return mapIcon.withPadding(activityIconPadding)
```
tigrim: private enum RouteActivityIconStyle {
    static let padding: CGFloat = 3
}

Sources/Controllers/MyPlaces/TracksViewController.swift
aleksandr-tata: Fixed

Sources/Helpers/Extensions/Extensions.swift
aleksandr-tata: fixed

Sources/Helpers/Extensions/Extensions.swift
aleksandr-tata: fixed

Sources/Helpers/Extensions/Extensions.swift
aleksandr-tata: The Figma design defines a 30×30 slot, and it's the shared leftIconView size for all rows — resizing the view only for activities would break icon-column alignment with folders/tracks. The ic_custom_* icons are already 30×30 with transparent padding (glyph ~24pt), while mx_ icons render edge-to-edge. Padding matches them within the same slot, so we pad the image rather than resize the view.

Sources/Helpers/Extensions/Extensions.swift
aleksandr-tata: Done — moved the constant into a private RouteActivityIconStyle enum

tigrim (approved):

PR #5532, opened 2026-07-13
tracks duplicated in visible tracks fixed

### LINKED ISSUE #5482: Tracks duplicated in "Visible Tracks" list after renaming using accented characters (e.g., 'A' to 'Á') due to string normalization issue
### Description

Tracks seem to appear duplicated just by having a simple character with accent in the name.

Here's the very same track, named **A**, then named **Á**.
Haven't tested all the possibilities, though.

<table style="width:100%">
  <tr>
<th> </th>
 <th> </th>
  </tr>
</table>

### Steps to reproduce

- Create new track, name it something like "A"
- Check the Visible Tracks list: One track should be listed
- Rename that track adding an accent, like "Á"
- Check the Visible Tracks list: Track should double up

### Actual result

- Tracks get duplicated

### Expected result

- Tracks should appear only once

```
OsmAnd Version: 5.4.0
iOS version: 26.5
Device model: iPhone 13 mini
```

Vladyslav-lys:


tigrim (commented):
1) Normalization is fixed only in this screen/helper, not at the persisted visible-track state boundary.
File/function: OAAppSettings.m showGpx, hideGpx, updateGpx; OASelectedGPXHelper.mm buildGpxList

2) loadGpxTracks() hides duplicate DB rows nondeterministically.
File/function: MapSettingsGpxViewController.loadGpxTracks
Why dangerous: filtering before sorting keeps whichever duplicate OAGPXDatabase.getDataItems() returns first. If one row is stale after rename, the UI may operate on the wrong GpxDataItem path while the real duplicate remains in the DB

tigrim (commented):
1. TracksViewController.updateRenamedGpx ignores the result of gpxDB.renameCurrentFile(src, newFile: dest). If the DB rename returns false, the code still continues with handleDeletedGpxFile/addTrackItemToSmartFolder, which can leave the file, GPX DB, and smart folders out of sync. It would be safer to handle failure here similarly to OAGPXUIHelper.

2. The fix normalizes specific rename paths, but the persisted visible tracks state is still compared as raw strings in OAAppSettings showGpx/updateGpx/hideGpx and OASelectedGPXHelper/buildGpxList. If a path reaches mapSettingVisibleGpx from another entry point in a different Unicode form, the issue may come back. It may be better to normalize or compare paths at the mapSettingVisibleGpx boundary.

tigrim (commented):
showGpx now stores paths using decomposedStringWithCanonicalMapping, and hideGpx compares them with NSString.compare, but the public OACommonStringList.contains still performs an exact indexOfObject lookup. There are many direct membership checks like settings.mapSettingVisibleGpx.contains(track.gpxFilePath) If the DB/TrackItem returns the same path in precomposed form, the UI will treat an already visible track as hidden, bulk show may try to show it again, and delete/move flows may skip hideGpx.

 It would be better to either introduce a canonical-aware API for visible GPX membership or normalize/deduplicate inside OACommonStringList.contains, if that's safe for this preference

tigrim (commented):
OAAppSettings
```
- (BOOL)isGpxVisible:(NSString *)filePath
{
    for (NSString *visiblePath in _mapSettingVisibleGpx.get)
    {
        if ([visiblePath compare:filePath] == NSOrderedSame)
            return YES;
    }
    return NO;
}
```

settings.mapSettingVisibleGpx.contains(track.gpxFilePath) ->  settings.isGpxVisible(track.gpxFilePath)

[settings.mapSettingVisibleGpx.get containsObject:filePath] ->
[settings isGpxVisible:filePath]


--------

tigrim (commented):
Additionally, renameVisibleTracks should normalize the new value before storing it, since it currently writes directly into the array:
`visibleGpx[i] = modifiedString.decomposedStringWithCanonicalMapping`

tigrim (approved):

PR #5534, opened 2026-07-13
Astro explore UI review fix

vitaliy-sova-ios:


OsmAnd.xcodeproj/project.pbxproj
```diff
@@ -208,6 +208,7 @@
 		15CD7EA725B1C1A700BCB36A /* ic_small_time_start@2x.png in Resources */ = {isa = PBXBuildFile; fileRef = 15CD7EA425B1C1A700BCB36A /* ic_small_time_start@2x.png */; };
 		15CD7EAD25B1C1B900BCB36A /* ic_small_waypoints@3x.png in Resources */ = {isa = PBXBuildFile; fileRef = 15CD7EAA25B1C1B700BCB36A /* ic_small_waypoints@3x.png */; };
 		15CD7EAF25B1C1B900BCB36A /* ic_small_waypoints@2x.png in Resources */ = {isa = PBXBuildFile; fileRef = 15CD7EAC25B1C1B800BCB36A /* ic_small_waypoints@2x.png */; };
+		228EDAF170A7A81DF3A85783 /* UIImage+RouteActivityIcon.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8BD9FA9C2F7AD66558B5B407 /* UIImage+RouteActivityIcon.swift */; };
```
tigrim: ?

Sources/Plugins/Astronomy/search/StarMapMyDataViewController.swift
```diff
@@ -80,31 +80,34 @@ final class StarMapMyDataViewController: UIViewController {
...
-        return bar
+    
+    private lazy var searchController: UISearchController = {
+        let controller = UISearchController(searchResultsController: nil)
+        controller.delegate = self
+        controller.obscuresBackgroundDuringPresentation = false
+        controller.hidesNavigationBarDuringPresentation = OAUtilities.isIPhone()
+        controller.searchBar.placeholder = localizedString("shared_string_search")
+        controller.searchBar.delegate = self
+        controller.searchBar.searchBarStyle = .prominent
```
tigrim: `searchBarStyle = .prominent `is only available on iOS 26+. Calling it on earlier iOS versions will crash!

Sources/Plugins/Astronomy/search/StarMapMyDataViewController.swift
```diff
@@ -289,45 +313,23 @@ final class StarMapMyDataViewController: UIViewController {
...
-            self?.searchCancelButtonAction()
-        }, for: .touchUpInside)
-        
-        searchContainer.addArrangedSubview(cancelButton)
-    }
 
     // MARK: - Navigation
 
     private func setupNavigationBar() {
+        navigationController?.viewRespectsSystemMinimumLayoutMargins = false
```
tigrim: ?

Sources/Plugins/Astronomy/search/StarMapMyDataViewController.swift
```diff
@@ -340,13 +342,29 @@ final class StarMapMyDataViewController: UIViewController {
...
         suppressQueryDispatch = true
-        inlineSearchBar.text = searchState.query
+        searchController.searchBar.text = searchState.query
         suppressQueryDispatch = false
     }
+    
+    private func updateNavigationBarForSearchState() {
+        if isSearchActive {
+            navigationItem.searchController = searchController
+            navigationItem.rightBarButtonItem = nil
```
tigrim: searchController?.isActive = true ?

Sources/Plugins/Astronomy/search/StarMapMyDataViewController.swift
```diff
@@ -594,8 +611,49 @@ extension StarMapMyDataViewController: UISearchBarDelegate {
         searchState.query = searchText
         applyFiltersAndSort(scrollToTop: true)
     }
+    
+    func searchBarTextDidEndEditing(_ searchBar: UISearchBar) {
+        guard OAUtilities.isIPad() else { return }
```
tigrim: ?

Sources/Plugins/Astronomy/search/StarMapMyDataViewController.swift
```diff
@@ -594,8 +611,49 @@ extension StarMapMyDataViewController: UISearchBarDelegate {
...
+    func presentSearchController(_ searchController: UISearchController) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+            if !searchController.searchBar.isFirstResponder {
+                searchController.searchBar.becomeFirstResponder()
+            }
+        }
+    }
+    
+    func didDismissSearchController(_ searchController: UISearchController) {
+        navigationItem.searchController = nil
```
tigrim: searchController?.isActive = false ?
...
navigationItem.searchController = nil

Sources/Plugins/Astronomy/search/StarMapObjectContextMenuBuilder.swift
```diff
@@ -0,0 +1,73 @@
...
+
+        let direction = UIAction(
+            title: localizedString("astro_direction"),
+            image: AstroIcon.template(object.showDirection ? "ic_custom_target_direction_on" : "ic_custom_target_direction_off")
+        ) { _ in
+            object.showDirection.toggle()
+            if object.showDirection {
+                object.colorIndex = handler.onDirectionChanged(object, true)
+            } else {
+                _ = handler.onDirectionChanged(object, false)
```
tigrim: @discardableResult

Sources/Plugins/Astronomy/search/StarMapSearchEmptyView.swift
```diff
@@ -97,15 +99,29 @@ final class StarMapSearchEmptyView: UIView {
...
+        containerStack.layoutMargins = UIEdgeInsets(
+            top: Layout.contentPadding,
+            left: Layout.myDataContentPadding,
+            bottom: Layout.myDataContentPadding,
+            right: Layout.myDataContentPadding
+        )
+        
+        contentStack.layoutMargins = UIEdgeInsets(
+            top: 0,
+            left: config == .searchNoResults ? 0 : Layout.myDataContentPadding,
```
tigrim: to let

Sources/Plugins/Astronomy/search/StarMapSearchViewController.swift
```diff
@@ -112,74 +111,73 @@ final class StarMapSearchViewController: UIViewController {
...
-        bar.layoutMargins = .zero
-        bar.directionalLayoutMargins = .zero
-        return bar
+    private lazy var searchController: UISearchController = {
+        let controller = UISearchController(searchResultsController: nil)
+        controller.obscuresBackgroundDuringPresentation = false
+        controller.hidesNavigationBarDuringPresentation = OAUtilities.isIPhone()
+        controller.searchBar.placeholder = localizedString("shared_string_search")
+        controller.searchBar.delegate = self
+        controller.searchBar.searchBarStyle = .prominent
```
tigrim: up!

Sources/Plugins/Astronomy/search/StarMapSearchViewController.swift
```diff
@@ -1006,20 +1047,28 @@ final class StarMapSearchViewController: UIViewController {
...
-        } else {
-            searchState.replaceRecentChips(plugin.recentSearchChips)
-        }
+        let persisted = AstronomyPluginSettings.load().starMap.recentlyViewed
+            .map { StarMapRecentChip(label: $0.label, objectId: $0.id) }
+        searchState.replaceRecentChips(persisted)
     }
 
     private func renderRecentChips() {
+        func displayTitle(_ fullName: String) -> String {
```
tigrim, line 1067:
```
func displayTitle(_ fullName: String) -> String {
    let title = fullName.trimmingCharacters(in: .whitespacesAndNewlines)
    let maxLength = 10

    guard title.count > maxLength else { return title }

    return title.prefix(maxLength) + localizedString("shared_string_ellipsis")
}
```

Sources/Plugins/Astronomy/StarMapViewController.swift
```diff
@@ -1319,6 +1332,36 @@ final class StarMapViewController: UIViewController, StarViewDelegate {
     func isSearchRedFilterEnabled() -> Bool {
         starView.showRedFilter
     }
+    
+    func makeSearchObjectActionHandler() -> StarMapObjectActionHandler {
+        StarMapObjectActionHandler(
+            onFavoriteChanged: { [weak self] object, enabled in
+                guard let self else { return }
+                if enabled { settings.addFavorite(id: object.id) }
```
tigrim: code style

Sources/Plugins/Astronomy/StarMapViewController.swift
```diff
@@ -1319,6 +1332,36 @@ final class StarMapViewController: UIViewController, StarViewDelegate {
...
+                guard let self else { return object.colorIndex }
+                let colorIndex = enabled ? settings.addDirection(id: object.id) : object.colorIndex
+                if !enabled { settings.removeDirection(id: object.id) }
+                viewModel.updateSettings(settings)
+                starView.refreshObjects()
+                return colorIndex
+            },
+            onCelestialPathChanged: { [weak self] object, enabled in
+                guard let self else { return }
+                if enabled { settings.addCelestialPath(id: object.id) }
```
tigrim: up

Sources/Plugins/Astronomy/StarMapViewController.swift
```diff
@@ -1407,8 +1450,12 @@ final class StarMapViewController: UIViewController, StarViewDelegate {
         let leading = navigationController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: offscreenLeading)
         leftPanelLeadingConstraint = leading
 
+        let additionalPadding: CGFloat = OAUtilities.isWindowed() ? 0 : Layout.contentPadding * 2
```
tigrim, line 1461: isWindowed?

tigrim (commented):
- [x] The recently viewed header can end up with zero height here. `systemLayoutSizeFitting` is called on `recentChipsScroll`, but `UIScrollView` does not provide an intrinsic content height, and the chip container is constrained to `recentChipsScroll.frameLayoutGuide.height`. That means the measured height can resolve to `0`, so `tableHeaderView` is assigned a zero-height frame and the chips may disappear.

Could we give the header an explicit/minimum height, or measure the chip container/buttons directly instead? The previous cell-based implementation had an explicit minimum height, so preserving that would make this more stable.

- [x] This can leave My Data search in an inconsistent active state on iPad. `didDismissSearchController` always removes `navigationItem.searchController`, but `hideSearch(clearQuery:)` only runs when the search text is empty. If the search controller is dismissed with a non-empty query, `isSearchActive` remains `true`, the search button stays hidden, and the search controller is no longer attached to the navigation item.

Could we reset `isSearchActive` / update the nav bar state whenever the controller is dismissed, or avoid detaching `navigationItem.searchController` for the non-empty-query case?

Sources/Plugins/Astronomy/search/StarMapObjectContextMenuBuilder.swift
vitaliy-sova-ios: @discardableResult cannot be applied to closure properties.

Sources/Plugins/Astronomy/StarMapViewController.swift
vitaliy-sova-ios: isWindowed — iPad in Split View / Stage Manager (window is not full-screen).Additional indents are needed

Sources/Plugins/Astronomy/search/StarMapObjectContextMenuBuilder.swift
tigrim: How can we avoid having to write _ = then?

Sources/Plugins/Astronomy/search/StarMapObjectContextMenuBuilder.swift
```diff
@@ -0,0 +1,73 @@
...
+            }
+            onStateChanged()
+        }
+
+        let path = UIAction(
+            title: localizedString("astro_path"),
+            image: AstroIcon.template(object.showCelestialPath ? "ic_custom_target_path_on" : "ic_custom_target_path_off")
```
tigrim: AstroIcon.template(object.showCelestialPath ? .icCustomTargetPathOn : .icCustomTargetPathOff)

tigrim (approved):

PR #5536, opened 2026-07-14
Context Menu optimization - open POI context menu without blocking the main thread

### LINKED ISSUE #5337: Context menu opens twice as slowly on iOS compared to Android
### Description

### Steps to reproduce

### Actual result

<table style="width:100%">
  <tr>
  </tr>
</table> 
</details> 

### Expected result

### Your Environment
Android/iOS version: 26.5
Device model: iPhone 15

aleksandr-tata:
Fix: resolve the detailed object in the background. The sheet is now presented
immediately from the tapped rendered object (name/type/icon already known) and
is enriched via `rebuildRows` once the lookup completes — matching Android,
which loads this data off the main thread.

Result on the test device: perceived open ~1.6s → ~0.86s, no main-thread freeze.
Note: `searchDetailedObject` now runs off the main thread.


Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -204,7 +226,24 @@ final class PlaceDetailsViewController: OAPOIViewController {
         setup(amenity)
         updateTargetPoint(with: amenity)
     }
-    
+
+    private func resolveDetailedObjectInBackground() {
+        guard let renderedObject else { return }
+        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
```
tigrim, line 236: This can introduce a race condition. An outdated background request may complete later and overwrite the current state. Consider validating that the result is still relevant before applying it

Sources/Controllers/TargetMenu/OATargetInfoViewController.mm
```diff
@@ -490,36 +490,51 @@ - (void)buildNearestWikiRow:(NSMutableArray<OAAmenityInfoRow *> *)rows
...
     }
+
+    _isFetchingNearestWiki = YES;
+    __weak __typeof(self) weakSelf = self;
+    [self fetchNearestPoi:poi filter:wikiFilter completion:^(NSArray<OAPOI *> *results) {
+        __strong __typeof(weakSelf) strongSelf = weakSelf;
+        if (!strongSelf)
+            return;
+
+        strongSelf->_nearestWiki = [results copy];
```
tigrim, line 542: This changes the nearest wiki lookup path from `processNearestWiki` to `fetchNearestPoi`, but the old wiki language filtering seems to be lost!

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -204,7 +226,24 @@ final class PlaceDetailsViewController: OAPOIViewController {
         setup(amenity)
         updateTargetPoint(with: amenity)
     }
-    
+
+    private func resolveDetailedObjectInBackground() {
```
tigrim, line 234: Async detail lookup can update the wrong target point PlaceDetailsViewController.resolveDetailedObjectInBackground / updateTargetPoint(with:) The background lookup resolves later, then blindly calls mapPanel.getCurrentTargetPoint() and mutates it. If the user taps another map object before the lookup finishes, the old controller can update the new current target’s icon and refresh the map with stale data

Sources/Controllers/TargetMenu/OATargetMenuViewController.mm
```diff
@@ -195,15 +195,7 @@ + (OATargetMenuViewController *)createMenuController:(OATargetPoint *)targetPoin
...
-            BaseDetailsObject *detailsObject = [OAAmenitySearcher.sharedInstance searchDetailedObject:selectedObject];
-            if (detailsObject)
-            {
-                controller = [[PlaceDetailsViewController alloc] initWithPoi:targetPoint.targetObj detailsObject:detailsObject renderedObject:targetPoint.targetObj];
-            }
-            else
-            {
-                controller = [[RenderedObjectViewController alloc] initWithRenderedObject:targetPoint.targetObj];
-            }
+            controller = [[PlaceDetailsViewController alloc] initWithRenderedObject:targetPoint.targetObj];
```
tigrim: This is more than just moving `searchDetailedObject` off the main thread: it removes the old fallback to `RenderedObjectViewController` for rendered objects where no `BaseDetailsObject` is found.

Previously:
- details found -> `PlaceDetailsViewController`
- details not found -> `RenderedObjectViewController`

Now every rendered object opens `PlaceDetailsViewController(initWithRenderedObject:)`, and if the async lookup returns nil it remains there. Please confirm this new rendered-only state is fully equivalent to the old `RenderedObjectViewController` behavior, including name/type/icon/OSM URL, target title updates, and rows shown when details are unavailable.

Also, the async lookup result should only be applied if the current target still corresponds to the same rendered object; otherwise a delayed lookup can update the wrong target after the user taps another object.

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
aleksandr-tata: fixed

Sources/Controllers/TargetMenu/OATargetInfoViewController.mm
aleksandr-tata: fixed

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
aleksandr-tata: fixed

Sources/Controllers/TargetMenu/OATargetMenuViewController.mm
aleksandr-tata: fixed

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -204,11 +230,34 @@ final class PlaceDetailsViewController: OAPOIViewController {
...
+                tableView.reloadData()
+            }
+        }
+    }
+
     private func updateTargetPoint(with amenity: OAPOI) {
         guard let mapPanel = OARootViewController.instance()?.mapPanel,
               let targetPoint = mapPanel.getCurrentTargetPoint() else { return }
 
+        targetPoint.title = amenity.nameLocalized ?? amenity.name
```
tigrim, line 262: ?

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -55,16 +67,30 @@ final class PlaceDetailsViewController: OAPOIViewController {
...
+    }
+
     override func buildPhotosRow(_ rows: NSMutableArray) {
         super.buildPhotosRow(rows)
         buildGuidesRow()
     }
     
     override func buildDescription(_ rows: NSMutableArray) {
+        if detailsObject == nil {
+            super.buildDescription(rows)
```
tigrim, line 91: ?

Sources/Controllers/TargetMenu/OATargetMenuViewController.mm
tigrim: ? 
```
case OATargetPOI:
        {
            controller = [[OAPOIViewController alloc] initWithPOI:targetPoint.targetObj];
            if (selectedObject && [selectedObject isKindOfClass:BaseDetailsObject.class])
            {
                BaseDetailsObject *detailsObject = [OAAmenitySearcher.sharedInstance searchDetailedObject:selectedObject];
                if (detailsObject)
                {
                    controller = [[PlaceDetailsViewController alloc] initWithPoi:targetPoint.targetObj detailsObject:detailsObject renderedObject:targetPoint.targetObj];
                }
                else
                {
                    controller = [[RenderedObjectViewController alloc] initWithRenderedObject:targetPoint.targetObj];
                }
            }
  
            break;
        }
```

Sources/Controllers/TargetMenu/OATargetInfoViewController.mm
```diff
@@ -849,43 +929,6 @@ - (void) rebuildRows
...
-    NSMutableArray<NSString *> *languagesToShow = [[wikiPlugin getLanguagesToShow] mutableCopy];
-    if ([languagesToShow containsObject:@"en"])
-    {
-        NSInteger index = [languagesToShow indexOfObject:@"en"];
-        [languagesToShow replaceObjectAtIndex:index withObject:@""];
-    }
-
-    while (osmwiki.count < kNearbyPoiMaxCount && radius <= kNearbyPoiMaxRadius)
-    {
-        osmwiki = [[OAAmenitySearcher findPOIsByTagName:nil name:nil location:locI categoryName:OSM_WIKI_CATEGORY poiTypeName:nil radius:radius] mutableCopy];
```
tigrim, line 867: Wiki targets lose “Wikipedia around” results

Sources/Controllers/TargetMenu/POI/RenderedObjectAmenityProvider.swift
```diff
@@ -120,19 +147,26 @@ final class RenderedObjectAmenityProvider: NSObject {
...
-        
-        if cachedTypeStr?.isEmpty ?? true {
-            let additionalInfoKeys = amenity?.getAdditionalInfoKeys()
-            
-            cachedTypeStr = searchObjectNameByRawTags(
-                tags: renderedObject.tags as? [String: String],
-                additionalInfoKeys: additionalInfoKeys
-            )
-        }
+        let byType = resolvedTypeName(amenity)
```
tigrim: Type label logic no longer matches Android and can become less specific

tigrim:
```
typeStr = searchObjectTypeByAmenityTags(that.getAmenity());

if (Algorithms.isEmpty(typeStr)) {
    typeStr = searchObjectNameByIconRes();
}

if (Algorithms.isEmpty(typeStr)) {
    typeStr = searchObjectNameByRawTags(...);
}

return typeStr != null ? typeStr : super.getTypeStr();
```

Sources/Controllers/TargetMenu/POI/RenderedObjectAmenityProvider.swift
```diff
@@ -147,26 +147,15 @@ final class RenderedObjectAmenityProvider: NSObject {
             }
         }()
         
-        let byType = resolvedTypeName(amenity)
         let byTags = amenity.flatMap { searchObjectTypeByAmenityTags($0) }
```
tigrim: This computes all fallback candidates eagerly, even though type resolution should short-circuit. Please keep the Android-like fallback order and return as soon as a non-empty type string is found, so we avoid unnecessary tag scans/translations and preserve priority semantics

Sources/Controllers/TargetMenu/POI/RenderedObjectAmenityProvider.swift
```diff
@@ -102,6 +107,28 @@ final class RenderedObjectAmenityProvider: NSObject {
         return cachedNameStr ?? ""
     }
     
+    func actualContentFromIconRes() -> String? {
```
tigrim: private?

Sources/Controllers/TargetMenu/POI/RenderedObjectAmenityProvider.swift
```diff
@@ -102,6 +107,28 @@ final class RenderedObjectAmenityProvider: NSObject {
...
+        guard let content = renderedObject?.iconRes, !content.isEmpty else { return nil }
+        if content == "osmand_steps" {
+            return "highway_steps"
+        }
+        return content
+    }
+
+    func searchObjectNameByIconRes() -> String? {
```
tigrim: private?

Sources/Controllers/TargetMenu/POI/RenderedObjectAmenityProvider.swift
aleksandr-tata: fixed

Sources/Controllers/TargetMenu/POI/RenderedObjectAmenityProvider.swift
aleksandr-tata: fixed

PR #5540, opened 2026-07-20
Cloud sync checking fix

### LINKED ISSUE #5359: Cloud sync Checking process is significantly slower on iOS than on Android
## Description
When logging into **OsmAnd Cloud**, the **Checking** process takes significantly longer on **iOS** compared to **Android**, even when both devices are using the same previously synced account.

### How to reproduce
* Go to **OsmAnd Cloud** and select **Login**.
* Wait until the **Checking** process is complete.

### Actual result
<table style="width:100%">
  <tr>
    <th>iOS</th>
    <th>Android</th>
  </tr>
  <tr>
    <td><video src="https://github.com/user-attachments/assets/e58d7a65-0c40-4c3e-b52e-bd96a369af5d"  width="300" height="auto"/> </td>
    <td><video src="https://github.com/user-attachments/assets/3aa97135-c338-4511-a4f9-4ae8c13d3fa4"  width="300" height="auto"/> </td>
  </tr>
</table>   
Log file – [Apr 28, 2026 12:45.log](https://github.com/user-attachments/files/27161372/Apr.28.2026.12.45.log)

Both devices use the same account, which was previously fully synced.
* **Checking** takes approximately 35 seconds on **iOS**.
* **Checking** takes approximately 3 seconds on **Android**.

### Expected result
* **Checking** process on **iOS** should be optimized to perform at a comparable speed to **Android** (approximately 3 seconds).

**Environment**
OsmAnd Version: OsmAnd Maps 5368, released 23.04.2026
iOS version: 26.5

vitaliy-sova-ios:


tigrim (commented):
Please split this PR into two parts:
1. Functional Unicode-normalization fix for cloud sync comparison.
2. Temporary COLLECT_LOCAL_DIAG logging.

The diagnostics should live in a separate PR/commit so we can safely revert it after collecting logs without reverting the actual sync fix.

tigrim (commented):
Unicode fix is incomplete in OACollectLocalFilesTask: PR normalizes local/remote keys in OAGenerateBackupInfoTask, but uploaded DB lookup still checks only original fileName and precomposedStringWithCanonicalMapping. If DB has decomposed name and local filename is composed, uploadTime stays 0, so unchanged files can become false uploads/merges. Add decomposed fallback or normalize _infos keys the same way. See OACollectLocalFilesTask.m - line 195

PR #5542, opened 2026-07-20
Functional Unicode-normalization fix for cloud sync comparison

### LINKED ISSUE #5359: Cloud sync Checking process is significantly slower on iOS than on Android

vitaliy-sova-ios:


tigrim (commented):
There is still one non-normalized DB lookup:

getUploadedFileInfo:remoteFile.type name:remoteFile.name

This is the code path for the case where the remote file exists but the local file has already been deleted. If a file with a Unicode name was uploaded using one normalization form, but the remote API returns the name in a different normalization form, the DB record won't be found.

As a result, instead of adding the file to filesToDelete, the code will add it to filesToDownload, meaning a file that was intentionally deleted locally may be downloaded from the cloud again.

This lookup should either use a normalized map, as in OACollectLocalFilesTask, or implement a normalized fallback for the DB lookup.

tigrim (approved):

alex-dev-neo:
[OACollectLocalFilesTask.m](https://github.com/osmandapp/OsmAnd-iOS/blob/1e45a580b80bea34970808e3563842eccb4e85f4/Sources/Backup/OACollectLocalFilesTask.m#L50-L55) and [OAGenerateBackupInfoTask.m](https://github.com/osmandapp/OsmAnd-iOS/blob/1e45a580b80bea34970808e3563842eccb4e85f4/Sources/Backup/OAGenerateBackupInfoTask.m#L74-L80), entries are assigned to a normalized key without handling collisions. The database can contain both NFC and NFD rows because lookups and updates use exact strings. After normalization, dictionary enumeration order determines which row survives. Selecting a stale uploadTime/MD5 can produce the wrong sync decision—including downloading a locally deleted file instead of deleting its remote copy. When normalized keys collide, retain the entry with the greatest uploadTime

PR #5547, opened 2026-07-21
Astro explore improvements

vitaliy-sova-ios:


Sources/Plugins/Astronomy/search/StarMapSearchViewController.swift
```diff
@@ -939,16 +932,18 @@ final class StarMapSearchViewController: UIViewController {
...
         let filteredEntries = preparedCatalogEntries.filter { entry in
             if entry.objectCount <= 0 {
                 return false
             }
             if queryLower.isEmpty {
                 return true
             }
-            return entry.displayName.lowercased(with: Locale.current).contains(queryLower) ||
-                (entry.description ?? "").lowercased(with: Locale.current).contains(queryLower)
+            let displayNameNormilized = entry.displayName.folding(options: [.caseInsensitive, .diacriticInsensitive], locale: .current)
```
tigrim: typo

Sources/Plugins/Astronomy/search/StarMapSearchViewController.swift
```diff
@@ -939,16 +932,18 @@ final class StarMapSearchViewController: UIViewController {
...
+            let descriptionNormilized = (entry.description ?? "").folding(options: [.caseInsensitive, .diacriticInsensitive], locale: .current)
```
tigrim: typo

tigrim (commented):
The empty state in My Data is positioned using sortFilterChipsView.frame.height + Layout.contentPadding, but the filters are now hosted inside tableHeaderView. The header itself has a minimum height of 48 pt, while the embedded chips view may only be 36 pt tall or may not have its final frame yet when this calculation runs. As a result, the empty state card can end up too high, overlapping the header or leaving incorrect spacing. It would be more reliable to calculate the offset from searchRecycler.tableHeaderView?.frame.height ?? 0 instead

tigrim (approved):

PR #5561, opened 2026-07-24
Car play map mode setting

### LINKED ISSUE #5398: CarPlay Appearance Overrides OsmAnd Map Mode Setting
## Description

### How to reproduce

### Actual result

### Expected result

**Environment**
OsmAnd Version: 5380
iOS version: 26.5 Beta

vitaliy-sova-ios:


Sources/AppHost/CarPlayDelegate/CarPlayService.swift
```diff
@@ -66,37 +62,41 @@ final class CarPlayService: NSObject {
...
+
+    func applyCarPlayMapAppearance() {
+        let mode = DayNightMode(rawValue: OAAppSettings.sharedManager().carPlayMapAppearanceMode.get()) ?? .appTheme
+        switch mode {
+        case .day, .night, .auto:
+            OADayNightHelper.instance().setCarPlayMode(Int(mode.rawValue))
+        case .appTheme:
+            if let lastContentStyle {
+                applyVehicleAppearance(with: lastContentStyle)
+            } else if let style = sessionConfiguration?.contentStyle {
```
tigrim: This may cause flickering. When the scene connects, `contentStyle` may still have an incorrect value, so the appearance is first applied here and then updated again when this callback is invoked:

```swift
func sessionConfiguration(
    _ sessionConfiguration: CPSessionConfiguration,
    contentStyleChanged contentStyle: CPContentStyle
)
```

Sources/AppHost/CarPlayDelegate/CarPlayService.swift
```diff
@@ -66,37 +62,41 @@ final class CarPlayService: NSObject {
...
+        } else {
+            NSLog("[CarPlayService] vehicle appearance: unknown contentStyle, keep previous")
         }
     }
     
     private func initSessionConfiguration() {
         guard sessionConfiguration == nil else { return }
         sessionConfiguration = CPSessionConfiguration(delegate: self)
+        if let style = sessionConfiguration?.contentStyle {
+            lastContentStyle = style
```
tigrim, line 104: up

Sources/CarPlay/CarPlayMapModeListController.swift
```diff
@@ -0,0 +1,85 @@
...
+//  OsmAnd Maps
+//
+//  Created by Vitaliy Sova on 23.07.2026.
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import CarPlay
+
+@objcMembers
+final class CarPlayMapModeListController: NSObject {
```
tigrim: :OABaseCarPlayInterfaceController

Sources/CarPlay/CarPlayMapModeListController.swift
```diff
@@ -0,0 +1,85 @@
...
+            return localizedString("carplay_map_mode_vehicle_appearance")
+        case .day, .night, .auto:
+            return mode.title
+        }
+    }
+
+    func present() {
+        let template = CPListTemplate(title: localizedString("map_mode"), sections: [makeSection()])
+        
+        interfaceController.pushTemplate(template, animated: true) { completed, error in
```
tigrim: use safePushTemplate

Sources/CarPlay/CarPlayMapModeListController.swift
```diff
@@ -0,0 +1,85 @@
...
+    private let interfaceController: CPInterfaceController
+    private let onModeChanged: (() -> Void)?
+
+    init(interfaceController: CPInterfaceController, onModeChanged: (() -> Void)? = nil) {
+        self.interfaceController = interfaceController
+        self.onModeChanged = onModeChanged
+        super.init()
+    }
+    
+    static func currentTitle() -> String {
```
tigrim: This shouldn't be static

tigrim:
conflicts

Sources/AppHost/CarPlayDelegate/CarPlayService.swift
vitaliy-sova-ios: We need to read contentStyle when creating CPSessionConfiguration so Vehicle appearance has an initial value.
sessionConfiguration(_:contentStyleChanged:) is only called when the style changes, not when the session is created

Sources/AppHost/CarPlayDelegate/CarPlayService.swift
```diff
@@ -44,9 +41,8 @@ final class CarPlayService: NSObject {
             return
         }
         sessionConfiguration = nil
-        restoreOriginalMapAppearanceModeIfNeeded()
-        appMapAppearanceMode = nil
-        carPlayMapAppearanceMode = nil
+        lastContentStyle = nil
+        OADayNightHelper.instance().resetCarPlayMode()
```
tigrim: resets OADayNightHelper before restoring the original app profile. resetCarPlayMode() immediately recalculates night mode using the still-active CarPlay profile, then restoreOnDisconnect() switches back to the phone profile without forcing another day/night recalculation. If CarPlay was forced Night/Day and the original profile differs, the main map/widgets can keep the wrong day/night state after disconnect until some later update. Restore the original profile first, or force OADayNightHelper again after restoreOnDisconnect

tigrim: We also need to consider that two CarPlay scenes can be connected at the same time. If one of them disconnects here, which theme will remain active for the other scene?

Sources/Helpers/OADayNightHelper.h
```diff
@@ -18,6 +18,8 @@
 - (void) forceUpdate;
 - (BOOL) setTempMode:(NSInteger)dayNightMode;
 - (BOOL) resetTempMode;
+- (void) setCarPlayMode:(NSInteger)dayNightMode;
```
tigrim: - (void)setCarPlayMode:(NSInteger)dayNightMode;
- (void)resetCarPlayMode;

Sources/AppHost/CarPlayDelegate/CarPlayService.swift
```diff
@@ -21,14 +21,16 @@ final class CarPlayService: NSObject {
     
     private var sessionConfiguration: CPSessionConfiguration?
     private var lastContentStyle: CPContentStyle?
+    private var activeScenes: Set<CarPlaySceneType> = []
```
tigrim: Why are you keeping this in CarPlayService? It doesn't belong there

Sources/AppHost/CarPlayDelegate/CarPlayService.swift
```diff
@@ -40,15 +42,18 @@ final class CarPlayService: NSObject {
...
             return
         }
-        sessionConfiguration = nil
-        lastContentStyle = nil
-        OADayNightHelper.instance().resetCarPlayMode()
-        OARoutingHelper.sharedInstance().onCarPlayConnectionStateChanged()
-        navigationModeProvider.restoreOnDisconnect()
+        activeScenes.remove(sceneType)
         if case .app = sceneType, isSearchUICorePrepared, UIApplication.shared.mainScene != nil {
             OAQuickSearchHelper.instance().setResourcesForSearchUICore()
```
tigrim: Why has `setResourcesForSearchUICore` been moved above `restoreOnDisconnect`? Move it back

tigrim (approved):

PR #5567, opened 2026-07-27
location position setting ignored fixed

### LINKED ISSUE #5506: Location Position on Screen setting is ignored after route calculation (Pedestrian profile)
Related task https://github.com/osmandapp/OsmAnd-iOS/issues/5088

### How to Reproduce
* Switch to the **Pedestrian** profile
* Open **Menu – Navigation – Set Destination**
* Calculate the route, then collapse the **Navigation** window

### Actual Result
<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td><video src="https://github.com/user-attachments/assets/f073f344-c06f-4131-ad74-134d27b2cf9c"  width="300" height="auto"/> </td>
    <td> </td>
    <td> </td>
  </tr>
</table>   

* When the route is calculated, **Location Position on Screen** appears to be ignored

### Expected Result
* **Location Position on Screen** should always be respected

**Environment**
OsmAnd Version: 5421
iOS Version: 27 Developer Beta
Device Model: iPhone 16 Pro Max

Vladyslav-lys:


Sources/Controllers/Map/OAMapViewTrackingUtilities.mm
```diff
@@ -1054,13 +1054,9 @@ - (void) updateSettings
         
         EOAPositionPlacement placement = (EOAPositionPlacement) [_settings.positionPlacementOnMap get];
         if (placement == EOAPositionPlacementAuto)
-        {
-            _mapViewController.mapPosition = ([_settings.rotateMap get] == ROTATE_MAP_BEARING && !_routePlanningMode ? BOTTOM_CONSTANT : CENTER_CONSTANT);
-        }
+            _mapViewController.mapPosition = ([_settings.rotateMap get] == ROTATE_MAP_BEARING ? BOTTOM_CONSTANT : CENTER_CONSTANT);
```
tigrim, line 1057: The PR only fixes the regular map view, but the same logic is still present in CarPlay

tigrim (approved):

PR #5568, opened 2026-07-27
Fixed context menu performance

### LINKED ISSUE #5337: Context menu opens twice as slowly on iOS compared to Android

aleksandr-tata:
Fixed Popular places freezes

### DIFF
--- a/Sources/Controllers/Map/Helpers/OAMapSelectionHelper.mm
+++ b/Sources/Controllers/Map/Helpers/OAMapSelectionHelper.mm
@@ -180,16 +180,16 @@ static int TILE_SIZE = 256;
                         if (name)
                             [names addObject:name];
                     }
-                    
+
                     NSString *nativeName = cppAmenity->nativeName.toNSString();
                     if (nativeName)
                         [names addObject:nativeName];
-                    
+
                     OAPOI *requestAmenity = [[OAPOI alloc] init];
                     requestAmenity.obfId = cppAmenity->id.id;
                     [requestAmenity setLatitude:result.objectLatLon.coordinate.latitude];
                     [requestAmenity setLongitude:result.objectLatLon.coordinate.longitude];
-                    
+
                     OAAmenitySearcherRequest *request = [[OAAmenitySearcherRequest alloc] initWithMapObject:requestAmenity names:[names copy]];
                     detailsObject = [amenitySearcher searchDetailedObjectWithRequest:request];
                 }
--- a/Sources/Controllers/Map/Helpers/OAMapSelectionHelper.mm
+++ b/Sources/Controllers/Map/Helpers/OAMapSelectionHelper.mm
@@ -166,33 +166,28 @@ static int TILE_SIZE = 256;
             
             if (cppAmenity != nullptr)
             {
-                OAPOI *parsedPoi = [OAAmenitySearcher parsePOIByAmenity:cppAmenity];
-                if (parsedPoi)
+                if ([self isAmenityAlreadyCollected:cppAmenity->id.id selectedObjects:result.allObjects])
+                    continue;
+
+                NSMutableArray<NSString *> *names = [NSMutableArray new];
+                for (const auto& entry : OsmAnd::rangeOf(OsmAnd::constOf(cppAmenity->localizedNames)))
                 {
-                    [result collect:parsedPoi provider:_provider];
+                    NSString *name = entry.value().toNSString();
+                    if (name)
+                        [names addObject:name];
                 }
-                else
-                {
-                    NSMutableArray<NSString *> *names = [NSMutableArray new];
-                    for (const auto& entry : OsmAnd::rangeOf(OsmAnd::constOf(cppAmenity->localizedNames)))
-                    {
-                        NSString *name = entry.value().toNSString();
-                        if (name)
-                            [names addObject:name];
-                    }
 
-                    NSString *nativeName = cppAmenity->nativeName.toNSString();
-                    if (nativeName)
-                        [names addObject:nativeName];
+                NSString *nativeName = cppAmenity->nativeName.toNSString();
+                if (nativeName)
+                    [names addObject:nativeName];
 
-                    OAPOI *requestAmenity = [[OAPOI alloc] init];
-                    requestAmenity.obfId = cppAmenity->id.id;
-                    [requestAmenity setLatitude:result.objectLatLon.coordinate.latitude];
-                    [requestAmenity setLongitude:result.objectLatLon.coordinate.longitude];
+                OAPOI *requestAmenity = [[OAPOI alloc] init];
+                requestAmenity.obfId = cppAmenity->id.id;
+                [requestAmenity setLatitude:result.objectLatLon.coordinate.latitude];
+                [requestAmenity setLongitude:result.objectLatLon.coordinate.longitude];
 
-                    OAAmenitySearcherRequest *request = [[OAAmenitySearcherRequest alloc] initWithMapObject:requestAmenity names:[names copy]];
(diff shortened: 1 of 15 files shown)


Sources/Controllers/Map/Layers/OAContextMenuLayer.mm
```diff
@@ -421,53 +422,48 @@ - (BOOL) isSecondaryProvider
...
     OATargetPoint *targetPoint = [[OATargetPoint alloc] init];
     targetPoint.location = CLLocationCoordinate2DMake(latitude, longitude);
     targetPoint.title = formattedTargetName;
     targetPoint.icon = [poiType icon];
-    targetPoint.titleAddress = roadTitle;
+    targetPoint.titleAddress = nil;
     targetPoint.type = OATargetPOI;
     targetPoint.targetObj = poi;
 
+    OATargetPoint *tp = targetPoint;
```
tigrim: Why add a custom async lookup here? The existing `shouldFetchAddress` flow should cover this case and already handles stale target updates

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -231,17 +244,43 @@ final class PlaceDetailsViewController: OAPOIViewController {
         updateTargetPoint(with: amenity)
     }
 
+    private func highlightPolygonInBackground() {
+        let seed: OAPOI? = (sourceObject as? OAPOI) ?? detailsObject?.syntheticAmenity
```
tigrim: poi

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -231,17 +244,43 @@ final class PlaceDetailsViewController: OAPOIViewController {
...
+        guard let seed else { return }
+        DispatchQueue.global(qos: .userInitiated).async { [weak self] in
+            guard let geoObject = OAAmenitySearcher.sharedInstance().resolveGeometryOnly(seed) else { return }
+            DispatchQueue.main.async {
+                guard let mapPanel = OARootViewController.instance()?.mapPanel,
```
tigrim: `highlightPolygonInBackground()` only checks that some current target point exists. If the user opens another object before this async geometry lookup finishes, this callback can highlight the old polygon over the new target.

tigrim: Is `PlaceDetailsViewController` the right place to drive polygon highlighting? This makes the details screen directly mutate map-layer state and also requires stale-target checks in async callbacks.

On Android, polygon highlighting is tied to the context menu selected object in `ContextMenuLayer`: the layer owns `selectedObject` and draws/clears the outline from that state. Could we keep the same model here and update the current selected/target object when details or geometry are resolved, letting the map panel/context menu layer handle highlighting for the current item?

Sources/Controllers/TargetMenu/OATargetMenuViewController.mm
```diff
@@ -176,11 +176,14 @@ + (OATargetMenuViewController *)createMenuController:(OATargetPoint *)targetPoin
...
         case OATargetPOI:
         {
-            controller = [[OAPOIViewController alloc] initWithPOI:targetPoint.targetObj];
             if (selectedObject && [selectedObject isKindOfClass:BaseDetailsObject.class])
             {
                 controller = [[PlaceDetailsViewController alloc] initWithPoi:targetPoint.targetObj detailsObject:selectedObject renderedObject:targetPoint.targetObj];
             }
+            else
+            {
+                controller = [[PlaceDetailsViewController alloc] initWithAmenityPoi:targetPoint.targetObj];
```
tigrim, line 178: controller = [[OAPOIViewController alloc] initWithPOI:targetPoint.targetObj]; ?

Sources/Controllers/TargetMenu/OATargetMenuViewController.mm
```diff
@@ -234,44 +237,7 @@ + (OATargetMenuViewController *)createMenuController:(OATargetPoint *)targetPoin
...
-                            break;
-                        }
-                    }
-                    if (!content)
-                        content = targetPoint.localizedContent.allValues.firstObject;
-
-                    controller = [[OAWikiMenuViewController alloc] initWithPOI:targetPoint.targetObj content:content];
-                }
-            }
+            controller = [[PlaceDetailsViewController alloc] initWithAmenityPoi:targetPoint.targetObj];
```
tigrim: ?

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -31,9 +34,17 @@ final class PlaceDetailsViewController: OAPOIViewController {
         let poi = BaseDetailsObject.convertRenderedObjectToAmenity(renderedObject)
         super.init(poi: poi)
         self.renderedObject = renderedObject
+        self.sourceObject = renderedObject
```
tigrim: ?

Sources/Controllers/Map/Helpers/OAMapSelectionHelper.mm
```diff
@@ -166,25 +166,33 @@ - (void)selectObjectsFromOpenGl:(MapSelectionResult *)result point:(CGPoint)poin
             
             if (cppAmenity != nullptr)
             {
-                NSMutableArray<NSString *> *names = [NSMutableArray new];
-                for (const auto& entry : OsmAnd::rangeOf(OsmAnd::constOf(cppAmenity->localizedNames)))
+                OAPOI *parsedPoi = [OAAmenitySearcher parsePOIByAmenity:cppAmenity];
```
tigrim: We previously agreed that this code wasn't needed. Android doesn't use anything like this, so why is it needed here?

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -57,6 +68,10 @@ final class PlaceDetailsViewController: OAPOIViewController {
     }
     
     override func getNameStr() -> String? {
+        if preservesAmenityPresentation {
```
tigrim: ?

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -259,8 +277,12 @@ final class PlaceDetailsViewController: OAPOIViewController {
         guard let mapPanel = OARootViewController.instance()?.mapPanel,
               let targetPoint = mapPanel.getCurrentTargetPoint() else { return }
 
-        targetPoint.title = amenity.nameLocalized ?? amenity.name
-        targetPoint.icon = amenity.type?.icon()
+        if let title = amenity.nameLocalized ?? amenity.name, !title.isEmpty {
```
tigrim, line 290: After the background searchDetailedObject(...) completes, only targetPoint.title and the icon are updated. targetPoint.targetObj itself remains the original, incomplete OAPOI, and OAMapPanelViewController.setSelectedObject(...) is never called again.

As a result, the table shows the updated details, while the map highlight/polygon and any actions that read targetPoint.targetObj, values, localizedContent, or obfId continue using the stale object.

On Android, MapContextMenu.init(...) synchronously replaces the selected object with BaseDetailsObject, so the menu state and map state never diverge.

It would be better to atomically replace the current target object/details after the async resolve completes and reapply the selected object/highlight, using the same stale-target guard

PR #5570, opened 2026-07-27
Optimize Wiki article opening

### LINKED ISSUE #5500: Long delay when opening Wikipedia articles from the context menu
## Description
There is a significant delay when attempting to open and read full Wikipedia articles within the app after downloading the Wikipedia data.

### How to reproduce
* Select **Berlin** on the map.
* In the context menu, tap on **Details**, then select **Read full article**, and choose **Download Wikipedia data**.
* Tap on **Read full article** again.

### Actual result

<table style="width:100%">
  <tr>
    <th>iPhone XS Max</th>
    <th>iPhone 16 Pro Max</th>
    <th>Android</th>
  </tr>
  <tr>
    <td><video src="https://github.com/user-attachments/assets/4a12075a-be7b-4cbe-9ff7-aca2487bc240" width="300" height="auto"/> </td>
    <td><video src="https://github.com/user-attachments/assets/44c4c631-a7a8-40be-9dc2-76d573dca682" width="300" height="auto"/> </td>
    <td><video src="https://github.com/user-attachments/assets/2ed6785a-b1f5-48c5-87de-2ea5e957260f" width="300" height="auto"/> </td>
  </tr>
</table>   

Opening the Wikipedia article takes a significant amount of time.
* iPhone XS Max with iOS 18 ~ 4 sec
* iPhone 16 Pro Max with iOS 27 ~ 14 sec
* * Log file – [Jul 01, 2026 10:09.log](https://github.com/user-attachments/files/29538772/Jul.01.2026.10.09.log)

### Expected result
The Wikipedia article should open quickly and smoothly without any noticeable delay.

**Environment**
OsmAnd Version: 
iOS version: iOS 18 & iOS 27 Developer beta
Device model: iPhone XS Max & iPhone 16 Pro Max

vitaliy-sova-ios:


Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
```diff
@@ -713,14 +713,16 @@ - (void)updateWikiData
...
-    [super loadWebView];
-  }
+    if (_externalURL)
+    {
+        [self.webView loadRequest:[NSURLRequest requestWithURL:_externalURL]];
+        self.webView.hidden = NO;
+    } else
+    {
+        self.webView.hidden = NO;
+        [self.webView loadHTMLString:_content baseURL:[self getUrl]];
```
tigrim: ?

Sources/Helpers/OAWikiArticleHelper.mm
```diff
@@ -539,4 +568,20 @@ + (nullable UIMenu *)createLanguagesMenu:(nullable NSArray<NSString *> *)availab
     return languageMenu;
 }
 
++ (BOOL)isWikipediaDownloadedAt:(CLLocation *)location
```
tigrim: is `isWikipediaDownloadedAt:` intentionally added as a new helper? We already have very similar logic in `OACollapsableNearestPoiWikiView` (`findWikiRegion` -> `findResourceItem` -> `isResourceInstalled`). Could we reuse/extract that path to avoid having two slightly different definitions of “Wikipedia is downloaded at this location”?

Also, this does not seem to be a direct Android parity method: Android checks either an existing wiki repository (`findWikiRepository`) or download items via `DownloadResources.findIndexItemsAt/findSmallestIndexItemAt`. So it would be good to confirm this iOS check behaves correctly for nested/overlapping wiki regions.

Sources/Controllers/TargetMenu/OATargetInfoViewController.mm
```diff
@@ -1874,10 +1885,24 @@ - (void) tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
             OAIAPHelper *helper = [OAIAPHelper sharedInstance];
             if ([helper.wiki isPurchased])
             {
-                [OAWikiArticleHelper showWikiArticle:[[CLLocation alloc] initWithLatitude:self.location.latitude
-                                                                                longitude:self.location.longitude]
-                                                 url:info.text
-                                          sourceView:[tableView cellForRowAtIndexPath:indexPath]];
+                UIView *hostView = OARootViewController.instance.view;
```
tigrim: The same `MBProgressHUD` initialization block is duplicated at both wiki-opening call sites in `OATargetInfoViewController`. Please extract it to a private helper, for example `createWikiArticleSearchHUD`

Sources/Helpers/OAAmenitySearcher.mm
```diff
@@ -1098,6 +1098,66 @@ + (OAPOI *)findPOIByName:(NSString *)name lat:(double)lat lon:(double)lon
     return nil;
 }
 
++ (NSArray<OAPOI *> *)searchAmenitiesByName:(NSString *)name
```
tigrim, line 1101: Potential regression/risk: the old iOS path searched wiki POIs inside the selected region bbox, while the new `searchAmenitiesByName` searches by title across the whole local wiki resource and then opens `found[0]`.

If there are multiple wiki POIs with the same title/name in the same wiki resource, this can open an article unrelated to the tapped location. Android seems to use a similar repository-wide lookup, but on iOS this still changes the previous spatial behavior.

Can we keep the location as a ranking/filtering hint here, e.g. pass `xy31`/bbox into the name-search criteria or sort matched results by distance to the tapped location before choosing the first result?

Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
```diff
@@ -515,36 +525,57 @@ - (void)setDownloadImagesOnlyNow:(BOOL)onlyNow
...
+    
+    NSInteger requestId = ++_headerImageRequestId;
+
+    [self fetchHeaderImageUrl:^(NSString *headerImageUrl) {
+        if (!headerImageUrl.length)
+            return;
+        [self->_imageCacheHelper fetchSingleImageByURL:headerImageUrl
+                                             customKey:dbKey
+                                          downloadMode:[self getImagesDownloadMode]
+                                               onlyNow:[self isDownloadImagesOnlyNow]
```
tigrim: Download only now is reset by the base loadWebView callback before the async header-image download starts, so fetchSingleImageByURL may receive onlyNow:NO and skip the requested one-time download. Please capture BOOL onlyNow = [self isDownloadImagesOnlyNow] before loadWebView(_content) and pass that captured value into the async fetch

Sources/Controllers/TargetMenu/OATargetInfoViewController.mm
```diff
@@ -1874,10 +1892,19 @@ - (void) tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
...
-                [OAWikiArticleHelper showWikiArticle:[[CLLocation alloc] initWithLatitude:self.location.latitude
-                                                                                longitude:self.location.longitude]
-                                                 url:info.text
-                                          sourceView:[tableView cellForRowAtIndexPath:indexPath]];
+                MBProgressHUD *progressHUD = [self createWikiArticleSearchHUD];
+                
+                CLLocation *location = [[CLLocation alloc] initWithLatitude:self.location.latitude
+                                                                  longitude:self.location.longitude];
+                
+                [OAWikiArticleHelper showWikiArticle:@[[[CLLocation alloc] initWithLatitude:self.location.latitude
```
tigrim: `location` is already created above. Reuse it here as `@[location]` instead of allocating the same `CLLocation` again

Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
```diff
@@ -515,36 +525,57 @@ - (void)setDownloadImagesOnlyNow:(BOOL)onlyNow
...
     
-    NSString *cachedHeaderImage = [_imageCacheHelper readImageByDbKey:[self getHeaderImageCacheDbKey]];
-    if (cachedHeaderImage)
+    if ([self isImageTagAppended])
+    {
+        loadWebView(_content);
+        return;
+    }
+
+    NSString *dbKey = [self getHeaderImageCacheDbKey];
```
tigrim: don't use get prefix

Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
```diff
@@ -515,36 +525,57 @@ - (void)setDownloadImagesOnlyNow:(BOOL)onlyNow
...
+    NSString *cached = [_imageCacheHelper readImageByDbKey:dbKey];
+    if (cached.length > 0)
     {
         NSString *html = [self appendHeaderImageTag];
-        [self injectCachedImagesToHtmlAndReload:html loadWebView:loadWebView];
+        NSString *src = [OAImageToStringConverter getHtmlImgSrcTagContent:cached];
```
tigrim: up

Sources/Controllers/TargetMenu/OACollapsableNearestPoiWikiView.mm
```diff
@@ -116,7 +116,7 @@ - (void) buildViews
             regionName = OALocalizedString(@"download_wiki_region_placeholder");
         }
         
-        if (_resourceItem && app.resourcesManager->isResourceInstalled(_resourceItem.resourceId))
+        if ([OAResourcesUIHelper isIndexItemDownloadedAt:CLLocationCoordinate2DMake(_latitude, _longitude) type:OsmAndResourceType::WikiMapRegion])
```
tigrim, line 119: This changes the condition from checking the specific `_resourceItem` to checking any downloaded WikiMapRegion at the coordinate, including parent regions. This matches the Android-style logic and is probably intended, but please confirm this banner should also be hidden when a parent wiki region is downloaded?

Sources/Helpers/OAWebImagesCacheHelper.h
```diff
@@ -19,6 +19,8 @@
 
 - (void) fetchSingleImageByURL:(NSString *)url customKey:(NSString *)customKey downloadMode:(OADownloadMode *)downloadMode onlyNow:(BOOL)onlyNow onComplete:(void (^)(NSString *imageData))onComplete;
 
+- (NSArray<NSString *> *) extractImagesLinksFromHtml:(NSString *)html;
```
tigrim: remove space

tigrim (commented):
- [x] The header placeholder is appended before we know that the article has a thumbnail. If `fetchHeaderImageUrl` returns nil, or image download fails/returns empty, `injectHeaderImageBase64` exits and the `wiki-header-shimmer` remains in the article forever. Can we remove/hide the placeholder on failure?
- [x] Potential issue: body image injection is kicked off from `didCommitNavigation`, but the HTML body may not be fully parsed yet. Cached images can complete almost immediately, so `querySelectorAll('img')` may run before all article images exist in the DOM and those images will never be replaced. Could this move to `didFinishNavigation` or a DOM-ready JS hook?

Sources/Controllers/TargetMenu/POI/OAPOIViewController.mm
```diff
@@ -422,6 +425,19 @@ - (BOOL)buildShortWikiDescription:(NSDictionary<NSString *, id> *)filteredInfo a
             
             if ([pairDescription.first isKindOfClass:NSString.class])
             {
+                NSArray<NSString *> *langs = [self.poi getNames:WIKI_LANG defTag:@"en"];
+                NSString *lang = langs.firstObject;
+                if (NSStringIsEmpty(lang))
+                    lang = @"en";
+
```
tigrim, line 430: This builds the online Wikipedia URL with the domain language from `wiki_lang`, but the title from `[self.poi getName:locale]`. For multilingual POIs these can differ, producing links like `en.wikipedia.org/wiki/<non-English title>`. The title is also not percent-encoded before being passed to `SFSafariViewController`. Could we use the actual `wikipedia` tag when available, or build and encode the title for the same language as the domain?

Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
```diff
@@ -515,42 +528,79 @@ - (void)setDownloadImagesOnlyNow:(BOOL)onlyNow
...
-                }];
-                
-            }];
-        }
-        else
-        {
-            loadWebView(_content);
-            [self printHtmlToDebugFileIfEnabled:_content];
-        }
+        _content = [self appendHeaderImageTag];
```
tigrim, line 601: We add the 170px header placeholder before knowing whether Wikipedia returns a page image. If `fetchHeaderImageUrl` later returns nil, the placeholder is removed from the DOM, but `getCustomTitleHeaderTopOffset` still returns `kHeaderImageHeight` based only on the download mode. That can leave the page with an incorrect large-title/header offset for articles without a header image. Could we track actual header image presence and reset the offset when the placeholder is removed?

Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
```diff
@@ -515,42 +528,79 @@ - (void)setDownloadImagesOnlyNow:(BOOL)onlyNow
...
+        loadWebView(_content);
+        [self printHtmlToDebugFileIfEnabled:_content];
+        [self markBodyImagesInjectWithOnlyNow:onlyNow];
     }
+    
+    NSInteger requestId = ++_headerImageRequestId;
+
+    [self fetchHeaderImageUrl:^(NSString *headerImageUrl) {
```
tigrim, line 610: weak?

Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
```diff
@@ -631,30 +712,157 @@ - (NSString *)appendHeaderImageTag
...
+        _hasHeaderImage = YES;
         return _content;
     }
     else
     {
-        return [_content stringByReplacingOccurrencesOfString:@"</head>" withString:[NSString stringWithFormat:@"<img src=\"%@\" style=\"object-fit:cover; object-position:center; height:%dpx;\"></head>", [self getHeaderImageCacheDbKey], kHeaderImageHeight]];
+        _hasHeaderImage = YES;
+        
+        BOOL isLight = [ThemeManager shared].isLightTheme;
+        NSString *color1 = isLight ? @"#E8E8E8" : @"#222526";
```
tigrim: Why aren't these colors coming from the asset catalog?

Sources/POI/ExplorePlacesOnlineProvider.swift
```diff
@@ -246,6 +246,13 @@ final class ExplorePlacesOnlineProvider: ExplorePlacesProvider {
         if let id {
             amenity.setAdditionalInfo("wikidata", value: "Q\(id)")
         }
+
+        let lang = (properties.lang?.isEmpty == false ? properties.lang : nil)
```
tigrim: It's hard to read. Could you make it similar to the Android implementation?
```
	String lang = properties.getLang();
		if (Algorithms.isEmpty(lang)) {
			lang = properties.getWikiLang();
		}
		if (lang != null && !lang.isEmpty()) {
			amenity.setAdditionalInfo("wiki_lang:" + lang, "yes");
		}
```

Sources/Controllers/TargetMenu/Wiki/OAWikiWebViewController.mm
vitaliy-sova-ios: Shimmer has been added to the article's style

tigrim (approved):

PR #5574, opened 2026-07-29
Support routing params in route deep links

vitaliy-sova-ios:
Add support for a params query in route deep links: RouteDeepLinkBuilder now appends "/navigate", includes serialized routing params (percent-encoded) and uses appMode.stringKey. DeepLinkParser reads the "params" query and applies it via new OARoutingHelperUtils APIs. Implement routingParamsQueryValueForAppMode: and applyRoutingParamsQueryValue:forAppMode: in OARoutingHelperUtils.mm to serialize/deserialize per-profile routing parameters using OAAppSettings and the router. Import the utils in the Swift bridging header. Also adjust one XIB bottom constraint.

### DIFF
--- a/Sources/Helpers/DeepLinkManager/DeepLinkParser.swift
+++ b/Sources/Helpers/DeepLinkManager/DeepLinkParser.swift
@@ -393,7 +393,7 @@ final class DeepLinkParser: NSObject {
         var intermediatePointsParam: String?
         var endLatLonParam: String?
         var appModeKeyParam: String?
-        var routingParamsParam: String?
+        var routeParamsQueryValue: String?
         for item in queryItems {
             switch item.name.lowercased() {
             case "end":
@@ -406,7 +406,7 @@ final class DeepLinkParser: NSObject {
             case "via":
                 intermediatePointsParam = item.value
             case "params":
-                routingParamsParam = item.value
+                routeParamsQueryValue = item.value
             default:
                 break
             }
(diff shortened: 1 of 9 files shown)


Sources/Helpers/DeepLinkManager/DeepLinkParser.swift
```diff
@@ -404,6 +405,8 @@ final class DeepLinkParser: NSObject {
                 appModeKeyParam = item.value
             case "via":
                 intermediatePointsParam = item.value
+            case "params":
+                routingParamsParam = item.value
```
tigrim: Image

tigrim: routeParamsQueryValue

Sources/Router/OARoutingHelperUtils.mm
```diff
@@ -198,4 +199,131 @@ + (void) updateDrivingRegionIfNeeded:(CLLocation *)newStartLocation force:(BOOL)
...
+        return;
+    
+    auto router = [OsmAndApp.instance getRouter:mode];
+    if (!router)
+        return;
+    
+    OAAppSettings *settings = OAAppSettings.sharedManager;
+    auto parameters = [self getParametersForDerivedProfile:mode router:router];
+    
+    for (const auto &it : parameters)
```
tigrim, line 271: This makes opening a route deep link mutate the user's persistent routing profile: all routing parameters are reset to defaults first, then the link values are applied. That means a shared route can overwrite the user's existing avoid/vehicle/routing preferences even if they cancel the route

Sources/Router/OARoutingHelperUtils.h
```diff
@@ -22,14 +24,15 @@ struct RoutingParameter;
                             ref:(NSString *)ref
                     destination:(NSString *)destination
                         towards:(NSString *)towards;
-
+#ifdef __cplusplus
```
tigrim: ?

tigrim (approved):

PR #5584, opened 2026-07-31
polygon highlighting fixes

aleksandr-tata:
Moves polygon highlighting state to OAContextMenuLayer, following the Android implementation. Async geometry results are applied only when the original context-menu target is still active, preventing stale polygons from being drawn over a newly selected object. Polygon highlighting is also cleared when the context menu closes.

### DIFF
--- a/Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
+++ b/Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
@@ -14,6 +14,7 @@ final class PlaceDetailsViewController: OAPOIViewController {
     private var detailsObject: BaseDetailsObject?
     private var renderedObject: OARenderedObject?
     private var provider: RenderedObjectAmenityProvider!
+    private var cityFetchStarted = false
     
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
@@ -39,12 +40,15 @@ final class PlaceDetailsViewController: OAPOIViewController {
     }
 
     override func viewDidLoad() {
+        applyPolygonAddressIfNeeded()
         if detailsObject != nil {
             updateMenuWithDetailedObject()
         }
         super.viewDidLoad()
         if detailsObject == nil {
             resolveDetailedObjectInBackground()
+        } else {
+            resolveGeometryInBackground()
         }
     }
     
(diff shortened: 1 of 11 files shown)


Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -262,6 +279,75 @@ final class PlaceDetailsViewController: OAPOIViewController {
...
+
+    private func applyPolygonAddressIfNeeded() {
+        guard let mapPanel = OARootViewController.instance()?.mapPanel,
+              let targetPoint = mapPanel.getCurrentTargetPoint() else { return }
+
+        let amenity: OAPOI? = detailsObject?.syntheticAmenity ?? poi
+        let polygonPoints = max(Int(renderedObject?.x.count ?? 0), Int(amenity?.x.count ?? 0))
+        guard polygonPoints > 2 else { return }
+
+        targetPoint.shouldFetchAddress = false
```
tigrim, line 294: For polygon targets we disable the standard address fetch before we know an address was actually found. If both addr:* tags and reverse geocoder return empty, the old map_no_address fallback will no longer be shown. Can we either keep shouldFetchAddress enabled until success, or set the same fallback on empty result?

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -231,17 +235,30 @@ final class PlaceDetailsViewController: OAPOIViewController {
...
     }
 
+    private func resolveGeometryInBackground() {
+        guard let poi = detailsObject?.syntheticAmenity ?? self.poi,
+              let targetPoint = OARootViewController.instance()?.mapPanel?.getCurrentTargetPoint()
+        else { return }
+        DispatchQueue.global(qos: .userInitiated).async {
+            guard let geoObject = OAAmenitySearcher.sharedInstance().resolveGeometryOnly(poi) else { return }
+            DispatchQueue.main.async {
+                _ = OARootViewController.instance()?.mapPanel?.updateContextMenuSelectedObject(geoObject, for: targetPoint)
```
tigrim, line 245: This return value is currently ignored, so the call reads a bit awkwardly with `_ =`. Could we either make `updateContextMenuSelectedObject` return `Void`, or use the Bool to guard the follow-up work where it matters?

Sources/Controllers/TargetMenu/POI/PlaceDetailsViewController.swift
```diff
@@ -262,6 +279,75 @@ final class PlaceDetailsViewController: OAPOIViewController {
...
+                      let currentTarget = mapPanel.getCurrentTargetPoint(),
+                      currentTarget === targetPoint else { return }
+                currentTarget.titleAddress = trimmed
+                currentTarget.addressFound = true
+                self.delegate?.refreshTargetPointHeader?()
+            }
+        }
+    }
+
+    private func polygonAddressString(_ amenity: OAPOI?) -> String? {
```
tigrim, line 325: How is this related to optimizing context menu opening? This helper adds new address-building behavior from addr:* / is_in tags, and I don’t see the same logic on Android. If the goal is only to move polygon highlighting/geometry

tigrim:
add link the task

PR #5587, opened 2026-07-31
map not render after download fixed

### LINKED ISSUE #5496: Map data does not render automatically after download finishes unless manually zoomed in/out
### Description

When downloading a region's offline map while viewing that specific area, the map does not automatically refresh to display the newly available road networks, terrain, and POIs once the download completes. The map screen remains blank/empty (showing only the low-resolution world map) until the user manually forces a re-render by performing a **Zoom In** or **Zoom Out** gesture.

This issue is present on iOS/macOS (as seen in the Actual behavior video), whereas Android handles this correctly by immediately refreshing the viewport upon successful map extraction (Expected behavior).

### Steps to reproduce

1. Pan/zoom the map to an area where offline maps are **not yet downloaded** (e.g., Dundee/Perth region in Scotland).
2. Trigger the offline map download for this specific region.
3. In the download prompt/popup, tap **"Later"** or close it via the cross "X" button so you remain on the current map view.
4. Wait for the background download to complete fully.
5. Observe the map layout immediately after the download progress finishes.

### Actual result (iOS / macOS)
The map stays blank and unrendered. OsmAnd behaves as if the map data is still missing, even though the file is downloaded. The roads and details only appear *after* you change the zoom level manually.

<video src="https://github.com/user-attachments/assets/e98588c3-b6a1-4abe-b615-b71dec54e2fb"></video>
</details> 

### Expected result (Android)

- [x] As soon as the map download completes, the map view should automatically refresh/trigger a redraw, displaying all detailed vector data instantly without requiring any manual zoom actions from the user.

<video src="https://github.com/user-attachments/assets/78979175-b860-4ed4-a96e-ff43721b0f00"></video>
</details> 

### Environment

**OsmAnd Version:** OsmAnd Maps 5419, released 29.06.26

### Technical Note
The map downloader routine on iOS/macOS is likely missing a post-download execution hook to invalidate the current map tile cache (`mapView.refresh()` or similar rendering engine trigger) when the state shifts from "Downloading" to "Installed".

Vladyslav-lys:


Sources/Helpers/OAResourcesInstaller.mm
```diff
@@ -445,6 +445,14 @@ - (void) processResource:(id<OADownloadTask>)task
                 AstronomyPlugin *plugin = (AstronomyPlugin *)[OAPluginsHelper getPlugin:AstronomyPlugin.class];
                 [plugin clearCachedData];
             }
+
+            if (success)
+            {
+                const auto resource = _app.resourcesManager->getResource(resourceId);
```
tigrim: Why aren't you checking getLocalResources() instead?

tigrim (approved):

PR #5590, opened 2026-08-03
Carplay multi destination navigation history

### LINKED ISSUE #5391: Multi-destination routes are not saved correctly in navigation history on CarPlay and Android Auto
Last time I had a navigation with 2 destinations running via CarPlay.

OsmAnd crashed and in history I only see the search history. So no chance to "recover" the original navigation.
Additional CarPlay doesn't support adding multiple destinations, so I  need to stop and plan the navigation on my mobile.

Please implement a navigation history which also stores all destinations.

vitaliy-sova-ios:


Sources/CarPlay/OABaseCarPlayInterfaceController.mm
```diff
@@ -107,6 +107,17 @@ - (void) startNavigationGivenLocation:(CLLocation *)loc historyName:(nullable OA
     }
 }
 
+- (void)startNavigationFromPreviousRoute
```
tigrim, line 111: This restores only the target points, not necessarily the previous route. The following `enterRoutePlanningModeGivenGpx:nil from:nil...` clears the restored start point, and `setAppMode:CAR` may discard the route's original profile. So previous routes with a custom start or non-Car profile will be rebuilt differently.

Sources/CarPlay/OABaseCarPlayInterfaceController.mm
```diff
@@ -109,13 +110,22 @@ - (void) startNavigationGivenLocation:(CLLocation *)loc historyName:(nullable OA
 
 - (void)startNavigationFromPreviousRoute
 {
-    [OARoutingHelper.sharedInstance setAppMode:OAApplicationMode.CAR];
+    OAAppSettings *settings = [OAAppSettings sharedManager];
+    OAApplicationMode *mode = [settings.lastRouteApplicationMode get];
```
tigrim: Did you verify this path?! I don't see `lastRouteApplicationMode` being set anywhere in the codebase

tigrim (approved):

PR #5599, opened 2026-08-04
disabled zoom and updated center for add waypoint

### LINKED ISSUE #5459: Pin changes location when adding waypoint to track
### Description

When you chose a location  in a track and go to **Points > Add Waypoint,** the screen shows an overview of the whole track and end up changing the location of the pin, requiring the user to go back to where you previously dropped the pin

### Steps to reproduce

1 - Select a location in a track
2 - Tap **Points**
3 - Tap **Add Waypoint**

### Actual result

It reframes the screen, giving an overview of the whole track instead.
That makes the pin go to a completely different position of where I just selected.

If you cancel the action, pin goes back to where it was before (the place I wanted the waypoint to be)

https://github.com/user-attachments/assets/4c4fc4bf-2583-458b-923b-1b7e4ad9291b

### Expected result

My suggestion is to keep it zoomed in at the same location I just selected – most likely the user has the intention to add a waypoint in that very exact place instead of somewhere else on the track at the first place

```
OsmAnd Version: 5.4.0
iOS version: 26.5
Device model: iPhone 13 mini
```

Vladyslav-lys:


Sources/Controllers/TargetMenu/GPX/TrackMenu/OAAddWaypointViewController.mm
```diff
@@ -91,6 +91,13 @@ - (void)commonInit
         auto rect = _currentGpx.getRect;
         movedPoint.position = CLLocationCoordinate2DMake(rect.centerY, rect.centerX);
     }
+
+    if ([_targetMenuState isKindOfClass:OATrackMenuViewControllerState.class])
+    {
+        OATrackMenuViewControllerState *state = (OATrackMenuViewControllerState *) _targetMenuState;
+        if (state.openedFromMap)
+            movedPoint.position = state.pinLocation;
```
tigrim, line 99: state.pinLocation can be invalid even when openedFromMap == YES (for example via openTargetViewWithGPX(... openedFromMap:YES), which initializes it with kCLLocationCoordinate2DInvalid). Please guard this with CLLocationCoordinate2DIsValid(state.pinLocation) before using it as the new waypoint position.

tigrim (approved):

PR #5601, opened 2026-08-04
Use new separator color

### LINKED ISSUE #5520: Use new separator color for iOS 26
Apple changed the separator height starting with iOS 26 — it's now 1pt instead of the previous 0.5pt. That's why we previously used a more contrasted color for separators; now that the separator is twice as thick, we need a different color for iOS 26+ to keep the same visual weight.

- [x] Use `customSeparatorSolid` for iOS 26 / iPadOS 26 and later; 
- [x] Keep `customSeparator` for earlier versions.

DmitrySvetlichny:


tigrim (commented):
Could we encapsulate this in SeparatorAppearance and SeparatorView? Keeping color and thickness together avoids placing separator-specific behavior in UIColor and UIScreen. SeparatorView can provide its thickness through intrinsicContentSize, use the view’s current screen, and protect its vertical size inside UIStackView. This also removes the need for per-screen outlets and manually updated height constraints.

tigrim (commented):
@objcMembers
final class SeparatorAppearance: NSObject {

    /// The separator color defined by the app's design.
    static var color: UIColor {
        if #available(iOS 26.0, *) {
            return .customSeparatorSolid
        }
        return .customSeparator
    }

    /// Returns the separator thickness for the main screen.
    @objc(thickness)
    static func thickness() -> CGFloat {
        thickness(for: UIScreen.main)
    }

    /// Returns the separator thickness for the specified screen:
    /// one point on iOS 26 and later, and one physical pixel
    /// on earlier versions.
    @objc(thicknessForScreen:)
    static func thickness(for screen: UIScreen) -> CGFloat {
        if #available(iOS 26.0, *) {
            return 1.0
        }
        return 1.0 / screen.scale
    }

    /// Returns the separator thickness for the screen containing the view.
    @objc(thicknessForView:)
    static func thickness(for view: UIView) -> CGFloat {
        thickness(for: view.window?.screen ?? UIScreen.main)
    }

    private override init() {}
}

final class SeparatorView: UIView {

    override var intrinsicContentSize: CGSize {
        CGSize(width: UIView.noIntrinsicMetric,
               height: SeparatorAppearance.thickness(for: self)
        )
    }

    override init(frame: CGRect) {
        super.init(frame: frame)
        configureAppearance()
    }

    required init?(coder: NSCoder) {
        super.init(coder: coder)
        configureAppearance()
    }

    override func didMoveToWindow() {
        super.didMoveToWindow()
        invalidateIntrinsicContentSize()
    }

#if TARGET_INTERFACE_BUILDER
    override func prepareForInterfaceBuilder() {
        super.prepareForInterfaceBuilder()
        configureAppearance()
        invalidateIntrinsicContentSize()
    }
#endif

    private func configureAppearance() {
        backgroundColor = SeparatorAppearance.color
        setContentHuggingPriority(.required, for: .vertical)
        setContentCompressionResistancePriority(.required, for: .vertical)
    }
}

PR #5608, opened 2026-08-05
Keep free text search result markers

### LINKED ISSUE #5550: Keep free-text search result markers visible regardless of map zoom level
### Description

Hi,
when I search for a POI and click show on map I can see all the results, even if I zoom out.
But when I search for free text, I have to zoom in a lot which is hard to use.

### Steps to reproduce

Search for „Restaurant“ works as expected.
Search for „Mediamarkt“ or something else, you need to know where it is and zoom in to see it.

### Actual result

Don‘t See search result when zooming out.

### Expected result

I should see all results on the map, even when zooming out. 

```
OsmAnd Version: OsmAnd Maps 5428, veröffentlicht 17.07.26
iOS version: 26.x
Device model: iPhone 15
```

vitaliy-sova-ios:


vitaliy-sova-ios:
https://github.com/user-attachments/assets/603381ef-727f-4ec3-87f7-96d1cb6be289

tigrim (commented):
It looks like some unnecessary data structures were carried over from Android. I don't think we should be using them here. Please discuss this with @alex-dev-neo.

PR #5621, opened 2026-08-07
Restored old dialog

aleksandr-tata:
Restore old Attach to road dialog


Sources/Controllers/RoutePlanning/OAGpxApproximationViewController.mm
```diff
@@ -99,35 +87,19 @@ - (void)applyLocalization
...
 }
 
 - (void)onLeftButtonPressed
 {
-    _isApplying = NO;
-    [_approximationHelper cancelApproximation];
-    [self stopProgressUpdates];
     if (self.delegate)
-        [self.delegate onCancelSnapApproximation:NO];
+        [self.delegate onCancelSnapApproximation:YES];
```
tigrim: Cancel now always calls onCancelSnapApproximation:YES, but approximation starts asynchronously in viewDidLoad. If the user cancels before didFinishAllApproximations... applies the preview, OAPlanRouteEditingBridge runs commandManager undo, so it can undo the user’s previous route edit instead of the approximation preview. This should track whether a preview command was actually applied, and cancel the helper while calculation is still running

Sources/Controllers/RoutePlanning/OAGpxApproximationViewController.mm
```diff
@@ -175,47 +147,20 @@ - (void)initData
...
 - (void)didUpdateProgress:(NSInteger)progress
 {
-    if (_isApplying && _progressBarView)
+    if (_progressBarView)
     {
         if (_progressBarView.hidden)
             _progressBarView.hidden = NO;
-        float normalizedProgress = MIN(MAX((float)progress / kProgressMaximumValue, 0.f), 1.f);
-        [_progressBarView setProgress:MAX(_progressBarView.progress, normalizedProgress) animated:YES];
+        _progressBarView.progress = progress;
```
tigrim: What's the issue here? Right now it looks like UIProgressView.progress expects a value in the 0...1 range, but this code assigns the raw percentage from the helper. Values like 10 or 50 will make the progress bar appear full, so the UI may indicate completion while the approximation is still running. If needed, normalize and clamp the value by using progress / 100.f.

Sources/Controllers/PlanRoute/OAPlanRouteEditingBridge.mm
tigrim, line 2332: This path now runs automatically when the approximation preview finishes, but it applies `OAApplyGpxApproximationCommand` even when the result is empty or inconsistent. If `gpxApproximations` is empty, or its count differs from `pointsList`, we can enter approximation mode and push/update a command without a valid preview. Could we mirror the guard used in the elevation path (`approximations.count > 0 && points.count == approximations.count`) before creating/applying the command?

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -217,8 +226,10 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
             dismiss()
             return
         }
-        UIView.animate(withDuration: duration, animations: {
-            self.sheetView.transform = CGAffineTransform(translationX: 0, y: self.height(for: self.sheetState))
+        UIView.animate(withDuration: duration, animations: { [weak self] in
+            guard let self else { return }
```
tigrim: remove guard

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -579,14 +590,73 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
...
-        warningViewController.navigationController?.setNavigationBarHidden(true, animated: false)
-        warningViewController.navigationController?.isModalInPresentation = true
+        let navigationController = UINavigationController(rootViewController: warningViewController)
+        navigationController.setNavigationBarHidden(true, animated: false)
+        navigationController.delegate = self
+        navigationController.view.frame = sheetView.bounds
+        navigationController.view.autoresizingMask = [.flexibleWidth, .flexibleHeight]
+        approximationPreviousSheetState = sheetState
+        approximationNavigationController = navigationController
+        addChild(navigationController)
```
tigrim: use `func add(_ child: UIViewController, frame: CGRect? = nil) {`

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -579,14 +590,73 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
...
+        sheetView.bringSubviewToFront(grabberView)
+        routeTypeButton.isHidden = true
+        crosshairView.isHidden = true
+        updateApproximationPopupHeight(warningViewController, animated: true)
         return true
     }
 
+    private func dismissApproximationPopup() {
+        guard let navigationController = approximationNavigationController else { return }
+        navigationController.willMove(toParent: nil)
```
tigrim: func remove()

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -227,9 +227,8 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
             return
         }
         UIView.animate(withDuration: duration, animations: { [weak self] in
```
tigrim: Don't use [weak self] in the animation block

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -133,6 +136,11 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
         dataProvider.onPointEditModeRequested = { [weak self] mode in
             self?.showPointEditingView(mode: mode)
         }
+        dataProvider.onApproximationPopupDismissed = { [weak self] in
+            DispatchQueue.main.async { [weak self] in
```
tigrim: No need to capture self weakly again here. The outer closure already captures self weakly

Sources/Controllers/RoutePlanning/OAGpxApproximationViewController.mm
```diff
@@ -99,36 +90,27 @@ - (void)applyLocalization
...
-    _isApplying = YES;
     [self setApplyButtonEnabled:NO];
-    [_approximationHelper calculateGpxApproximationAsync];
-}
-
-- (void)finishApplying
-{
-    [_progressBarView setProgress:1.f animated:YES];
     __weak __typeof(self) weakSelf = self;
     [self dismissViewControllerAnimated:YES completion:^{
```
tigrim: In the Plan Route flow this VC is pushed into an embedded navigation controller, not presented modally. `dismissViewControllerAnimated` can be a no-op here, so the completion may not run and `onApplyGpxApproximation` may never be called. Can we call the apply delegate directly and let the parent close the embedded popup?

tigrim (approved):

PR #5626, opened 2026-08-10
Prevent duplicate wiki open alerts

### LINKED ISSUE #5500: Long delay when opening Wikipedia articles from the context menu

vitaliy-sova-ios:


Sources/Helpers/OAWikiArticleHelper.mm
```diff
@@ -42,6 +42,7 @@ @implementation OAWikiArticleSearchTask
     NSString *_name;
     UIView *_sourceView;
     BOOL _isCanceled;
+    BOOL _isHowToOpenAlertShowed;
```
tigrim: Could we reuse the existing `canPresentAlertController(_:completion:)` here instead of adding a task-specific `_isHowToOpenAlertShowed` flag? The duplicate alert happens because `showHowToOpenWikiAlert` is scheduled first and then `onPostExecute` calls `warnAboutExternalLoad` for the same empty result. Both helpers currently walk through `presentedViewController` and present on top of whatever is already shown

tigrim (approved):

PR #5628, opened 2026-08-10
Fixed incorrect zoom out/in for plan route compact mode

aleksandr-tata:
Fixed incorrect zoom out/in for plan route compact mode

### DIFF
--- a/Sources/Controllers/Map/OAMapViewController.h
+++ b/Sources/Controllers/Map/OAMapViewController.h
@@ -88,7 +88,7 @@ typedef NS_ENUM(NSInteger, EOAMapPanDirection) {
 @property (nonatomic) int mapPositionX;
 
 @property(readonly) CGFloat displayDensityFactor;
-@property(readonly) double viewportYScale;
+@property(nonatomic) double viewportYScale;
 
 @property(readonly) OAObservable *framePreparedObservable;
 
--- a/Sources/Controllers/Map/OAMapViewController.h
+++ b/Sources/Controllers/Map/OAMapViewController.h
@@ -88,6 +88,7 @@ typedef NS_ENUM(NSInteger, EOAMapPanDirection) {
 @property (nonatomic) int mapPositionX;
 
 @property(readonly) CGFloat displayDensityFactor;
+@property(readonly) double viewportYScale;
 
 @property(readonly) OAObservable *framePreparedObservable;
 
--- a/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
+++ b/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
@@ -760,6 +760,7 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
         }
         if animated {
             UIView.animate(withDuration: Self.sheetAnimationDuration, animations: updates)
+            crosshairView.layer.removeAnimation(forKey: "position")
         } else {
             updates()
         }
--- a/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
+++ b/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
@@ -54,6 +54,7 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
     private var pendingSegmentPointIndexes: [Int]?
     private var pointEditingView: OAInfoBottomView?
     private var pointEditingHeight: CGFloat?
+    private var cachedMapViewportYScale: Double?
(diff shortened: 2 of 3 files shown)


Sources/Controllers/Map/OAMapViewController.h
```diff
@@ -88,6 +88,7 @@ typedef NS_ENUM(NSInteger, EOAMapPanDirection) {
 @property (nonatomic) int mapPositionX;
 
 @property(readonly) CGFloat displayDensityFactor;
+@property(readonly) double viewportYScale;
```
tigrim: Is this property exposed because PlanRouteScrollableViewController cannot access the Objective-C++ OAMapRendererView directly from Swift? If so, it may be worth documenting that this getter is intended for saving and restoring the current viewport state

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -701,12 +709,20 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
...
         let x = view.bounds.midX
-        guard x > 0 else { return }
+        let screenHeight = currentScreenHeight
+        guard x > 0, screenHeight > 0 else { return }
+        let viewportYScale = Double(2 * centerY / screenHeight)
+        OARootViewController.instance().mapPanel.mapViewController.setViewportScaleY(viewportYScale)
         dataProvider.setCrosshairPosition(screenPoint: CGPoint(x: x, y: centerY))
     }
 
+    private func restoreMapViewport() {
```
tigrim, line 725: Сould this cause a race when switching to Navigation? hide() completes asynchronously, so restoreMapViewport() may run after Navigation has already configured its own viewport and overwrite it

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
tigrim, line 199: Should this call `super.viewWillTransition(to:with:)` and use `size.height` for the viewport recalculation? Relying on `view.bounds.height` inside the transition animation block feels a bit timing-dependent, while `size` already contains the target screen size

tigrim (commented):
Could you clarify the expected behavior for the Position placement: Center / Bottom setting?
The current implementation temporarily overrides viewportYScale to align the map target with the Plan Route crosshair, regardless of this setting, and restores the cached value when the screen is closed. Is this intentional, or should the selected position placement also affect centering inside Plan Route?

tigrim:
image

Sources/Controllers/Map/OAMapViewController.h
tigrim: The property is getter-only, while mutation is exposed through the differently named setViewportScaleY(_:). This makes the API look asymmetric and may suggest that changing the value is unsupported. Could this be exposed as a read-write viewportYScale property, with its setter preserving the existing CarPlay guard, or at least documented?

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -682,42 +692,51 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
...
-        let centerY = crosshairCenterY(sheetHeight: sheetHeight)
-        let x = view.bounds.midX
-        guard x > 0 else { return }
-        dataProvider.setCrosshairPosition(screenPoint: CGPoint(x: x, y: centerY))
+    private func updateCrosshairMapCenter(sheetHeight: CGFloat, screenSize: CGSize? = nil) {
+        let targetScreenSize = screenSize ?? CGSize(width: view.bounds.width, height: currentScreenHeight)
+        let centerY = crosshairCenterY(sheetHeight: sheetHeight, screenHeight: targetScreenSize.height)
+        let centerX = targetScreenSize.width / 2
+        guard centerX > 0, targetScreenSize.height > 0 else { return }
+        let viewportYScale = Double(2 * centerY / targetScreenSize.height)
```
tigrim, line 720: Account for renderer offsets when calculating the viewport scale
This calculation assumes that the renderer center is simply height * viewportYScale / 2, but OAMapRendererView.getCenterPixel() also applies its top and bottom offsets. Plan Route does not override shouldIgnoreTopBottomOffsets, so configured widget offsets can remain active even while those widgets are hidden in context-menu mode. In that case, the actual map center will not match the crosshair and the incorrect zoom-center behavior can still occur. Please either clear these offsets while Plan Route is visible, as the previous route-planning controller does, or include them in this calculation.

tigrim (approved):

PR #5630, opened 2026-08-11
Use new separator color ios

### LINKED ISSUE #5520: Use new separator color for iOS 26

DmitrySvetlichny:


Sources/Controllers/RouteInfoMenu/OARoutingTargetCell.m
```diff
@@ -7,30 +7,32 @@
...
 - (void) awakeFromNib
 {
     [super awakeFromNib];
 
-    // Initialization code
-    _divider = [CALayer layer];
-    _divider.backgroundColor = [[UIColor colorWithWhite:0.50 alpha:0.3] CGColor];
-    [self.contentView.layer addSublayer:_divider];
+    _divider = [[SeparatorView alloc] init];
+    _divider.userInteractionEnabled = NO;
```
tigrim: ?

Sources/Views/SeparatorView.swift
```diff
@@ -0,0 +1,113 @@
...
+    static var color: UIColor {
+        if #available(iOS 26.0, *) {
+            return .customSeparatorSolid
+        }
+        return .customSeparator
+    }
+
+    private override init() {}
+
+    static func thickness() -> CGFloat {
```
tigrim: static var

tigrim (approved):

PR #5631, opened 2026-08-11
Plan route fixes for analyze tab

### LINKED ISSUE #3527: Analyze tab – Plan a route
# Tasks

> * Android tasks – https://github.com/osmandapp/OsmAnd/issues/18872
> * Related duplicate - #3682

**Related:**
1. #1323
2. #3527
3. #5668
4. #5609

# ToDo, [Figma](https://www.figma.com/design/DyC3BkWuaDb5pFvLvYnr8k/Plan-a-route?node-id=1805-55067&m=dev)

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

The overall layout will be similar to the Route Details screen but with some minor redesigns to certain elements. 

- [x] **Overview**
- [x] **Overview statistic** 
- [x] **Road types**
- [x] **Steepness**
- [x]  **Surface**
- [x] **Smoothness**

### Overview

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

This UI element is similar the one we user in Route details, but we need to redesign Axis selection buttons.

- [x] Show graph for the selected axis

**Axis selection buttons**:

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

- [x] Background color – **cellButtonBg**
- [x] Corner radius – 
- [x] Show popover symbol on the right side `chevron.up.chevron.down`

**Action**
- [x] **Recalculate elevation** – should launch **Get elevation data** bottom sheet.

### Overview statistic

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

- [x] Display a 3x2 native data grid containing
- [x] Row 1: **Uphill**, **Downhill**, **Alt. range**
- [x] Row 2: **Avg. speed**, **Max. speed**, **Time in motion**
- [x] Show `–` if data is missing. For example for planned routes Time in motion, and Speed will be missed.
- [x] Data text size – `footnote`
	- [x] Value text color – `textColorActive`
	- [x] Data field name text color – `textColorSecondary`

### Road types / Steepness / Surface / Smoothness

<table style="width:100%">
  <tr>
    <th>Collapsed</th>
    <th>Open</th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
  </tr>
</table>   

**Collapsed State (Default):**
- [x] Header with trailing chevron icon (>).
- [x] Horizontal graph (Previous version of it is used in Route details)
- [x] Compact inline legend with color dots and category labels below the bar.

**Open state**

- [x] Show the same horizontal graph
- [x] detailed legend for the graph

## Empty state, [Figma](https://www.figma.com/design/DyC3BkWuaDb5pFvLvYnr8k/Plan-a-route?node-id=1805-68423&m=dev)

<table style="width:100%">
  <tr>
    <th>Empty state</th>
    <th>Bottom sheet</th>
    <th>Calculation</th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
    <td> </td>
  </tr>
</table>  

- [x] Show if track doesn't have any data to show
- [x] Title – `No elevation data`
- [x] Description – `OsmAnd can retrieve elevation data from nearby roads or terrain maps.`
- [x] Action – **Get elevation data**
	- [x] Show **Get elevation data** bottom sheet on tap

### Get elevation data

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

- [x] Title – `Get elevation data`
- [x] Description – `Select a method to calculate elevation using data from offline maps or terrain maps.`

**Actions:**
- [x] **Use nearby roads**
	- [x] Icon – `ic_custom_attach_track`
	- [x] Second line – `May adjust track geometry`
- [x] **Use Terrain maps**
	- [x] Icon – `ic_custom_terrain`
	- [x] Second line – `Track geometry stays unchanged`
- [x] Tap on action should trigger elevaton calculation using one of the selected method

### Calculating elevation...

<table style="width:100%">
  <tr>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
  </tr>
</table>   

- [x] Show while calculating 
- [x] Description, depend on selected method:
	- [x] For **Use nearby roads** – `Adjusting track to nearby roads to retrieve altitude data.`
	- [x] For **Use Terrain maps** – `Retrieving altitude data from terrain maps.`
- [x] Action – **Cancel**
	- [x] Should stop elevation calculation

aleksandr-tata:
## Summary

Aligned the iOS Plan Route Analyze flow with Android.

## Problem

On iOS, the altitude/speed graph was empty when opening a GPX track, even when the track already contained the required data. Average speed, maximum speed, and time in motion were also unavailable before elevation recalculation.

After recalculation, the iOS speed statistics differed significantly from Android because iOS used routing-profile speeds instead of the original GPX timestamps.

## Changes

- Display the altitude/speed graph immediately when the GPX contains the required data.
- Display average speed, maximum speed, and time in motion before elevation recalculation.
- Make **Recalculate elevation** trigger the complete elevation and route-data calculation flow.
- Display Road type, Surface, Smoothness, and Steepness after the enriched data becomes available.
- Preserve the original GPX timestamps during route approximation.
- Recalculate segment speed, time, and distance using the original timestamps instead of routing-profile speeds.
- Propagate the recalculated timestamps and speeds to the exported track and iOS UI state.

## Result

The initial graph and available track statistics are now displayed before recalculation. After elevation recalculation, the speed statistics are much closer to the Android results.

### DIFF
--- a/Sources/Controllers/RoutePlanning/OAGpxApproximationHelper.mm
+++ b/Sources/Controllers/RoutePlanning/OAGpxApproximationHelper.mm
@@ -17,6 +17,96 @@
 #import "OAGPXDatabase.h"
 #import "OAGpxData.h"
 
+#include <gpxRouteApproximation.h>
+#include <routeSegment.h>
+#include <routeSegmentResult.h>
+
+static BOOL hasValidExternalTimestamps(NSArray<OASWptPt *> *points)
+{
+    if (points.count == 0)
+        return NO;
+
+    int64_t lastTimestamp = 0;
+    for (OASWptPt *point in points)
+    {
+        if (point.time == 0 || point.time < lastTimestamp)
+            return NO;
+        lastTimestamp = point.time;
+    }
+    return YES;
+}
+
+static float externalSpeed(const SHARED_PTR<GpxPoint> &gpxPoint,
+                           const SHARED_PTR<RouteSegmentResult> &segment,
+                           NSArray<OASWptPt *> *sourcePoints)
+{
+    NSInteger startIndex = gpxPoint->ind;
+    NSInteger endIndex = gpxPoint->targetInd;
+    if (endIndex == -1 && startIndex >= 0 && startIndex + 1 < sourcePoints.count)
+        endIndex = startIndex + 1;
+
+    if (startIndex < 0 || endIndex <= 0 || startIndex >= endIndex || endIndex >= sourcePoints.count)
+        return segment->segmentSpeed;
+
+    int64_t duration = sourcePoints[endIndex].time - sourcePoints[startIndex].time;
+    if (duration <= 0)
+        return segment->segmentSpeed;
+
+    double distance = 0;
+    for (NSInteger index = startIndex; index < endIndex; index++)
+    {
+        OASWptPt *firstPoint = sourcePoints[index];
+        OASWptPt *secondPoint = sourcePoints[index + 1];
+        distance += getDistance(firstPoint.getLatitude,
+                                firstPoint.getLongitude,
+                                secondPoint.getLatitude,
+                                secondPoint.getLongitude);
+    }
+    return distance > 0 ? distance / (duration / 1000.0) : segment->segmentSpeed;
+}
+
+static void recalculateTimeAndDistance(const vector<SHARED_PTR<RouteSegmentResult>> &segments)
+{
+    for (const auto &segment : segments)
+    {
+        float speed = segment->segmentSpeed;
+        if (speed == 0)
+            continue;
+
+        BOOL isForward = segment->getStartPointIndex() < segment->getEndPointIndex();
+        double distance = 0;
+        for (NSInteger index = segment->getStartPointIndex(); index != segment->getEndPointIndex();)
+        {
+            NSInteger nextIndex = isForward ? index + 1 : index - 1;
+            distance += measuredDist31(segment->object->getPoint31XTile((int)index),
+                                       segment->object->getPoint31YTile((int)index),
+                                       segment->object->getPoint31XTile((int)nextIndex),
+                                       segment->object->getPoint31YTile((int)nextIndex));
+            index = nextIndex;
+        }
+        segment->segmentTime = distance / speed;
+        segment->segmentSpeed = speed;
+        segment->distance = distance;
+    }
+}
+
+static void applyExternalTimestamps(OAGpxRouteApproximation *approximation,
+                                    OALocationsHolder *locationsHolder)
+{
+    NSArray<OASWptPt *> *sourcePoints = locationsHolder.getWptPtList;
+    if (approximation == nil || approximation.gpxApproximation == nullptr)
+        return;
+    if (!hasValidExternalTimestamps(sourcePoints))
+        return;
+
+    for (const auto &gpxPoint : approximation.gpxApproximation->finalPoints)
+    {
+        for (const auto &segment : gpxPoint->routeToTarget)
+            segment->segmentSpeed = externalSpeed(gpxPoint, segment, sourcePoints);
+        recalculateTimeAndDistance(gpxPoint->routeToTarget);
+    }
+}
+
 @interface OAGpxApproximationHelper () <OAGpxApproximationProgressDelegate>
 
 @end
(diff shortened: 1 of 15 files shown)


Sources/Controllers/RoutePlanning/OAGpxApproximationHelper.mm
```diff
@@ -105,13 +195,16 @@ - (void)approximateMultipleGpxAsync:(NSMutableArray<OAGpxApproximator *> *)appro
         OAGpxApproximator *gpxApproximator = approximationsToDo.firstObject;
         [approximationsToDo removeObjectAtIndex:0];
         _currentApproximator = gpxApproximator;
+        __weak __typeof(self) weakSelf = self;
         [gpxApproximator calculateGpxApproximation:[[OAResultMatcher alloc] initWithPublishFunc:^BOOL(OAGpxRouteApproximation *__autoreleasing *approxPtr) {
             OAGpxRouteApproximation *strongApprox = (approxPtr && *approxPtr) ? *approxPtr : nil;
+            applyExternalTimestamps(strongApprox, gpxApproximator.locationsHolder);
```
tigrim: Android applies external timestamps only when `useExternalTimestamps` is enabled and propagates this decision through the routing API. Here the adjustment is applied automatically whenever timestamps are valid, so callers cannot opt out and use routing-profile speeds. Please introduce an explicit `useExternalTimestamps` parameter in `OAGpxApproximator`/routing API and call this logic conditionally.

Sources/Controllers/RoutePlanning/OAMeasurementEditingContext.mm
```diff
@@ -1368,7 +1457,21 @@ - (OASTrkSegment *)getRouteSegment:(NSInteger)startPointIndex endPointIndex:(NSI
...
     {
         [_before.points[startPointIndex] setTrkPtIndexIndex:0];
-        return [[[OARouteExporter alloc] initWithName:@"" route:route locations:locations routePointIndexes:routePointIndexes points:nil] generateRouteSegment];
+        OARouteExporter *routeExporter = [[OARouteExporter alloc] initWithName:@""
+                                                                        route:route
+                                                                    locations:locations
+                                                            routePointIndexes:routePointIndexes
+                                                                       points:nil];
+        OASTrkSegment *routeSegment = [routeExporter generateRouteSegment];
+        NSInteger pointsCount = MIN(routeSegment.points.count, sourcePoints.count);
```
tigrim: Could we preserve time and speed through `OARouteExporter` instead of copying them back by matching array indexes? Android puts these values into `Location` and lets `RouteExporter` serialize them. The current parallel `sourcePoints` array and `MIN(...)` hide possible count/order mismatches and duplicate exporter responsibility. Please construct `CLLocation` with the original timestamp/speed and make `OARouteExporter` preserve timestamp.

tigrim:
Android treats elevation and speed availability independently (`hasElevationData()` vs `isSpeedSpecified()`). On iOS, the Analyze screen still uses `hasElevationData` as the condition for showing both the graph and statistics. Please verify a GPX with timestamps/speed but without elevation: it appears that early analysis will run, but the screen will still resolve to `.noData`.

Sources/Controllers/RoutePlanning/OARouteExporter.mm
```diff
@@ -170,6 +170,10 @@ - (OASTrkSegment *)generateRouteSegment
         OASWptPt *pt = [[OASWptPt alloc] initWithLat:loc.coordinate.latitude lon:loc.coordinate.longitude];
         if (loc.speed > 0)
             pt.speed = loc.speed;
+
+        NSTimeInterval timestamp = loc.timestamp.timeIntervalSince1970;
```
tigrim: `OARouteExporter` is also used by `generateGPXFileWithRoute()`, where route locations are generated with `[NSDate date]` rather than real track timestamps. Serializing every positive `CLLocation.timestamp` will therefore add near-identical synthetic timestamps to ordinary exported routes, which can later produce unrealistic speed statistics and enable external-timestamp processing incorrectly. Android avoids this because generated route `Location` objects have `time == 0`. Please either make timestamp preservation explicit for the measurement-tool export path or ensure synthetic route locations are exported without timestamps.

Sources/Router/OALocationsHolder.mm
```diff
@@ -110,6 +110,27 @@ - (double) getLongitude:(NSInteger)index
 	return 0;
 }
 
+- (int64_t)timeAtIndex:(NSInteger)index
+{
+	switch (_locationType)
+	{
+		case LOCATION_TYPE_LOCATION:
```
tigrim: Android’s `LocationsHolder.getTime()` returns a timestamp only for `WptPt` and returns `0` for `Location`. Here any `CLLocation` timestamp can enable external-timestamp processing, even though regular `CLLocation` instances commonly contain their creation time rather than a real GPX timestamp. Could we keep this behavior aligned with Android and return time only for `LOCATION_TYPE_WPTPT`?

tigrim (approved):

PR #5640, opened 2026-08-13
Added debug logs for Context menu

### LINKED ISSUE #5337: Context menu opens twice as slowly on iOS compared to Android

aleksandr-tata:
Added debug logs for Context menu

### DIFF
--- a/Sources/Helpers/OAAmenitySearcher.mm
+++ b/Sources/Helpers/OAAmenitySearcher.mm
@@ -31,11 +31,11 @@
 #import "OAMapPanelViewController.h"
 #import "OAMapViewController.h"
 #import "OAMapRendererView.h"
-#import "OAOperationLog.h"
 #import "Localization.h"
 #import "OANativeUtilities.h"
 #import "OrderedDictionary.h"
 #import "OsmAnd_Maps-Swift.h"
+#import <QuartzCore/QuartzCore.h>
 
 
 #include <OsmAndCore/CommonTypes.h>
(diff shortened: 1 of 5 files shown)


Sources/Helpers/OAAmenitySearcher.mm
```diff
@@ -330,13 +331,22 @@ - (nullable BaseDetailsObject *)searchDetailedObjectWithRequest:(OAAmenitySearch
     NSInteger radius = [request.type isEqualToString:kEntityTypeRelation]
         ? AMENITY_SEARCH_RADIUS_FOR_RELATION
         : AMENITY_SEARCH_RADIUS;
+    NSUInteger repositoryCount = [self getAmenityRepositories:YES].count;
```
tigrim: ?

Sources/Views/OATargetPointView.mm
```diff
@@ -887,16 +889,19 @@ - (void) show:(BOOL)animated onComplete:(void (^)(void))onComplete
             frame.origin.y = 0;
         }
         
+        __weak __typeof(self) weakSelf = self;
         [UIView animateWithDuration:0.2 animations:^{
             
-            self.frame = frame;
+            weakSelf.frame = frame;
```
tigrim: weakSelf?

tigrim (commented):
OAOperationLog is redundant here and is being used outside its intended purpose. Since NSLog is already persisted to a file, it should be sufficient to capture a monotonic timestamp when opening starts and use a single NSLog to report the total duration after the animation completes. This will reduce logging noise and minimize the impact of logging on the measurement

tigrim (approved):

PR #5650, opened 2026-08-14
Fix Plan Route save behavior and track naming

aleksandr-tata:
### Summary

- Restored the legacy unique naming logic for new routes:
  - `date`
  - `date_(2)`
  - `date_(3)`
- Prevented routes created on the same day from overwriting each other.
- Updated the Navbar Save action to skip the save dialog and show the saved-track bottom sheet.
- Restored the previous `Save as copy` behavior:
  - opens the save dialog;
  - uses the original track name without automatically adding `_copy`;
  - preserves the original track folder.

### DIFF
--- a/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
+++ b/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
@@ -67,6 +67,13 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
         }
     }
 
+    private var suggestedFilePath: String? {
+        guard case .editTrack(let fileName) = dataProvider.mode,
+              let gpxFileName = (fileName as NSString).appendingPathExtension("gpx") else { return nil }
+        guard let folder = dataProvider.editTrackFolder, !folder.isEmpty else { return gpxFileName }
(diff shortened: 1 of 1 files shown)


Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -963,14 +964,30 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
         return true
     }
 
-    private func presentSaveDialog(duplicate: Bool) {
-        isPendingSaveAsCopy = duplicate
+    private func presentSaveAsCopyDialog() {
+        isPendingSaveAsCopy = true
         pendingSegmentPointIndexes = nil
-        guard let vc = OASaveTrackViewController(fileName: suggestedFileName, filePath: nil, showOnMap: true, simplifiedTrack: false, duplicate: duplicate) else { return }
+        guard let vc = OASaveTrackViewController(fileName: suggestedFileName, filePath: suggestedFilePath, showOnMap: true, simplifiedTrack: false, duplicate: false) else { return }
```
tigrim: Should we validate the initial file name here? Otherwise, tapping Save immediately may overwrite the original track instead of creating a copy.

tigrim (approved):

PR #5653, opened 2026-08-15
Task 1323 plan route UI review 14 08

aleksandr-tata:
-  3D Mode button should be visible if it visible on the main map
-  Don't calculate route and don't show progress for one point
-  Align padding size between segmented control and empty state / graph. Use bigger padding from POI tab
-  Why text in those headers has different size and paddings
-  Don't move map after opening Plan a Route

### DIFF
--- a/Sources/Views/OAMapRendererView.mm
+++ b/Sources/Views/OAMapRendererView.mm
@@ -450,6 +450,21 @@ forcedUpdate:(BOOL)forcedUpdate
     return _renderer->getState().fixedPixel;
 }
 
+- (CGPoint)mapTargetScreenPoint
+{
+    CGFloat scale = self.contentScaleFactor;
+    OsmAnd::PointI targetScreenPoint = _renderer->getState().fixedPixel;
+    return CGPointMake(targetScreenPoint.x / scale, targetScreenPoint.y / scale);
+}
+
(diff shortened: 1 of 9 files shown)


Sources/Views/OAMapRendererView.mm
```diff
@@ -450,6 +450,21 @@ - (void)setMyLocationSectorRadius:(float)radius
     return _renderer->getState().fixedPixel;
 }
 
+- (CGPoint)mapTargetScreenPoint
```
tigrim, line 453: How is the bottom sheet offset handled on other screens? These methods aren't used there, are they?

aleksandr-tata: Correct, these methods are currently used only by Plan Route. Other bottom sheets adjust viewportXScale/viewportYScale, and the renderer moves the map when its calculated center changes.
Plan Route additionally requires the map not to move on initial presentation, while the crosshair must still become the zoom anchor. After a subsequent sheet change, the same geographic target must be moved under the crosshair. A viewport-only update cannot represent both operations and does nothing when the calculated viewport value is already equal to the current one.
reanchorMapTarget changes fixedPixel to the crosshair without visually moving the map. Setting mapTargetScreenPoint keeps the existing geographic target and moves it to the crosshair after the sheet position changes.

PR #5656, opened 2026-08-17
Task 1323 plan route UI review 17 08

aleksandr-tata:
- NavBar title color – always white
- Show menu in closed stated for New route and in the Middled for editing after opening
- iOS 18: Save button shape
- iOS 18: Done button is hardly visible
- POI and Route buttons paddings and truncation
- Glass style effect for POI and Route buttons

### DIFF
--- a/Sources/Controllers/PlanRoute/PlanRouteToolbarsView.swift
+++ b/Sources/Controllers/PlanRoute/PlanRouteToolbarsView.swift
@@ -68,7 +68,6 @@ final class PlanRouteTopToolbarView: TouchesPassView {
     }
 
     func updateMapTheme() {
-        titleLabel.textColor = UIColor.textColorPrimary.currentMapThemeColor
         closeButton.updateColors(forPressedState: false)
         optionsButton.updateColors(forPressedState: false)
         saveButton.updateColors(forPressedState: false)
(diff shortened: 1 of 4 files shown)


Sources/Controllers/PlanRoute/PlanRouteToolbarsView.swift
```diff
@@ -79,6 +78,7 @@ final class PlanRouteTopToolbarView: TouchesPassView {
         setupBackgroundView()
 
         titleLabel.font = .scaledSystemFont(ofSize: 17, weight: .semibold, maximumSize: 22)
+        titleLabel.textColor = .white
```
tigrim, line 81: The system .white color is rarely used for titleLabel in the design. Should we use textColorPrimary instead?

Sources/Controllers/PlanRoute/SegmentRouteSettingsViewController.swift
```diff
@@ -81,7 +81,13 @@ final class SegmentRouteSettingsViewController: UIViewController {
             navigationItem.leftBarButtonItem = closeButton
         }
 
-        let checkmarkImage = UIImage.icCheckmarkDefault.withTintColor(.white, renderingMode: .alwaysOriginal)
+        let checkmarkColor: UIColor
+        if #available(iOS 26.0, *) {
```
tigrim, line 85: Why is a different color used for different OS versions? If this was agreed upon, please attach screenshots from different OS versions to the task.

Sources/Controllers/PlanRoute/PlanRouteToolbarsView.swift
aleksandr-tata: White color for title is request of @dmpr0

Sources/Controllers/PlanRoute/SegmentRouteSettingsViewController.swift
aleksandr-tata: IMG_0750
IMG_9595

tigrim (approved):

PR #5657, opened 2026-08-18
Carplay multi destination routes

### LINKED ISSUE #5391: Multi-destination routes are not saved correctly in navigation history on CarPlay and Android Auto

vitaliy-sova-ios:


Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -159,6 +161,104 @@ - (void) enterBrowsingState
     [self onMap3dModeUpdated];
 }
 
+- (NSString *)fullAddressForPoint:(OARTargetPoint *)point
```
tigrim: Reverse geocoding is performed synchronously while building the CarPlay route preview. poiNameForPoint, addressForPoint and intermediateSubtitleForPoints may trigger several OBF lookups before the UI is presented. This can block the main thread, especially when map data is large or unavailable. Please avoid synchronous geocoding here and resolve the addresses asynchronously before updating the trip preview

Resources/Localizations/en.lproj/Localizable.stringsdict
```diff
@@ -61,5 +61,21 @@
 			<string>%ld points</string>
 		</dict>
 	</dict>
+    <key>carplay_trip_via_stops</key>
```
tigrim, line 64: Why wasn't this added for all locales?

tigrim (commented):
fix merge conflict

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -641,12 +641,19 @@ - (void)lookupAddressForMyLocationPoint
...
 {
-    if (_pointToNavigate != nil && [_pointToNavigate isSearchingAddress] && !_isSearchingDestination)
+    BOOL isNameNotValid = _pointToNavigate != nil && [_pointToNavigate isSearchingAddress];
+    BOOL isAddresEmpty = _pointToNavigate != nil && NSStringIsEmpty(_pointToNavigate.pointDescription.address);
+    
+    if ((isNameNotValid || isAddresEmpty) && !_isSearchingDestination)
     {
         _isSearchingDestination = YES;
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^(void) {
             NSString *pointName = [self getLocationName:_pointToNavigate.point];
```
tigrim: Capture the target point and its location before dispatching this work, then apply the result on the main queue only if `_pointToNavigate` still matches the captured target. Otherwise, if the destination changes while reverse geocoding is in progress, the old address can be assigned to the new destination.

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -641,12 +641,19 @@ - (void)lookupAddressForMyLocationPoint
...
-            [_pointToNavigate.pointDescription setName:pointName];
+            if (isNameNotValid)
+                [_pointToNavigate.pointDescription setName:pointName];
+            
+            _pointToNavigate.pointDescription.address = pointName;
+                
             [_app.data setPointToNavigate:_pointToNavigate];
```
tigrim: `setPointToNavigate:` is not only a persistence setter: it also adds an `OAHistoryItem`. The destination was already recorded when it was initially selected, so calling this setter again after reverse geocoding creates a duplicate navigation-history entry. Update and persist the address without invoking the history-producing setter.

Sources/Models/OAPointDescription.h
```diff
@@ -51,6 +51,7 @@ static const int POINT_LOCATION_LIST_HEADER = 201;
 @property (nonatomic) NSString *name;
 @property (nonatomic) NSString *typeName;
 @property (nonatomic) NSString *iconName;
+@property (nonatomic) NSString *address;
```
tigrim, line 54: Include `address` in the backward-compatible serialization and deserialization format. `OARTargetPoint` persists its point description through `serializeToString:`, which currently stores only the type, name, and icon. As a result, the exact POI address is lost after archiving or relaunching the app.

Sources/CarPlay/CarPlayDirectionsControllers/OACarPlayCategoryResultListController.mm
```diff
@@ -227,6 +227,15 @@ - (void)onItemSelected:(CPListItem * _Nonnull)item completionHandler:(dispatch_b
     }
     CLLocation *loc = searchItem.getSearchResult.location;
     OAPointDescription *historyName = [[OAPointDescription alloc] initWithType:POINT_TYPE_POI name:[searchItem getName]];
+    
+    OASearchResult *searchResult = searchItem.getSearchResult;
+    if (searchResult.objectType == EOAObjectTypePoi) {
```
tigrim: code style

Sources/CarPlay/CarPlayDirectionsControllers/OACarPlayCategoryResultListController.mm
```diff
@@ -227,6 +227,15 @@ - (void)onItemSelected:(CPListItem * _Nonnull)item completionHandler:(dispatch_b
...
+        NSString *address = searchResult.alternateName;
+        if (address.length > 0 && ![address isEqualToString:historyName.name]) {
```
tigrim: up

Sources/Models/OAPointDescription.h
vitaliy-sova-ios: That string format is shared with Android (PointDescription.serializeToString), which currently persists only type / name / icon. Adding an address segment would diverge from Android and break cross-platform settings/history sync (and older clients would also mis-parse the new # segments as part of iconName).

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -639,30 +639,56 @@ - (void)lookupAddressForMyLocationPoint
...
+    
+    if (isNameNotValid || isAddresEmpty)
     {
         dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^(void) {
             NSString *pointName = [self getLocationName:point.point];
-            [point.pointDescription setName:pointName];
+            if (isNameNotValid)
+                [point.pointDescription setName:pointName];
+            
+            point.pointDescription.address = pointName;
```
tigrim: Mirror the destination lookup pattern here: capture the location before dispatching, do the reverse-geocode work in the background, then update `pointDescription` on the main queue only after checking that this point is still the intended first intermediate. In the new `isAddresEmpty` path this can currently write an address into a waypoint after the intermediate list has changed, while listeners are refreshed for stale data.

Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -5032,3 +5032,5 @@
 "calculating_elevation_terrain_maps_description" = "Retrieving altitude data from terrain maps.";
 "recalculate_elevation" = "Recalculate elevation";
 "average_speed" = "Avg. speed";
+
+"carplay_trip_via_point" = "Via %@";
```
tigrim, line 5036: Add this key to the other Localizable.strings files as well. `carplay_trip_via_stops` is localized in every stringsdict, but `carplay_trip_via_point` exists only in `en.lproj`, so the one-intermediate-point CarPlay subtitle falls back to English in all non-English locales.

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -50,6 +50,7 @@
 
 static const CGFloat kTurnArrowSize = 16.0;
 static const CGFloat kTurnArrowRenderSize = 32.0;
+static const NSInteger kTripSubtitleMaxLength = 30;
```
tigrim, line 53: ?

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -639,30 +639,56 @@ - (void)lookupAddressForMyLocationPoint
...
-            NSString *pointName = [self getLocationName:_pointToNavigate.point];
-            [_pointToNavigate.pointDescription setName:pointName];
-            [_app.data setPointToNavigate:_pointToNavigate];
-            dispatch_async(dispatch_get_main_queue(), ^(void) {
-                [self updateRouteAndRefresh:NO];
-                _isSearchingDestination = NO;
+        CLLocation *location = destination.point;
+        __weak __typeof(self) weakSelf = self;
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+            __strong __typeof(weakSelf) self = weakSelf;
```
tigrim: strongSelf

Sources/Router/OATargetPointsHelper.mm
tigrim:
```
- (void)lookupAddressForIntermediatePoint:(OARTargetPoint *)point
{
    BOOL isNameNotValid = point != nil && [point isSearchingAddress];
    BOOL isAddressEmpty = point != nil
        && NSStringIsEmpty(point.pointDescription.address)
        && _intermediatePoints.firstObject == point;

    if (isNameNotValid || isAddressEmpty)
    {
        OARTargetPoint *targetPoint = point;
        CLLocation *location = targetPoint.point;

        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
            __strong __typeof(weakSelf) strongSelf = weakSelf;
            if (!strongSelf)
                return;

            NSString *pointName = [strongSelf getLocationName:location];

                __strong __typeof(weakSelf) strongSelf = weakSelf;
                if (!strongSelf || strongSelf->_intermediatePoints.firstObject != targetPoint)
                    return;

                if (isNameNotValid)
                    [targetPoint.pointDescription setName:pointName];

                targetPoint.pointDescription.address = pointName;
                [strongSelf->_app.data backupTargetPoints];
                [strongSelf updateRouteAndRefresh:NO];
            });
        });
    }
}
```

tigrim (approved):

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -639,32 +639,71 @@ - (void)lookupAddressForMyLocationPoint
...
+        dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+            __strong __typeof(weakSelf) strongSelf = weakSelf;
+            if (!strongSelf)
+                return;
+
+            NSString *pointName = [strongSelf getLocationName:location];
+
+            dispatch_async(dispatch_get_main_queue(), ^{
+                __strong __typeof(weakSelf) strongSelf = weakSelf;
+                if (!strongSelf || strongSelf->_intermediatePoints.firstObject != targetPoint)
```
alex-dev-neo: Is it only accepts a lookup result when the target is currently the first intermediate? What about others?

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -639,32 +639,71 @@ - (void)lookupAddressForMyLocationPoint
...
 
-- (void) lookupAddressForDestinationPoint
+- (void)lookupAddressForDestinationPoint
 {
-    if (_pointToNavigate != nil && [_pointToNavigate isSearchingAddress] && !_isSearchingDestination)
+    OARTargetPoint *destination = _pointToNavigate;
+    BOOL isNameNotValid = destination != nil && [destination isSearchingAddress];
+    BOOL isAddressEmpty = destination != nil && NSStringIsEmpty(destination.pointDescription.address);
+
+    if ((isNameNotValid || isAddressEmpty) && !_isSearchingDestination)
```
alex-dev-neo: Changing the destination during lookup leaves the replacement unresolved?

Sources/Models/OAPointDescription.h
alex-dev-neo: It is not clear - how it will be synchronized with android?

Sources/CarPlay/CarPlayDirectionsControllers/OACarPlayCategoryResultListController.mm
```diff
@@ -227,6 +227,17 @@ - (void)onItemSelected:(CPListItem * _Nonnull)item completionHandler:(dispatch_b
     }
     CLLocation *loc = searchItem.getSearchResult.location;
     OAPointDescription *historyName = [[OAPointDescription alloc] initWithType:POINT_TYPE_POI name:[searchItem getName]];
+    
+    OASearchResult *searchResult = searchItem.getSearchResult;
+    if (searchResult.objectType == EOAObjectTypePoi)
+    {
+        NSString *address = searchResult.alternateName;
```
alex-dev-neo, line 234: Is alternateName always an address?

PR #5658, opened 2026-08-18
UI review fixes 18 08

aleksandr-tata:
- Use monospacedDigitSystemFont for Distance, Travel duration, and ETA
- Use monospacedDigitSystemFont for Uphill / Downhill
- Use proper formatting for Travel duration (e.g., 1 h 17 m). See how it is done in Route details.
- Drop seconds from Estimated Time of Arrival (ETA) – 17:47

-  Add gap info between segments. Should show staright line distance and angle between previous segment last point and next segment first point

-  Show total length for Complex segment on the second line below title

-  Show empty segments (without points) in Route between points, so user can change route type for them

-  Each segment should have Start and Destination points subtitle on first and last points

### DIFF
--- a/Sources/Controllers/PlanRoute/Cells/PlanRouteSegmentGapCell.swift
+++ b/Sources/Controllers/PlanRoute/Cells/PlanRouteSegmentGapCell.swift
@@ -1,0 +1,94 @@
+//
+//  PlanRouteSegmentGapCell.swift
+//  OsmAnd Maps
+//
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import UIKit
+
+final class PlanRouteSegmentGapCell: UITableViewCell {
+
+    private static let horizontalInset: CGFloat = 16
+    private static let topInset: CGFloat = 10
+    private static let iconSize: CGFloat = 30
+    private static let iconTitleSpacing: CGFloat = 16
+    private static let cornerRadius: CGFloat = 26
+
+    private let borderView = UIView()
+    private let iconView = UIImageView()
+    private let titleLabel = UILabel()
+
+    override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+        setupCell()
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        updateBorderColor()
+    }
+
+    func configure(title: String) {
+        titleLabel.text = title
+        accessibilityLabel = title
+    }
+
+    private func setupCell() {
+        let horizontalInset = Self.horizontalInset
+        let topInset = Self.topInset
+        let iconSize = Self.iconSize
+        let iconTitleSpacing = Self.iconTitleSpacing
+        let cornerRadius = Self.cornerRadius
+
+        backgroundColor = .clear
+        contentView.backgroundColor = .clear
+        selectionStyle = .none
+        separatorInset = UIEdgeInsets(top: 0, left: .greatestFiniteMagnitude, bottom: 0, right: 0)
+        isAccessibilityElement = true
+
+        borderView.layer.borderWidth = 1
+        borderView.layer.cornerRadius = cornerRadius
+        updateBorderColor()
+
+        iconView.image = .icCustomSegmentsGap
+        iconView.tintColor = .iconColorDefault
+        iconView.contentMode = .scaleAspectFit
+        iconView.isAccessibilityElement = false
+
+        titleLabel.font = .scaledSystemFont(ofSize: 17)
+        titleLabel.textColor = .textColorSecondary
+        titleLabel.adjustsFontForContentSizeCategory = true
+
+        borderView.translatesAutoresizingMaskIntoConstraints = false
+        contentView.addSubview(borderView)
+        [iconView, titleLabel].forEach {
+            $0.translatesAutoresizingMaskIntoConstraints = false
+            borderView.addSubview($0)
+        }
+
+        NSLayoutConstraint.activate([
+            borderView.topAnchor.constraint(equalTo: contentView.topAnchor, constant: topInset),
+            borderView.leadingAnchor.constraint(equalTo: contentView.leadingAnchor),
+            borderView.trailingAnchor.constraint(equalTo: contentView.trailingAnchor),
+            borderView.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
+
+            iconView.leadingAnchor.constraint(equalTo: borderView.leadingAnchor, constant: horizontalInset),
+            iconView.centerYAnchor.constraint(equalTo: borderView.centerYAnchor),
+            iconView.widthAnchor.constraint(equalToConstant: iconSize),
+            iconView.heightAnchor.constraint(equalToConstant: iconSize),
+
+            titleLabel.leadingAnchor.constraint(equalTo: iconView.trailingAnchor, constant: iconTitleSpacing),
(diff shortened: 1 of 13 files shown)


Sources/Controllers/PlanRoute/Cells/PlanRouteSegmentGapCell.swift
```diff
@@ -0,0 +1,94 @@
...
+//  OsmAnd Maps
+//
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import UIKit
+
+final class PlanRouteSegmentGapCell: UITableViewCell {
+
+    private static let horizontalInset: CGFloat = 16
```
tigrim: use private enum Constants

Sources/Controllers/PlanRoute/Cells/PlanRouteSegmentGapCell.swift
```diff
@@ -0,0 +1,94 @@
...
+            borderView.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
+
+            iconView.leadingAnchor.constraint(equalTo: borderView.leadingAnchor, constant: horizontalInset),
+            iconView.centerYAnchor.constraint(equalTo: borderView.centerYAnchor),
+            iconView.widthAnchor.constraint(equalToConstant: iconSize),
+            iconView.heightAnchor.constraint(equalToConstant: iconSize),
+
+            titleLabel.leadingAnchor.constraint(equalTo: iconView.trailingAnchor, constant: iconTitleSpacing),
+            titleLabel.trailingAnchor.constraint(lessThanOrEqualTo: borderView.trailingAnchor, constant: -horizontalInset),
+            titleLabel.centerYAnchor.constraint(equalTo: borderView.centerYAnchor)
```
tigrim, line 89: Large Dynamic Type / Accessibility font sizes might cause titleLabel to overflow outside borderView vertically

Sources/Controllers/PlanRoute/Cells/PlanRouteSegmentGapCell.swift
```diff
@@ -0,0 +1,94 @@
...
+        ])
+    }
+
+    private func updateBorderColor() {
+        borderView.layer.borderColor = UIColor.customSeparatorSolid.resolvedColor(with: traitCollection).cgColor
```
tigrim: Why not just use UIColor.customSeparatorSolid.cgColor here?

Sources/Controllers/PlanRoute/OAPlanRouteEditingBridge.mm
```diff
@@ -898,17 +914,34 @@ - (PlanRouteSegmentData *)buildSegmentWithIndex:(NSInteger)segmentIndex
     }
     BOOL routed = routedCount > 0;
     BOOL multiMode = groups.count > 1;
+    BOOL hasGapAfter = allPoints[segmentEndIndex].isGap && segmentEndIndex + 1 < (NSInteger)allPoints.count;
```
tigrim: Potential out-of-bounds access

Sources/Controllers/PlanRoute/OAPlanRouteEditingBridge.mm
```diff
@@ -2022,6 +2059,33 @@ - (void)onJoinPoints
...
+{
+    if (ctx == nil)
+        return NO;
+    BOOL hasRoutePair = wholeRoute
+        ? ctx.getPointsCount > 1
+        : pointIndex >= 0 && pointIndex < ctx.getPointsCount - 1 && !ctx.getPoints[pointIndex].isGap;
+    if (!hasRoutePair || mode == OAApplicationMode.DEFAULT)
+        return NO;
+    _isCalculatingRoute = YES;
+    if (self.onChange)
```
tigrim: Calling self.onChange() synchronously inside beginRouteCalculationIfNeededForContext: before executing the actual context commands (OAChangeRouteModeCommand, OAMovePointCommand) triggers UI updates while the underlying context is still in an intermediate state. Ensure onChange is dispatched after the context state transition completes or on the main queue to prevent potential race conditions with data readers

Sources/Controllers/PlanRoute/PlanRouteTopPartView.swift
```diff
@@ -196,12 +193,37 @@ final class PlanRouteTopPartView: UIView {
...
+        let secondary: [NSAttributedString.Key: Any] = [.font: monospacedDigitFont, .foregroundColor: UIColor.textColorSecondary]
+        var hours: Int32 = 0
+        var minutes: Int32 = 0
+        var seconds: Int32 = 0
+        OAUtilities.getHMS(duration, hours: &hours, minutes: &minutes, seconds: &seconds)
+
+        let result = NSMutableAttributedString()
+        if hours > 0 {
+            result.append(NSAttributedString(string: "\(hours)", attributes: primary))
+            result.append(NSAttributedString(string: " \(localizedString("int_hour"))", attributes: secondary))
```
tigrim: String concatenation like \(hours) \(localizedString("int_hour")) bypasses plural localization rules for languages with complex plural

Sources/Controllers/PlanRoute/PlanRouteTopPartView.swift
```diff
@@ -196,12 +193,37 @@ final class PlanRouteTopPartView: UIView {
...
+        }
+        if minutes > 0 {
+            if result.length > 0 {
+                result.append(NSAttributedString(string: " ", attributes: secondary))
+            }
+            result.append(NSAttributedString(string: "\(minutes)", attributes: primary))
+            result.append(NSAttributedString(string: " \(localizedString("shared_string_minute_lowercase"))", attributes: secondary))
```
tigrim: We already have duration formatting in OAOsmAndFormatter, and the task explicitly asks to follow Route Details formatting. Why do we introduce another manual getHMS-based implementation here? It would be better to reuse the existing formatter and apply attributed styling separately, so formatting/pluralization stays consistent across the app

Sources/Controllers/PlanRoute/Tabs/PlanRouteRouteViewController.swift
```diff
@@ -383,6 +404,8 @@ extension PlanRouteRouteViewController: UITableViewDelegate {
             return Self.profileGroupRowHeight
         case .point:
             return Self.pointRowHeight
+        case .gap:
+            return Self.gapRowHeight
```
tigrim, line 408: PlanRouteSegmentGapCell supports Dynamic Type, but the row height is fixed to 62. With larger accessibility text sizes the title can be clipped

Sources/Controllers/PlanRoute/OAPlanRouteEditingBridge.mm
tigrim: ?

tigrim (commented):
- [x] PlanRouteEditingContextDataProvider.swift, undo()/redo():

pendingEmptySegmentIndex is cleared on undo, but redo does not restore it. onNewSegmentStarted is only called from startNewSegment(), while redo() only triggers onChange, and updatePendingEmptySegment() cannot restore the state once it is nil. After Start new segment → Undo → Redo, the empty segment may disappear from the UI even though the split was restored.

- [x] OAPlanRouteEditingBridge.mm, beginRouteCalculationIfNeededForContext / applyMode:

onChange() is emitted before OAChangeRouteModeCommand updates the editing context, and when startsRouteCalculation == true there is no corresponding onChange() after the command. This can make the data provider rebuild its caches from the old context state and keep them stale until another routing callback arrives. Please publish the calculation state without invalidating route data before the command, or notify again after the context mutation.

- [ ] OAPlanRouteEditingBridge.mm, whole-route calculation:

For wholeRoute, hasRoutePair only checks getPointsCount > 1, while the single-segment case explicitly excludes gaps. Shouldn't the whole-route case verify that there is at least one actual routable pair? Otherwise _isCalculatingRoute may be enabled for a route that only contains disconnected segments.

tigrim (approved):

PR #5663, opened 2026-08-19
Fixed Dynamic Text for Gap cell

aleksandr-tata:
Fixed Dynamic Type support for the Plan Route screen.

Changes:
- Replaced fixed row heights with UITableView.automaticDimension.
- Added self-sizing constraints for point, profile group, segment header, gap, and “Start new segment” cells.
- Added minimum heights to preserve the Figma layout at the default text size:
  - Gap capsule: 52 pt
  - “Start new segment”: 50 pt
- Allowed labels to wrap and grow vertically at Accessibility text sizes.
- Updated Plan Route text styles according to Figma:
  - Body: 17 pt
  - Subheadline: 15 pt
  - Segment title: scalable 20 pt semibold
- Made capsule and header layouts expand with their content instead of clipping it.

The default layout remains aligned with Figma, while larger Dynamic Type sizes no longer clip the segment title or other text.

### DIFF
--- a/Sources/Controllers/PlanRoute/Cells/PlanRoutePointCell.swift
+++ b/Sources/Controllers/PlanRoute/Cells/PlanRoutePointCell.swift
@@ -10,6 +10,8 @@ import UIKit
 final class PlanRoutePointCell: UITableViewCell {
 
     private static let horizontalInset: CGFloat = 16
+    private static let minimumHeight: CGFloat = 68
+    private static let verticalInset: CGFloat = 12
     private static let circleSize: CGFloat = 28
     private static let deleteButtonSize: CGFloat = 30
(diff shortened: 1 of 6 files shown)


tigrim (commented):
PlanRouteRouteViewController now uses automatic row/header heights, but the table view does not invalidate already measured heights when Dynamic Type changes. Labels update their fonts and PlanRouteProfileGroupCell may switch its stack axis, yet visible/reused rows can keep cached heights until a full reload/navigation, so text may still clip after changing the content size category while this screen is open.

Handle preferredContentSizeCategory changes in the view controller and trigger a height recalculation, for example with performBatchUpdates(nil), beginUpdates/endUpdates, or reloadData()

tigrim (approved):

PR #5665, opened 2026-08-19
Different quantity favorite points after import

Vladyslav-lys:


tigrim:
Please also check for a performance regression when importing a large number of favorite points. The new merge logic does extra duplicate/index handling and may be slower on big GPX/backup imports, especially when many points are in the same folder or share the same name.

tigrim:
There seems to be a correctness risk in the new import merge flow. If the destination group already contains multiple favorites with the same name/category, and the imported point is equal to one of them, the code removes the equal existing favorite first. After that, because another favorite with the same name still remains in the list, the imported replacement can be skipped, which reduces the total favorites count after import. The replacement/skip decision should happen before mutating existing storage, or the matched item should be kept when the imported point is skipped.

tigrim (commented):
I think the new Update / Replace all flow can still create a count mismatch.

Example:
- Local favorites contain: "Cafe" / "Trip" at location A.
- Imported GPX contains two points with the same name/category:
  - "Cafe" / "Trip" at location B
  - "Cafe" / "Trip" at location C

If the user chooses Update for the first conflict, the code deletes the local favorite and immediately adds the imported favorite. Then it adds only the name ("Cafe") to `_ignoredNames` and calls `onRightNavbarButtonPressed` again, which runs the full `importFavoritesFromGpx(_gpxFile)`.

At that point validation skips all points named "Cafe", but the full import still processes the whole GPX. `checkDuplicateNames` can rename the duplicated imported points to "Cafe (1)" and "Cafe (2)", so `mergeImportedFavorites` no longer sees them as existing by name/category and can add them again. As a result, the point that was already added manually by Update may be imported a second time under a renamed duplicate name.

Could we avoid adding the imported favorite manually before the full import, or remove/mark the handled GPX point so it cannot be imported again by the subsequent full import? The same scenario should also be checked for Replace all with duplicated names

tigrim (approved):

PR #5666, opened 2026-08-20
Fix Plan Route undo/redo for empty segments and route profiles

aleksandr-tata:
## Summary

Fixes Undo/Redo state restoration for empty segments and route profiles in Plan Route.

## Changes

- Synchronize the pending empty segment with the actual trailing GPX gap after Undo/Redo.
- Restore the empty segment cell after redoing the `Start new segment` action.
- Remove the empty segment cell after undoing the split.
- Preserve and restore the original point profile when undoing a segment split.
- Preserve both old and new route profiles inside the route mode command.
- Apply profile changes to multiple points as a single command.
- Make grouped profile changes atomic: one Undo reverts the entire user action, and one Redo reapplies it.
- Restore the application mode and route profile after Undo/Redo.
- Update the selected profile and its icon in `Route between points` after history operations.
- Keep road segment data snapshots independent between Undo and Redo.

### DIFF
--- a/Sources/Controllers/PlanRoute/PlanRouteEditingContextDataProvider.swift
+++ b/Sources/Controllers/PlanRoute/PlanRouteEditingContextDataProvider.swift
@@ -167,6 +167,7 @@ final class PlanRouteEditingContextDataProvider: PlanRouteDataProvider {
     private var hasCachedAnalysisData = false
     private var analysisGeneration = 0
     private var pendingAnalysisGeneration: Int?
+    private var pendingSegmentHistoryState: (wasPending: Bool, hadTrailingGap: Bool)?
     private var initialApplicationMode: OAApplicationMode {
         var supportedModes = bridge.availableModes()
         if let directLineMode = OAApplicationMode.default() {
@@ -271,12 +272,15 @@ final class PlanRouteEditingContextDataProvider: PlanRouteDataProvider {
     }
 
     func undo() {
-        pendingEmptySegmentIndex = nil
+        pendingSegmentHistoryState = (pendingEmptySegmentIndex != nil, bridge.hasTrailingGap)
         bridge.undo()
+        pendingSegmentHistoryState = nil
     }
 
     func redo() {
+        pendingSegmentHistoryState = (pendingEmptySegmentIndex != nil, bridge.hasTrailingGap)
         bridge.redo()
+        pendingSegmentHistoryState = nil
     }
 
     func reverseRoute() {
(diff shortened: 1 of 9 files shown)


aleksandr-tata:
## Test scenarios

### Empty segment history

- [x] Add route points and select `Start new segment` — an empty segment cell is displayed.
- [x] Press Undo after starting a new segment — the trailing gap is removed and the empty segment cell disappears.
- [x] Press Redo — the trailing gap and the empty segment cell are restored.
- [x] Repeat Undo and Redo several times — the empty segment cell remains synchronized with the route state.
- [x] Press Undo when no history operation is available — the current empty segment state is not changed.

### Empty segment profile

- [x] Use `Straight line`, create an empty segment, and change its route type through `Route between points`.
- [x] Press Undo — the previous route type is restored.
- [x] Verify that the profile icon in `Route between points` is updated after Undo.
- [x] Press Redo — the selected route type and its icon are restored again.
- [x] Undo the original `Start new segment` action — the empty segment cell disappears.

### Existing route profile

- [x] Create a route with three points using `Straight line`.
- [x] Change the route type to `Driving` through `Route between points`.
- [x] Press Undo — the route type returns to `Straight line`.
- [x] Verify that the `Route between points` icon returns to the Straight line icon.
- [x] Press Redo — `Driving` and its corresponding icon are restored.

### Grouped profile changes

- [x] Change the profile for a complete profile group — all affected point pairs are updated.
- [x] Press Undo once — the complete profile change is reverted as one action.
- [x] Press Redo once — the complete profile change is reapplied.
- [x] Apply a profile starting from a selected point — only the expected range is changed.
- [x] Apply a profile up to a selected point — only the expected range is changed.
- [x] Undo and Redo range-based profile changes — the complete range is restored atomically.

tigrim (approved):

Sources/Controllers/RoutePlanning/Commands/OAChangeRouteModeCommand.mm
```diff
@@ -39,84 +49,116 @@ - (instancetype)initWithLayer:(OAMeasurementToolLayer *)measurementLayer appMode
...
     }
+    _newProfileTypes = [self profileTypesForPoints:_newPoints];
     [self executeCommand];
     return true;
 }
 
 - (void)undo
 {
     OAMeasurementEditingContext *editingCtx = [self getEditingCtx];
+    [self applyProfileTypes:_oldProfileTypes toPoints:_oldPoints];
```
alex-dev-neo, line 158: Undo during route calculation can leave loading stuck.
Undo now restores old point profiles immediately. If the new profile’s route calculation is still running, its result is rejected because the point profile no longer matches, but that rejection path does not clear `_currentPair` or hide progress. A quick Undo after selecting Driving/Bicycle therefore leaves isCalculatingRoute true indefinitely and suppresses analysis updates. Cancel the in-flight snap calculation before restoring profiles, or complete progress cleanup when rejecting stale results.

Sources/Controllers/PlanRoute/PlanRouteEditingContextDataProvider.swift
```diff
@@ -482,13 +486,15 @@ final class PlanRouteEditingContextDataProvider: PlanRouteDataProvider {
...
     private func updatePendingEmptySegment() {
-        guard let pendingEmptySegmentIndex else { return }
-        let segmentCount = bridgeSegments().count
-        if segmentCount == 0 || segmentCount >= pendingEmptySegmentIndex {
-            self.pendingEmptySegmentIndex = nil
+        let hasTrailingGap = bridge.hasTrailingGap
+        let shouldShowPendingSegment: Bool
+        if let pendingSegmentHistoryState {
+            shouldShowPendingSegment = hasTrailingGap
+                && (pendingSegmentHistoryState.wasPending || !pendingSegmentHistoryState.hadTrailingGap)
```
alex-dev-neo, line 493: Redo can invent a pending empty segment after deleting a point.
The condition treats every no-gap → trailing-gap history transition as restoring a pending segment. Reproduction: create a second segment with one point, delete that point, Undo, then Redo. The original deletion leaves `pendingEmptySegmentIndex` `nil`, while Redo creates an empty-segment cell. Pending state needs to be preserved per history action rather than inferred solely from the GPX gap transition.

Sources/Controllers/RoutePlanning/Commands/OAChangeRouteModeCommand.mm
aleksandr-tata: Fixed

PR #5675, opened 2026-08-20
Use system styling for Plan Route options buttons

aleksandr-tata:
## Summary

- Replaced the custom `blue_circle_fill` background with the standard tinted `UIButton.Configuration`.
- Updated the Plan Route options buttons to use the system capsule styling and accent colors.
- Removed the obsolete `blue_circle_fill` asset.
- Updated the day and night map ruler center icons


Resources/Images.xcassets/Icons/Map/map_ruler_center.imageset/map_ruler_center.svg
tigrim, line 1: Why do we need to keep two icons? We should be mindful of the app’s build size. There should be a single icon, with tintColor used to apply the appropriate color

aleksandr-tata: Fixed

tigrim (approved):

PR #5677, opened 2026-08-21
Fix Plan Route map controls and elevation graph navigation

aleksandr-tata:
## Summary

Fixes several Plan Route map UI issues related to RTL layout, full-screen mode, and elevation graph interaction.

## Changes

- Keep the route type button anchored to the physical left side in RTL layouts to prevent overlap with the zoom controls.
- Hide the route type, map HUD, and 3D buttons when the Plan Route bottom sheet is opened in full-screen mode.
- Recenter the map when a point selected on the elevation graph is outside the visible map viewport.

### DIFF
--- a/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
+++ b/Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
@@ -789,6 +789,7 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
 
     private func setState(_ state: EOADraggableMenuState, animated: Bool) {
         sheetState = state
+        routeTypeButton.isHidden = state == .fullScreen
         let height = height(for: state)
         sheetHeightConstraint?.constant = height
         crosshairCenterYConstraint?.constant = crosshairCenterY(sheetHeight: height)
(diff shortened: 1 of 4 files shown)


Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -80,6 +80,13 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
         return view.bounds.height
     }
 
+    var mapViewportBounds: CGRect {
```
tigrim, line 83: getViewHeight() returns mapControlsReservedHeight(for:), which is capped at the .expanded height. As a result, mapViewportBounds does not represent the actual visible map area when the sheet is in .fullScreen state. Since this rect is later used by fitTrackOnMap to decide whether the highlighted point is visible, we can skip recentering for a point that is actually covered by the sheet. Shouldn't the viewport use the actual current sheet height here?

aleksandr-tata: Fixed

tigrim (approved):

PR #5682, opened 2026-08-23
Car play map mode setting UI fix + migration fix

### LINKED ISSUE #5398: CarPlay Appearance Overrides OsmAnd Map Mode Setting

vitaliy-sova-ios:


Sources/Helpers/OADayNightHelper.m
```diff
@@ -113,6 +121,45 @@ - (BOOL)isNightMode
...
+            if (daynightSwitch)
+            {
+                _recalcIntervalCarPlay = 60.0;
+                BOOL daytime = [daynightSwitch isDaytime];
+                nightMode = !daytime;
+            }
+        }
+    }
+
+    if (_lastNightModeCarPlay != nightMode)
```
tigrim, line 157: The observable is triggered only when `_lastNightModeCarPlay` changes, but switching the rendering context itself may also require a refresh. For example, if the phone map is in Night mode and CarPlay is configured for Day, the initial CarPlay value is `false`, which matches the default `_lastNightModeCarPlay` value, so no notification is sent and the existing night map source may remain active. The reverse can happen after disconnecting because removing `resetCarPlayMode()` also removes the refresh that used to restore the phone appearance. Please explicitly reload the map whenever the rendering context switches between phone and CarPlay.

Sources/Helpers/OADayNightHelper.m
```diff
@@ -113,6 +121,45 @@ - (BOOL)isNightMode
...
+    BOOL nightMode = _lastNightModeCarPlay;
+    if (dayNightMode == DayNightModeDay)
+        nightMode = NO;
+    else if (dayNightMode == DayNightModeNight)
+        nightMode = YES;
+    else if (dayNightMode == DayNightModeAuto)
+    {
+        NSTimeInterval currentTime = CACurrentMediaTime();
+        
+        if (currentTime - _lastTimeCarPlay > _recalcIntervalCarPlay)
```
tigrim, line 144: `_recalcIntervalCarPlay` only limits the recalculation frequency; it does not schedule another calculation. Unlike the regular `nightMode` getter, which is read frequently by map layers, widgets, and HUD components, `nightModeCarPlay` is only read from `updateCurrentMapSource`. As a result, Auto mode may not switch at sunrise or sunset until an unrelated event rebuilds the map source. Could we provide a reliable periodic trigger for the CarPlay calculation?

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -2615,7 +2631,10 @@ - (void) updateCurrentMapSource
             NSString *baseMode = am.parent && am.parent.stringKey.length > 0 ? am.parent.stringKey : am.stringKey;
             newSettings[@"baseAppMode"] = baseMode;
 
-            if (settings.nightMode)
+            BOOL mapNightMode = UIApplication.sharedApplication.isAnyCarPlaySceneActive
```
tigrim: The CarPlay-specific mode is applied here and to sky/fog settings, but POI, route, and other map layers still use the regular `settings.nightMode`, and they do not observe `carPlayDayNightModeObservable`. When the phone and CarPlay modes differ, this can produce a dark base map with daytime symbols or route colors. Please make all map-specific consumers use the same resolved appearance mode.

Sources/Helpers/MigrationManager.swift
```diff
@@ -617,9 +617,14 @@ final class MigrationManager: NSObject {
     }
     
     private func migrateCarPlayMapAppearanceMode() {
-        let carPlayMode = settings.carPlayMode.get()
-        let profileMapMode = settings.appearanceMode.get(carPlayMode)
-        settings.carPlayMapAppearanceMode.set(profileMapMode)
+        let mode: OAApplicationMode
+        if settings.isCarPlayModeDefault.get() {
```
tigrim: This resolution differs from `CarPlayNavigationModeProvider.configureForCarPlay()`: when the default CarPlay mode is enabled, the provider preserves the current application mode if it is car-derived, while this migration always selects `firstCarMode()`. Users with an active custom or derived car profile may therefore receive the appearance setting from the base Car profile. Please reuse the same profile-resolution logic here.

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -2241,7 +2246,7 @@ - (float) currentZoomOutDelta
     return 0.0f;
 }
 
-- (void) onDayNightModeChanged
+- (void)onDayNightModeChangedAction
```
tigrim: The `Action` suffix is a bit misleading here because this method is not an action/selector entry point anymore; it is the shared implementation that reloads the map source after either the regular or CarPlay day/night observable fires. Could we name it after the work it performs, for example `reloadMapSourceForDayNightModeChange` or `handleResolvedDayNightModeChanged`?

Sources/Helpers/OAAppSettings.m
```diff
@@ -7850,6 +7850,13 @@ - (BOOL) nightMode
     return [_dayNightHelper isNightMode];
 }
 
+- (BOOL) mapNightMode
```
tigrim: `mapNightMode` can be misleading here because the same map can be presented by different display contexts, and each context has its own appearance state. This is not really a single generic map value. Could we use explicit display-specific names, such as `appMapNightMode` and `carPlayMapNightMode`, or expose something like `mapNightModeForDisplay:` instead?

Sources/Helpers/OADayNightHelper.m
```diff
@@ -177,6 +180,47 @@ - (BOOL)resetTempMode
...
+    else
+        [self stopCarPlayAutoRecalcTimer];
+}
+
+- (void)startCarPlayAutoRecalcTimer
+{
+    executeOnMainThread(^{
+        if (self->_carPlayAutoRecalcTimer)
+            return;
+        __weak OADayNightHelper *weakSelf = self;
```
tigrim: __weak __typeof(self) weakSelf = self;

Sources/AppHost/CarPlayDelegate/CarPlayService.swift
```diff
@@ -82,6 +85,12 @@ final class CarPlayService: NSObject {
             }
         }
     }
+
+    private func reloadMapForRenderingContextChange() {
+        executeOnMainThread {
```
tigrim: Are these calls made off the main thread?

Sources/Helpers/OAAppSettings.h
```diff
@@ -990,7 +990,7 @@ typedef NS_ENUM(NSInteger, EOAWikiDataSourceType)
 @property (nonatomic) OACommonInteger *appearanceProfileTheme; // 0 - System; 1 - Light; 2 - Dark
 @property (nonatomic) OACommonDouble *mapManuallyRotatingAngle;
 @property (readonly, nonatomic) BOOL nightMode;
-@property (readonly, nonatomic) BOOL nightModeCarPlay;
+@property (readonly, nonatomic) BOOL mapNightMode;
```
tigrim: isMapNightMode

Sources/Controllers/Map/Helpers/POIImageLoader/POIImageLoader.swift
```diff
@@ -156,11 +156,12 @@ final class POIImageLoader: NSObject, @unchecked Sendable {
     }
     
     private func makeIconProcessor(metrics: IconMetrics) -> ImageProcessor {
+        let color: UIColor = OAAppSettings.sharedManager().mapNightMode ? .popularPlaceBgDefault.dark : .popularPlaceBgDefault.light
```
tigrim: Why not just use isMapNightMode directly in currentMapThemeColor?

Sources/Controllers/Map/OAMapViewController.mm
tigrim: reloadMapSourceForDayNightModeChange

Sources/Helpers/OAAppSettings.h
vitaliy-sova-ios: renamed to activeMapNightMode

Sources/Controllers/Map/Helpers/POIImageLoader/POIImageLoader.swift
vitaliy-sova-ios: It’s already used by map HUD/widgets (OATextInfoWidget, OAMapHudViewController, OATurnDrawable, OALanesDrawable, OAStreetNameWidgetParams), which should follow the phone day/night mode.

Sources/Helpers/OADayNightHelper.m
```diff
@@ -129,15 +179,51 @@ - (BOOL)resetTempMode
...
+        self->_carPlayAutoRecalcTimer = [NSTimer scheduledTimerWithTimeInterval:kCarPlayAutoRecalcInterval
```
tigrim: Since this timer is responsible for keeping CarPlay Auto appearance up to date, should we add it to `NSRunLoopCommonModes` instead of relying on the default run-loop mode? Otherwise the timer can be paused during UI tracking/interactions and delay the day/night switch.

Sources/Controllers/Map/Helpers/POIImageLoader/POIImageLoader.swift
```diff
@@ -156,11 +156,12 @@ final class POIImageLoader: NSObject, @unchecked Sendable {
     }
     
     private func makeIconProcessor(metrics: IconMetrics) -> ImageProcessor {
+        let color: UIColor = OAAppSettings.sharedManager().activeMapNightMode ? .popularPlaceBgDefault.dark : .popularPlaceBgDefault.light
```
tigrim: add extension

Sources/Controllers/Map/Helpers/POIImageLoader/POIImageLoader.swift
```diff
@@ -173,23 +174,29 @@ final class POIImageLoader: NSObject, @unchecked Sendable {
     
     private func placeholderCacheKey(placeholderImageName: String,
                                      metrics: IconMetrics) -> String? {
+        let color: UIColor = OAAppSettings.sharedManager().activeMapNightMode ? .popularPlacePlaceholderBg.dark : .popularPlacePlaceholderBg.light
```
tigrim: use new extension

Sources/Controllers/Map/Helpers/POIImageLoader/POIImageLoader.swift
```diff
@@ -173,23 +174,29 @@ final class POIImageLoader: NSObject, @unchecked Sendable {
...
             "poi_placeholder",
             placeholderImageName,
             "scale_\(metrics.textScale)",
-            "icon_\(UIColor.popularPlacePlaceholderBg.currentMapThemeColor.hashValue)"
+            "icon_\(color.hashValue)"
         ].joined(separator: "_")
     }
     
     func createProcessedPlaceholder(with image: UIImage, metrics: IconMetrics, option: KingfisherParsedOptionsInfo) -> UIImage? {
+        let nightMode = OAAppSettings.sharedManager().activeMapNightMode
```
tigrim: up

Sources/Helpers/OAAppSettings.m
tigrim: ?

Sources/SwiftExtensions/UIColor+Extension.swift
```diff
@@ -11,8 +11,8 @@ import UIKit
 extension UIColor {
     @objc var dark: UIColor { resolvedColor(with: .init(userInterfaceStyle: .dark)) }
     @objc var light: UIColor { resolvedColor(with: .init(userInterfaceStyle: .light)) }
-    @objc var currentMapThemeColor: UIColor { OAAppSettings.sharedManager().nightMode ? dark : light }
+    @objc var currentMapThemeColor: UIColor { OAAppSettings.sharedManager().isAppMapNightMode ? dark : light }
     @objc var activeMapThemeColor: UIColor {
```
tigrim: activeMapThemeColor -> currentMapThemeColor

Sources/SwiftExtensions/UIColor+Extension.swift
```diff
@@ -11,8 +11,8 @@ import UIKit
 extension UIColor {
     @objc var dark: UIColor { resolvedColor(with: .init(userInterfaceStyle: .dark)) }
     @objc var light: UIColor { resolvedColor(with: .init(userInterfaceStyle: .light)) }
-    @objc var currentMapThemeColor: UIColor { OAAppSettings.sharedManager().nightMode ? dark : light }
+    @objc var currentMapThemeColor: UIColor { OAAppSettings.sharedManager().isAppMapNightMode ? dark : light }
```
tigrim: currentMapThemeColor -> appMapThemeColor

Sources/Helpers/OADayNightHelper.m
```diff
@@ -129,15 +179,53 @@ - (BOOL)resetTempMode
...
+- (void)startCarPlayAutoRecalcTimer
+{
+    executeOnMainThread(^{
+        if (self->_carPlayAutoRecalcTimer)
+            return;
+        __weak __typeof(self) weakSelf = self;
+        NSTimer *timer = [NSTimer timerWithTimeInterval:kCarPlayAutoRecalcInterval
+                                               repeats:YES
+                                                 block:^(NSTimer * _Nonnull timer) {
+            [weakSelf forceUpdateCarPlay];
```
tigrim:
```
   if (!strongSelf)
 {
         [timer invalidate];
          return;
            }
```

Sources/Helpers/OADayNightHelper.m
```diff
@@ -208,9 +208,15 @@ - (void)startCarPlayAutoRecalcTimer
             return;
         __weak __typeof(self) weakSelf = self;
         NSTimer *timer = [NSTimer timerWithTimeInterval:kCarPlayAutoRecalcInterval
-                                               repeats:YES
-                                                 block:^(NSTimer * _Nonnull timer) {
-            [weakSelf forceUpdateCarPlay];
+                                                repeats:YES
+                                                  block:^(NSTimer * _Nonnull timer) {
+            __strong __typeof(weakSelf) strongSelf = weakSelf;
+            if (!strongSelf) {
```
tigrim: formatting

tigrim (approved):

PR #5684, opened 2026-08-24
Follow up my places redesign

### LINKED ISSUE #5612: Follow up for My Places redesign

DmitrySvetlichny:


Sources/Controllers/Settings/OsmandDevelopment/TouchIndicatorController.swift
```diff
@@ -101,20 +101,46 @@ final class TouchIndicatorController: NSObject, UIGestureRecognizerDelegate {
 
         weak var overlay: OverlayWindow?
 
+        private var activeTouches: Set<ObjectIdentifier> = []
```
tigrim, line 104: What problem were we trying to solve in this class?

Sources/Controllers/Cells/PointRowContentView.swift
```diff
@@ -144,6 +145,13 @@ private final class PointRowContentView: UIView, UIContentView {
         fatalError("init(coder:) has not been implemented")
     }
 
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+        guard window != nil, separatorConstraint == nil, let cell = sequence(first: self as UIResponder, next: \.next).first(where: { $0 is UICollectionViewListCell }) as? UICollectionViewListCell, let textLayoutGuide = rowContentView.textLayoutGuide else { return }
```
tigrim: ?

tigrim:
Found one issue in the new search dismissal flow:

`MyPlacesContainerViewController.hideSearchController()` sets `navigationItem.searchController?.isActive = false` and immediately removes `navigationItem.searchController = nil`. The restoration logic for the segmented control/top inset now lives in `willDismissSearchController` / `didDismissSearchController`, so removing the search controller synchronously can bypass or race those delegate callbacks. In that case My Places can remain with the segment container hidden, stale `additionalSafeAreaInsets.top`, or `isSearchBarAnimating`/`isSearchActive` left in the wrong state.

I think this should only deactivate the controller here, and move `navigationItem.searchController = nil` to `didDismissSearchController` after UIKit finishes dismissal.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+Actions.swift
```diff
@@ -85,14 +91,17 @@ extension FavoriteListViewController {
         }
 
         collectionView.isEditing = isEditing
-        collectionView.reloadData()
```
Vladyslav-lys, line 88: Why have you removed this? Do items update correctly on the list?

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController.swift
```diff
@@ -207,14 +212,7 @@ final class FavoriteListViewController: UIViewController, MyPlacesScrollResettab
...
         navigationController?.setNavigationBarHidden(false, animated: false)
-        if !isRootFolder {
-            let appearance = UINavigationBarAppearance()
-            appearance.configureWithDefaultBackground()
-            navigationController?.navigationBar.standardAppearance = appearance
-            navigationController?.navigationBar.scrollEdgeAppearance = appearance
-            navigationController?.navigationBar.tintColor = .iconColorActive
-        }
-
+        navigationController?.setDefaultNavigationBarAppearance()
```
Vladyslav-lys, line 215: Do non-root screen navbar have half-transparent background?

Sources/Controllers/Settings/OsmandDevelopment/TouchIndicatorController.swift
DmitrySvetlichny: This fixes scrolling when Show touches is enabled. activeTouches keeps the recognizer active until the last finger is released, while canPrevent(false) ensures it doesn’t block scroll gestures.

Sources/Controllers/Cells/PointRowContentView.swift
DmitrySvetlichny: This aligns the separator with the primary text of the nested UIListContentView. The guard waits until the content view is attached to its list cell, avoids creating the constraint more than once, and safely handles the optional textLayoutGuide.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+Actions.swift
DmitrySvetlichny: reloadData() is no longer needed because editing-dependent accessories now use their displayed state and are updated automatically when collectionView.isEditing changes. The list data itself does not change here; when it does, we apply a new snapshot explicitly. This also avoids an unnecessary full reload and preserves filtered results when entering selection mode from search.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController.swift
DmitrySvetlichny: Yes, non-root screens keep the system semi-transparent/blurred navigation bar appearance. I verified it on nested folders.

Vladyslav-lys (approved):

Sources/Controllers/Cells/PointRowContentView.swift
tigrim: This works, but it feels fairly fragile: `PointRowContentView` reaches up the responder chain in `didMoveToWindow` to find its containing `UICollectionViewListCell`, then constrains the cell separator to `UIListContentView.textLayoutGuide`. That couples the content view to the current UIKit cell hierarchy and may break if the view is reused in another cell type or UIKit changes the responder/layout structure. See: Astronomy

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+Actions.swift
```diff
@@ -284,16 +296,15 @@ extension FavoriteListViewController {
     }
 
     @objc func searchSelectButtonPressed() {
-        isSelectionModeInSearch = true
-        isSearchActive = false
-        hideSearchController()
-
+        isCancellingSearch = true
         selectButtonPressed()
+        DispatchQueue.main.async {
```
tigrim: ?

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+Delegates.swift
```diff
@@ -132,6 +146,9 @@ extension FavoriteListViewController: MyPlacesSearchable, UISearchResultsUpdatin
         configureToolbar()
         navigationController?.setToolbarHidden(!collectionView.isEditing, animated: true)
         applySnapshot(animatingDifferences: false)
+        DispatchQueue.main.async {
```
tigrim: ?

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+Models.swift
```diff
@@ -185,10 +185,17 @@ struct FavoriteFolderStats: Hashable {
...
     private static let rowHeight: CGFloat = 68.0
+    private var separatorConstraint: NSLayoutConstraint?
 
     override func preferredLayoutAttributesFitting(_ layoutAttributes: UICollectionViewLayoutAttributes) -> UICollectionViewLayoutAttributes {
         let attributes = super.preferredLayoutAttributesFitting(layoutAttributes)
         attributes.frame.size.height = ceil(max(Self.rowHeight, attributes.frame.height))
         return attributes
     }
+    
+    func alignSeparatorToPrimaryText() {
```
tigrim: ?

Sources/Controllers/MyPlaces/MyPlacesContainerViewController.swift
```diff
@@ -199,12 +215,31 @@ final class MyPlacesContainerViewController: OACompoundViewController {
         searchController?.searchBar.delegate = self
         searchController?.delegate = self
         searchController?.obscuresBackgroundDuringPresentation = false
+        searchController?.hidesNavigationBarDuringPresentation = true
         definesPresentationContext = true
-        if #available(iOS 26.0, *) {
+        navigationItem.hidesSearchBarWhenScrolling = false
+        navigationItem.largeTitleDisplayMode = .never
+        if #available(iOS 16.0, *) {
```
tigrim: What is this? .stacked is not available on iOS 16.0. This will crash!

Sources/Controllers/MyPlaces/MyPlacesContainerViewController.swift
```diff
@@ -375,36 +412,53 @@ extension MyPlacesContainerViewController: MyPlacesDelegate {
...
-                self.view.layoutIfNeeded()
-            }
+            hideSearchController()
         }
     }
     
     func updateContentScrollView(_ scrollView: UIScrollView) {
         setContentScrollView(scrollView, for: .top)
+        if #available(iOS 26.0, *) {
+            segmentContainerView.backgroundColor = .clear
```
tigrim, line 444: ?

Sources/Controllers/MyPlaces/MyPlacesContainerViewController.swift
tigrim: Maybe not. We need to investigate why different approaches are used in different parts of the code

Sources/Controllers/MyPlaces/MyPlacesContainerViewController.swift
DmitrySvetlichny: This connects the active tab’s scroll view to the segmented-control container, allowing iOS 26 to render its native top scroll-edge/glass effect while content scrolls underneath it. UIScrollEdgeElementContainerInteraction was introduced in iOS 26, hence the availability check. Earlier iOS versions use an opaque background and the existing navigation-bar appearance as a fallback.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+Models.swift
```diff
@@ -185,6 +185,14 @@ struct FavoriteFolderStats: Hashable {
 
 final class FavoriteListCell: UICollectionViewListCell {
     private static let rowHeight: CGFloat = 68.0
+    private var separatorConstraint: NSLayoutConstraint?
+
+    override func layoutSubviews() {
+        super.layoutSubviews()
+        guard separatorConstraint == nil, let textLayoutGuide = (contentView as? PrimaryTextLayoutGuideProviding)?.primaryTextLayoutGuide else { return }
```
tigrim: СontentView is the cell's container UIView, while PrimaryTextLayoutGuideProviding is implemented by PointRowContentView inside it. This cast will therefore fail and the custom separator alignment won't be applied. Could we keep a direct reference to the configured content view/layout guide instead?

tigrim (approved):

PR #5688, opened 2026-08-24
Auto-merge independent Favorites changes across devices

### LINKED ISSUE #5177: Merge favorites edited on two devices into one folder via Cloud
### Description

Initially reported here: https://github.com/osmandapp/OsmAnd/issues/21497#issuecomment-3017952326 Maybe related: #3347

Favorites that are updated locally and on another device or the web cannot be synced at all

The UI only gives the option to discard either the local data or the cloud data.

### Steps to reproduce

* Create favorites
* Add new favorites on two different devices

### Actual result

conflicts

### Expected result

no conflicts, successful sync

```
OsmAnd Version: 5228 (?)
iOS version: 18.7.4
Device model: Xs
```

alex-dev-neo:
Fixes #5177.

### Summary

Adds three-way merging for Favorites groups changed on multiple devices.

The merge uses a locally stored last-synced GPX snapshot as the base.

Merge preparation creates only an in-memory upload payload. Local Favorites are changed only after the merged file is uploaded successfully.

### Behavior

Automatically merges:

- Additions with different names.
- Edits, moves, deletions, and renames made to different points.
- The same point deleted on both devices.
- A one-sided rename when the other device did not change that point.

Keeps a conflict when:

- Both devices change the same point.
- The same point is renamed differently.
- One device renames a point while the other deletes it.
- Both devices add points with the same name.
- Group appearance is changed.
- A cloud-deleted group was also changed locally.
- The group is personal.
- Point names are empty or duplicated.

If no matching last-synced snapshot exists, the existing conflict behavior is preserved.

### Group deletion behavior

- If a group was deleted in the cloud and is unchanged locally, it is deleted locally.
- If the cloud-deleted group was changed locally, it remains a conflict.
- A cloud deletion marker is not treated as a missing remote file, so the group is not incorrectly shown as a new local upload.

### Implementation

The merge logic is isolated in `OAFavoritesBackupMerger`.

Small integration hooks were added for:

- Preparing a merged upload.
- Saving the snapshot after a successful upload.
- Saving the snapshot after a successful download.

`OANetworkWriter` and the Favorites data model are unchanged.

### Test plan

- Add different points to the same group on two devices and verify both are preserved.
- Edit, move, delete, or rename different points and verify automatic merging.
- Delete the same point on both devices and verify no conflict.
- Rename the same point differently on two devices and verify a conflict.
- Test rename versus deletion and verify a conflict.
- Add the same point name on both devices and verify a conflict.
- Change group appearance and verify a conflict.
- Delete a group in the cloud and verify an unchanged local copy is deleted.
- Delete a group in the cloud, modify it locally, and verify a conflict.
- Verify a cloud-deleted group is not shown as a new local upload.
- Verify pull-to-refresh/preparation does not modify local Favorites.

### DIFF
--- a/Sources/Backup/OAFavoritesBackupMerger.h
+++ b/Sources/Backup/OAFavoritesBackupMerger.h
@@ -5,6 +5,8 @@
 
 #import <Foundation/Foundation.h>
 
+NS_ASSUME_NONNULL_BEGIN
+
 @class OABackupHelper, OABackupInfo, OARemoteFile, OASettingsItem;
 
 @interface OAFavoritesBackupMerger : NSObject
@@ -14,3 +16,5 @@
 + (void)onDownloadSuccess:(OASettingsItem *)item remoteFile:(OARemoteFile *)remoteFile;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/Sources/Backup/OAFavoritesBackupMerger.h
+++ b/Sources/Backup/OAFavoritesBackupMerger.h
@@ -1,0 +1,16 @@
+//
+//  OAFavoritesBackupMerger.h
+//  OsmAnd Maps
+//
+
+#import <Foundation/Foundation.h>
+
+@class OABackupHelper, OABackupInfo, OARemoteFile, OASettingsItem;
+
+@interface OAFavoritesBackupMerger : NSObject
+
++ (void)prepareMergeUploads:(OABackupInfo *)info backupHelper:(OABackupHelper *)backupHelper;
++ (void)onUploadSuccess:(OASettingsItem *)item fileName:(NSString *)fileName uploadTime:(long)uploadTime;
++ (void)onDownloadSuccess:(OASettingsItem *)item remoteFile:(OARemoteFile *)remoteFile;
+
+@end
--- a/Sources/Backup/OAFavoritesBackupMerger.mm
+++ b/Sources/Backup/OAFavoritesBackupMerger.mm
@@ -20,31 +20,6 @@
 
 static NSString * const kFavoritesSnapshotDirectory = @"favorites_sync";
 
-@interface OAFavoritesBackupMerger ()
-
-+ (OAFavoriteGroup *)downloadGroup:(OARemoteFile *)remoteFile backupHelper:(OABackupHelper *)backupHelper;
-+ (OAFavoriteGroup *)loadSnapshot:(NSString *)fileName syncTime:(long)syncTime;
-+ (void)saveSnapshot:(OAFavoriteGroup *)group fileName:(NSString *)fileName syncTime:(long)syncTime;
-+ (NSString *)snapshotRoot;
-+ (NSString *)snapshotPath:(NSString *)fileName syncTime:(long)syncTime;
-+ (OAFavoriteGroup *)readGroup:(NSString *)path;
-+ (OAFavoriteGroup *)mergeBase:(OAFavoriteGroup *)base local:(OAFavoriteGroup *)local remote:(OAFavoriteGroup *)remote;
-+ (NSMutableDictionary<NSString *, OAFavoriteItem *> *)pointsByName:(OAFavoriteGroup *)group;
-+ (BOOL)sameAppearance:(OAFavoriteGroup *)first other:(OAFavoriteGroup *)second;
-+ (BOOL)sameGroup:(OAFavoriteGroup *)first other:(OAFavoriteGroup *)second;
-+ (BOOL)samePoint:(OAFavoriteItem *)first other:(OAFavoriteItem *)second;
-+ (BOOL)samePointContent:(OAFavoriteItem *)first other:(OAFavoriteItem *)second;
-+ (BOOL)hasRenameOf:(OAFavoriteItem *)basePoint
-                 in:(NSDictionary<NSString *, OAFavoriteItem *> *)additions;
-+ (BOOL)sameText:(NSString *)first other:(NSString *)second;
-+ (BOOL)sameTime:(NSDate *)first other:(NSDate *)second;
-+ (NSString *)normalizedIcon:(NSString *)icon;
-+ (NSString *)normalizedBackground:(NSString *)background;
-+ (OAFavoriteItem *)copyPoint:(OAFavoriteItem *)point;
-+ (OAFavoriteGroup *)copyGroup:(OAFavoriteGroup *)group;
-
-@end
-
 @interface OAMergedFavoritesSettingsItem : OAFavoritesSettingsItem
 
 - (instancetype)initWithBaseItem:(OAFavoritesSettingsItem *)baseItem
--- a/Sources/Backup/OAFavoritesBackupMerger.mm
+++ b/Sources/Backup/OAFavoritesBackupMerger.mm
@@ -33,6 +33,9 @@ static NSString * const kFavoritesSnapshotDirectory = @"favorites_sync";
 + (BOOL)sameAppearance:(OAFavoriteGroup *)first other:(OAFavoriteGroup *)second;
 + (BOOL)sameGroup:(OAFavoriteGroup *)first other:(OAFavoriteGroup *)second;
 + (BOOL)samePoint:(OAFavoriteItem *)first other:(OAFavoriteItem *)second;
++ (BOOL)samePointContent:(OAFavoriteItem *)first other:(OAFavoriteItem *)second;
++ (BOOL)hasRenameOf:(OAFavoriteItem *)basePoint
+                 in:(NSDictionary<NSString *, OAFavoriteItem *> *)additions;
 + (BOOL)sameText:(NSString *)first other:(NSString *)second;
 + (BOOL)sameTime:(NSDate *)first other:(NSDate *)second;
 + (NSString *)normalizedIcon:(NSString *)icon;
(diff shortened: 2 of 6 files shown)


Sources/Backup/OAFavoritesBackupMerger.h
```diff
@@ -0,0 +1,16 @@
+//
+//  OAFavoritesBackupMerger.h
+//  OsmAnd Maps
+//
+
+#import <Foundation/Foundation.h>
+
+@class OABackupHelper, OABackupInfo, OARemoteFile, OASettingsItem;
+
```
tigrim, line 11: NS_ASSUME_NONNULL_BEGIN/NS_ASSUME_NONNULL_END

Sources/Backup/OAFavoritesBackupMerger.mm
```diff
@@ -0,0 +1,467 @@
...
+#import "OARemoteFile.h"
+#import "OAUtilities.h"
+#import "OsmAndApp.h"
+#import "OsmAnd_Maps-Swift.h"
+
+static NSString * const kFavoritesSnapshotDirectory = @"favorites_sync";
+
+@interface OAFavoritesBackupMerger ()
+
++ (OAFavoriteGroup *)downloadGroup:(OARemoteFile *)remoteFile backupHelper:(OABackupHelper *)backupHelper;
```
tigrim: These private method declarations are not strictly required in the .mm file

Sources/Backup/OAFavoritesBackupMerger.mm
```diff
@@ -0,0 +1,442 @@
...
+
+    if ([OAFavoritesBackupMerger sameGroup:_mergedGroup other:current])
+    {
+        self.localModifiedTime = _sourceModifiedTime;
+        [OAFavoritesBackupMerger saveSnapshot:_mergedGroup fileName:fileName syncTime:uploadTime];
+    }
+    else
+    {
+        if (current)
+            self.localModifiedTime = MAX((long)(NSDate.date.timeIntervalSince1970 * 1000.0), uploadTime + 1);
```
alex-dev-neo, line 438: Post-merge review note. Same shape as the Android counterpart (osmandapp/OsmAnd#25721, `FavoritesBackupMerger.java:394`).

When the group changes between merge preparation and upload completion, this branch runs neither `apply` nor `saveSnapshot:`, yet the merged payload is already in the cloud — and it still marks the local file newer than `uploadTime`.

Scenario: devices A and B both edited `Trip`. `prepareMergeUploads:` builds merged `M` (carrying B's remote change `R`) and uploads it; `uploadTime = T`. The user edits `Trip` while the upload is in flight, so the live group is `L'`. Both `sameGroup:_mergedGroup other:current` and `sameGroup:_localGroup other:current` are then NO, so `apply` is skipped and we land here. On the next `PrepareBackup`: `localModifiedTime > T` while the remote `updatetimems == T`, so only `fileChangedLocally` is true — `OAGenerateBackupInfoTask` puts the file in `filesToUpload` and `L'` overwrites `M` in the cloud. B's change `R` is gone, and since no snapshot was written the merger cannot recover it later either.

Leaving the pair as a conflict here, rather than marking the local file dirty, would stop the un-merged state from winning by default. (The `if (current)` guard is a good addition over the Android version, but it does not cover this case.)

Sources/Backup/OAFavoritesBackupMerger.mm
```diff
@@ -0,0 +1,442 @@
...
+{
+    return icon.length > 0 ? icon : DEFAULT_ICON_NAME_KEY;
+}
+
++ (NSString *)normalizedBackground:(NSString *)background
+{
+    return background.length > 0 ? background : DEFAULT_ICON_SHAPE_KEY;
+}
+
++ (OAFavoriteItem *)copyPoint:(OAFavoriteItem *)point
```
alex-dev-neo, line 369: Post-merge review note — this one is iOS-only; the Android port is not affected.

`copyFavoriteLocation` (core/src/FavoriteLocationsCollection.cpp:59) forwards only position, elevation, time, pickupTime, title, description, address, group, icon, background, color, extensions, calendarEvent and amenityOriginName. `isHidden` is a separate field on `IFavoriteLocation` (IFavoriteLocation.h:40), is not among them, and `FavoriteLocation_P` initialises it to `false`. Since `OAFavoriteItem.isVisible` is `!favorite->isHidden()`, **every point returned by `copyPoint:` is visible**, whatever the source point was.

`samePointContent:other:` compares `first.isVisible != second.isVisible`, and hiding a group hides all of its points (OAFavoritesHelper.mm:821). So for a hidden group: `saveSnapshot:` copies the group before writing, the snapshot GPX records every point as visible, and on the next conflict both `samePoint:base other:local` and `samePoint:base other:remote` are NO — `mergeBase:local:remote:` returns nil and the conflict is kept. Hidden Favorites groups can therefore never auto-merge, deterministically and with no diagnostic.

Android is unaffected because its `copyPoint` uses `new FavouritePoint(point)`, whose copy constructor carries `visible`.

The same call also drops `specialPointType`, which Android's `copyPoint` restores explicitly. Low impact here, since special points live in the personal group and `prepareMergeUploads:` skips personal groups.

A `[copy setVisible:point.isVisible]` after the copy would fix the visibility half.

Sources/Backup/OAFavoritesBackupMerger.mm
```diff
@@ -0,0 +1,442 @@
...
+
+- (void)finishUpload:(NSString *)fileName uploadTime:(long)uploadTime
+{
+    OAFavoriteGroup *current = [OAFavoritesHelper groupByName:_localGroup.name];
+    if (![OAFavoritesBackupMerger sameGroup:_mergedGroup other:current] &&
+        [OAFavoritesBackupMerger sameGroup:_localGroup other:current])
+    {
+        self.shouldReplace = YES;
+        [self processDuplicateItems];
+        [self apply];
```
alex-dev-neo, line 426: Post-merge review note. Corresponds to the Android finding at `FavoritesBackupMerger.java:387`, with a wider concurrency window but partly mitigated here.

`OABackupExporter` runs light tasks at `maxConcurrentOperationCount = 10` (OABackupExporter.m:166; Android uses 4), so up to ten `onItemFileUploadDone` -> `finishUpload:` -> `apply` calls can overlap when several groups auto-merge in one sync.

`+[OAFavoritesSettingsItem applyItems:]` does wrap `addFavoriteGroups` / `sortAll` / `saveCurrentPointsIntoFile` / `loadFavorites` in `@synchronized (self)`, which is a real improvement over the Android version. But `prepareItemsToApply` calls `[OAFavoritesHelper deleteFavorites:… saveImmediately:NO]` outside that block, so the delete-then-add for one group is not atomic with respect to another group's save + reload: thread B's `saveCurrentPointsIntoFile` can persist group X with its existing points already deleted and its merged points not yet added.

Extending the synchronised region to cover `prepareItemsToApply`, or applying merged items once after `writeItems` returns instead of per-upload-callback, would close this.

Sources/Backup/OAGenerateBackupInfoTask.m
```diff
@@ -162,6 +168,7 @@ - (OABackupInfo *) doInBackground
                 [info.filesToUpload addObject:localFile];
         }
     }
+    [OAFavoritesBackupMerger prepareMergeUploads:info backupHelper:OABackupHelper.sharedInstance];
```
alex-dev-neo, line 171: Post-merge review note. Same as the Android finding at `GenerateBackupInfoTask.java:132`. Two separable concerns:

1. `prepareMergeUploads:` performs a synchronous `[backupHelper downloadFile:remoteFile:listener:]` per conflicting group (OABackupHelper.mm:503) from inside `doInBackground`. This task runs on every backup-info generation, including auto-sync preparation and simply opening the Cloud screen — so N conflicting groups means N full GPX downloads before backup info is available.

2. It never checks whether the Favorites export type is enabled, even though the loop directly above does (`[OAExportSettingsType isTypeEnabled:exportType]`). A merge is prepared, and the remote file downloaded, for a type the user has turned off; `createItemCollections` then filters the result away.

(2) looks like a straightforward fix. (1) is more a question of whether network I/O belongs in this task's contract.

alex-dev-neo (commented):
Post-merge review, cross-checked against the Android counterpart (osmandapp/OsmAnd#25721). The merge algorithm itself holds up — the rename/deletion/edit cases in `mergeBase:local:remote:` resolve as described.

Three findings from the Android review carry over, and one is new and specific to this port (`copyPoint:` losing point visibility).

Worth calling out as better than the Android version: encoding the sync time in the snapshot path (`<root>/<fileName>/<syncTime>.gpx`) instead of stamping it onto the file's modification date. The Android side compares `file.lastModified() == syncTime` exactly, which silently no-ops if the filesystem truncates the timestamp; this design has no such dependency, and it also prunes stale generations, writes via a temp file plus atomic move, and applies `precomposedStringWithCanonicalMapping` consistently on both the save and load paths.

Also checked and found fine: `copyFavoriteLocation` does not insert into the live collection (`createFavoriteLocation` only constructs; `addFavoriteLocation` is what inserts), so the "preparation never changes local Favorites" invariant holds. Colour normalisation is fine too — `OAFavoriteItem.getColor` and `OAFavoriteGroup.color` already fall back to the default internally, so the explicit `sameColor` normalisation the Android version needs is not required here. And the `_uniqueRemoteFilesWithDecomposedNames` -> `_remoteFilesWithDecomposedNames` rename has exactly one use site, so folding the deleted files into that map does not affect anything else.

PR #5690, opened 2026-08-25
Retry only failed OSM uploads

### LINKED ISSUE #5591: "Retry" button re-uploads already successfully uploaded OSM Notes
### Description

When uploading multiple OSM Notes, tapping "Retry" after one upload fails causes all Notes to be uploaded again, including those that were already uploaded successfully.

Instead of tapping "Retry", close the upload dialog and open My Places → OSM Edits again.

The list is refreshed correctly and contains only the Notes that were not uploaded successfully.

### Steps to reproduce

1. Open My Places → OSM Edits.
2. Select several OSM Notes for upload.
3. Start uploading.
4. Let one Note fail (for example because of a temporary network error).
5. When the error dialog appears, tap "Retry".

### Actual result

All selected Notes are uploaded again, including those that were already uploaded successfully before the error.

### Expected result

Only the Notes that failed to upload should be retried.
Already uploaded Notes should not be uploaded again.

Platform: iOS
OsmAnd version: (last)
iOS version: (last)

aleksandr-tata:
## Summary

- Retry only OSM points that failed during the previous upload attempt.
- Keep the original upload list unchanged and store failed points separately.
- Calculate progress using the number of points in the current attempt.
- Adjust progress increments automatically as the number of remaining points changes between retries.

### DIFF
--- a/Sources/OsmEdit/OAUploadOsmPointsAsyncTask.m
+++ b/Sources/OsmEdit/OAUploadOsmPointsAsyncTask.m
@@ -23,12 +23,19 @@
 #import "OAOsmBugResult.h"
 #import "OARootViewController.h"
 
+@interface OAUploadOsmPointsAsyncTask ()
+
+- (void)uploadPoints:(NSArray<OAOsmPoint *> *)points;
+
+@end
+
 @implementation OAUploadOsmPointsAsyncTask
 {
     OAOsmEditingPlugin *_plugin;
     OsmAndAppInstance _app;
     
     NSArray<OAOsmPoint *> *_points;
+    NSArray<OAOsmPoint *> *_failedPoints;
     
     NSString *_comment;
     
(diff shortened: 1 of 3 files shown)


Sources/OsmEdit/OAUploadOsmPointsAsyncTask.m
```diff
@@ -23,12 +23,19 @@
 #import "OAOsmBugResult.h"
 #import "OARootViewController.h"
 
+@interface OAUploadOsmPointsAsyncTask ()
+
+- (void)uploadPoints:(NSArray<OAOsmPoint *> *)points;
```
tigrim: ?

aleksandr-tata: Not needed here. Removed the redundant private method declaration.

tigrim (approved):

PR #5702, opened 2026-08-28
Unify coordinate format picker - add screen

### LINKED ISSUE #5512: Unify coordinate format picker across Settings, Search and Widget
> Core 

<table style="width:100%">
  <tr>
    <th>Main screen</th>
  </tr>
  <tr>
    <td> </td>
  </tr>
</table>   

### NavBar

- [x] Leading action – Back

**Menu:**

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

### List
	- [x] DDD.DDDDD 
	- [x] DDD MM.MM
	- [x] DDD MM SS.SS
	- [x] Open Location Code
- [x] Second line: 

### ToolBar

<table style="width:100%">
  <tr>
    <th>Edit</th>
    <th>Confirmation dialog</th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
  </tr>
</table>   

**NavBar**

**List:**
	- [x] Leading – **Remove**
	- [x] Trailing – **Move**

**Confirmation alert:**

<table style="width:100%">
  <tr>
    <th></th>
    <th></th>
  </tr>
  <tr>
    <td> </td>
    <td> </td>
  </tr>
</table> 

**NavBar**

- [x] Title – `Add format`

**Search bar**

**Item:**
<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

**General group**

**Empty state**

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

**Search**

## Context menu

<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

vitaliy-sova-ios:


Sources/Controllers/Cells/OASimpleTableViewCell.h
```diff
@@ -37,6 +37,7 @@ typedef NS_ENUM(NSInteger, EOATableViewCellContentStyle) {
 
 - (void)updateSeparatorInset;
 - (void)setCustomLeftSeparatorInset:(BOOL)isCustom;
+- (void)setRightSeparatorInset:(CGFloat)inset;
```
tigrim: Why do we need the right inset here? We haven't used it elsewhere in the project before. How will it behave in RTL layouts?

Sources/Controllers/Settings/Coordinates/CoordinateFormat.swift
```diff
@@ -36,4 +36,44 @@ struct CoordinateFormat {
             isResolved: true
         )
     }
+    
+    static func epsg(code: Int, title: String?, subtitle: String?, isDeprecated: Bool) -> CoordinateFormat {
+        let fallback = "EPSG:\(code)"
+        return CoordinateFormat(
+            id: CoordinateFormatIds.epsg(code),
+            type: .epsg,
+            title: title?.isEmpty == false ? title! : fallback,
```
tigrim: force unwrap

Sources/Controllers/Settings/Coordinates/CoordinateFormatHelper.swift
```diff
@@ -11,11 +11,41 @@ import UIKit
 enum CoordinateFormatHelper {
     static let exampleLat = 50.43855
     static let exampleLon = 30.50124
+    static let unavailablePlaceholder = "—"
+    
+    private static let searchDebounce: TimeInterval = 0.25
+    private static var searchWorkItem: DispatchWorkItem?
```
tigrim: Should the search cancellation state really be global in CoordinateFormatHelper? searchWorkItem is static, so starting or cancelling a search from one screen also affects every other consumer. It might be safer to keep the debounce/cancellation state owned by the view controller or by a dedicated search instance

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
```diff
@@ -0,0 +1,150 @@
...
+//
+
+import Foundation
+import SQLite3
+
+final class EpsgCatalogRepository {
+
+    static let shared = EpsgCatalogRepository()
+
+    private static let defaultListLimit = 1000
```
tigrim, line 16: Why is the empty search limited to 1,000 entries if the catalog contains 5,000+ formats? Is the intention to show only a partial list until the user enters a query? If so, should the UI make this explicit or avoid presenting the partial result as the full catalog?

Sources/Controllers/Settings/Coordinates/CoordinatesFormatAddViewController.swift
```diff
@@ -0,0 +1,271 @@
...
+    
+    // MARK: - Search
+    
+    private func setupSearchController() {
+        searchController.delegate = self
+        searchController.searchResultsUpdater = self
+        searchController.obscuresBackgroundDuringPresentation = false
+        searchController.searchBar.placeholder = localizedString("coordinate_format_add_search_hint")
+        navigationItem.searchController = searchController
+        if #available(iOS 16.0, *) {
```
tigrim: 16.0 or 26?

Sources/Controllers/Settings/Coordinates/CoordinatesFormatAddViewController.swift
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// MARK: - UISearchResultsUpdating
+
+extension CoordinatesFormatAddViewController: UISearchResultsUpdating {
+    func updateSearchResults(for searchController: UISearchController) {
+        searchQuery = searchController.searchBar.text ?? ""
+        if isSearching {
+            let query = searchQuery
+            CoordinateFormatHelper.search(query) { [weak self] results in
```
tigrim: Can we reuse runCatalogSearch(_:) here instead of duplicating the search logic? The two paths already have slightly different completion checks (isSearchActive is checked only in runCatalogSearch), which makes their behavior easier to diverge

Sources/Controllers/Settings/Coordinates/CoordinateFormatHelper.swift
```diff
@@ -34,11 +64,38 @@ enum CoordinateFormatHelper {
...
+        }
+        return unavailablePlaceholder
+    }
+
+    static func formatEpsgPoint(easting: Double, northing: Double) -> String {
+        "\(formatEpsgValue(easting)), \(formatEpsgValue(northing))"
+    }
+
+    static func formatEpsgValue(_ value: Double) -> String {
+        let formatter = NumberFormatter()
```
tigrim: static let

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
vitaliy-sova-ios: This matches Android (DEFAULT_LIST_LIMIT = 1000, empty query → listAll(), typed query → search with limit 50).

Sources/Controllers/Settings/Coordinates/CoordinatesFormatAddViewController.swift
vitaliy-sova-ios: PreferredSearchBarPlacement was introduced in iOS 16.0 (not 26). We use #available(iOS 16.0, *) consistently elsewhere in the project for .stacked

Sources/Controllers/Settings/Coordinates/CoordinatesFormatAddViewController.swift
vitaliy-sova-ios: Done — both updateSearchResults and willPresentSearchController now go through performSearch(_:), with a single completion check (isSearchActive + matching searchQuery). The old duplicated CoordinateFormatHelper.search paths are gone.

tigrim (approved):

PR #5705, opened 2026-08-28
Fix CarPlay map button icons on iOS 26+

### LINKED ISSUE #5662: Update CPMapButton CarPlay
<table style="width:100%">
  <tr>
    <th>iOS 27</th>
    <th>iOS 18.5</th>
  </tr>
  <tr>
    <td> </td> 
	<td> </td>
  </tr>
</table>   

The problem is limited to `CPMapButton`: those assets bake in their own background, because before iOS 26 CarPlay drew none.

- [x] Add glyph-only versions of the four `CPMapButton` assets: transparent background, no baked shape, template rendering mode
- [x] Select the asset set at runtime: `if #available(iOS 26.0, *)` → glyph-only, otherwise the existing composite asset
- [x] Keep the existing composite assets unchanged — iOS 15–18 rendering must stay exactly as it is now
- [x] Update `focusedImage` the same way, for both sets (rotary-controller focus state)
- [x] Check optical sizing of the new assets — removing the background changes the glyph's effective size inside the same box
- [x] Verify on iOS 26 in both light and dark car appearance, and confirm no regression on iOS 18

### Assets

New for iOS 26+
- [x] **3D** – `ic_custom_3d`
- [x] **2D** – `ic_custom_2d`
- [x] **My location** – `ic_custom_map_location_position`
- [x] **Zoom in** – `ic_custom_map_zoom_in`
- [x] **Zoom out** – `ic_custom_map_zoom_out`

aleksandr-tata:
## Summary

- Use the glyph-only `ic_custom_*` assets for `CPMapButton` controls on iOS 26 and later.
- Keep the existing composite `btn_map_*` assets unchanged on iOS 18 and earlier.
- Update both `image` and `focusedImage` for rotary-controller focus.
- Preserve the existing 2D/3D icon selection based on the current map mode.
- Reuse the assets provided in the task without adding CarPlay-specific duplicates.


Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -351,6 +351,7 @@ - (void) enterRoutePreviewMode:(BOOL)shouldCheckConnectedMainScene
             destinationSubtitle = nil;
     }
     
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 260400
```
aleksandr-tata, line 358: Xcode 26.2 Fix

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -469,6 +473,53 @@ - (void)openNavigation
...
+                break;
+            case EOACarPlayButtonType3D:
+                image = [UIImage imageNamed:[OAMapViewTrackingUtilities.instance is3DMode] ? ACImageNameBtnMap2DMode : ACImageNameBtnMap3DMode];
+                break;
+            default:
+                break;
+        }
+    }
+    mapButton.image = image;
+    mapButton.focusedImage = image;
```
tigrim: Do we need to set `focusedImage` explicitly here? According to `CPMapButton` documentation, when `focusedImage` is nil, CarPlay automatically creates the focused appearance from `image`. Using the same image for both may prevent the standard focus effect

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -469,27 +476,102 @@ - (void)openNavigation
     [directionsGrid present];
 }
 
+- (void)updateMapButton:(CPMapButton *)mapButton
```
tigrim: Could we move this into a common image helper and pass whether the image should use template rendering? This would avoid having separate [UIImage templateImageNamed:] and [UIImage imageNamed:] branches here.

```
+ (UIImage *)imageNamed:(NSString *)imageName template:(BOOL)template
{
    return template
        ? [UIImage templateImageNamed:imageName]
        : [UIImage imageNamed:imageName];
}
```

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -469,27 +476,102 @@ - (void)openNavigation
...
             OAMapViewTrackingUtilities *mapViewTrackingUtilities = [OAMapViewTrackingUtilities instance];
+            BOOL is3DMode = [mapViewTrackingUtilities is3DMode];
             Map3DModeVisibility map3DMode = [[[OAMapButtonsHelper sharedInstance] getMap3DButtonState] getVisibility];
             BOOL hideButton = map3DMode == Map3DModeVisibilityHidden
-                || (map3DMode == Map3DModeVisibilityVisibleIn3DMode && ![mapViewTrackingUtilities is3DMode]);
-            _3DModeMapButton.hidden = hideButton ? YES : NO;
-            if ([mapViewTrackingUtilities is3DMode])
+                || (map3DMode == Map3DModeVisibilityVisibleIn3DMode && !is3DMode);
+            BOOL shouldRecreateButton = NO;
+            if (@available(iOS 26.0, *))
```
tigrim: The button creation logic is also unclear

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -155,7 +156,9 @@ - (void) enterBrowsingState
     _mapTemplate.trailingNavigationBarButtons = @[panningButton];
     _mapTemplate.leadingNavigationBarButtons = @[[self createBarButton:EOACarPlayButtonTypeSettings], [self createBarButton:EOACarPlayButtonTypeDirections]];
 
-    _3DModeMapButton = [self createMapButton:EOACarPlayButtonType3D];
+    BOOL is3DMode = [OAMapViewTrackingUtilities.instance is3DMode];
+    _lastIs3DMode = is3DMode;
```
tigrim: Why do we need to store is3DMode in _lastIs3DMode here? How will it behave if the 2D/3D mode is changed from the phone while CarPlay is connected?

Sources/CarPlay/OACarPlayDashboardInterfaceController.mm
```diff
@@ -469,27 +477,92 @@ - (void)openNavigation
...
+- (void)replaceMapModeButton:(CPMapButton *)mapButton
+{
+    NSUInteger buttonIndex = [_mapTemplate.mapButtons indexOfObjectIdenticalTo:_mapModeButton];
+    if (buttonIndex == NSNotFound)
+        return;
+
+    NSMutableArray<CPMapButton *> *mapButtons = _mapTemplate.mapButtons.mutableCopy;
+    mapButtons[buttonIndex] = mapButton;
+    _mapModeButton = mapButton;
+    _mapTemplate.mapButtons = mapButtons;
```
tigrim, line 524: copy

tigrim: Why do we need to recreate CPMapButton on iOS 26 when switching between 2D and 3D? If this is required because CarPlay caches the automatically generated focused image, could we document that workaround here? Otherwise, updating the existing button’s image would avoid replacing the entire mapButtons array

aleksandr-tata: Yes, this is a workaround for CarPlay caching the focused appearance generated from image when focusedImage is nil. Updating image changes the normal state, but the focused state may keep the previous 2D/3D glyph. Assigning focusedImage explicitly would disable the standard CarPlay focus appearance, so the button is recreated to force CarPlay to generate it again. I’ll document this workaround in the code.

aleksandr-tata:
Recreating the 2D/3D CPMapButton on iOS 26+ works around a stale focused icon: updating image changes the normal icon, but rotary-controller focus can still show the previous 2D/3D glyph. Replacing the button makes CarPlay regenerate its standard focused appearance.
The trade-off is that replacing a focused button can reset focus or move it to another control, as shown in the recording. If the button is focused using the rotary controller and its state then changes, this focus jump is a side effect of the workaround. Replacement only happens when the 2D/3D state changes.

aleksandr-tata:
Demo of issue
https://github.com/user-attachments/assets/d0a853b8-a8de-4e88-960d-c742ea2dec98

aleksandr-tata:
https://github.com/user-attachments/assets/0263c6d1-d1c1-4c86-bab3-c3d5cc13b929

Result after fix

tigrim (approved):

PR #5706, opened 2026-08-28
Follow up my places redesign UI second

### LINKED ISSUE #5612: Follow up for My Places redesign

DmitrySvetlichny:


Sources/SwiftExtensions/NumberFormatter+Extension.swift
```diff
@@ -24,4 +24,29 @@ extension NumberFormatter {
         formatter.locale = Locale(identifier: OAUtilities.currentLang() ?? Locale.current.identifier)
         return formatter
     }()
+
+    private static let countFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .decimal
+        formatter.locale = .current
```
tigrim: Should this formatter use the OsmAnd language instead of `Locale.current`?

Other formatters in this extension use `OAUtilities.currentLang() ?? Locale.current.identifier`. With the current implementation, if the app language differs from the system locale, we can get UI text in one language but count formatting from another locale.

Resources/Localizations/uk.lproj/Localizable.strings
```diff
@@ -2774,7 +2774,7 @@
 "history_preferences_descr" = "Ви можете вимкнути запис історії окремо для кожного типу історії.";
 "history_actions_footer_text" = "Ви можете експортувати історію у файл. Ви можете видалити всі записи історії, записані до цього моменту.";
 "history_clear_alert_message" = "Ви впевнені, що хочете видалити всі елементи історії?";
-"ltr_or_rtl_combine_via_of" = "%ld з %ld";
+"ltr_or_rtl_combine_via_of" = "%@ з %@";
```
tigrim, line 2777: Why did you change the localization values manually? Only the script should modify these values

DmitrySvetlichny: I’m not sure the localization script propagates placeholder type changes from the English localization to all other languages. These changes only migrate %d/%ld to %@; the translations themselves were not modified.
Updating all existing locales also allowed the formatting and pluralization to be tested consistently across different languages. Reverting these changes would leave some locales expecting numeric arguments while the code now passes preformatted strings, so keeping the placeholder migration seems necessary.

tigrim: I see that you changed it manually. The build clean script will reset these changes, so they will only remain locally

tigrim (approved):

Sources/SwiftExtensions/NumberFormatter+Extension.swift
```diff
@@ -24,4 +24,29 @@ extension NumberFormatter {
         formatter.locale = Locale(identifier: OAUtilities.currentLang() ?? Locale.current.identifier)
         return formatter
     }()
+
+    private static let countFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .decimal
+        formatter.locale = Locale(identifier: OAUtilities.currentLang() ?? Locale.current.identifier)
```
alex-dev-neo: **Locale drops the region, so counts are grouped with the wrong separator**

`OAUtilities.currentLang()` returns `[NSLocale preferredLanguages].firstObject` split on `-`, i.e. a language-only code. Feeding that to `Locale(identifier:)` discards the region, and grouping conventions are region-specific:

| device language | `currentLang()` | 1234567 formatted | correct |
|---|---|---|---|
| `es-US` | `es` | `1.234.567` | `1,234,567` |
| `en-IN` | `en` | `1,234,567` | `12,34,567` |
| `pt-BR` | `pt` | `1.234.567` | `1.234.567` (ok) |

For an es-US user `1.234` reads as a decimal number. This is also a behaviour regression: `FavoriteFolderStats.formattedPointsCount` previously used `NumberFormatter.localizedString(from:number:.decimal)`, which honours the full current locale including region.

Consider `Locale.autoupdatingCurrent` (or at least preserving the region) for number formatting, which is a display-region concern rather than a UI-language concern.

Sources/SwiftExtensions/NumberFormatter+Extension.swift
```diff
@@ -24,4 +24,29 @@ extension NumberFormatter {
         formatter.locale = Locale(identifier: OAUtilities.currentLang() ?? Locale.current.identifier)
         return formatter
     }()
+
+    private static let countFormatter: NumberFormatter = {
```
alex-dev-neo, line 28: **`countFormatter` duplicates `localizedNumberFormatter`**

This formatter is configured identically to `localizedNumberFormatter` four lines above (`.decimal` + the same locale expression). `usesGroupingSeparator = true` is already implied by `.decimal`, so the two are functionally the same object.

The cost is that any future fix to the locale expression has to be applied in two places, and whichever one is missed silently produces different digits in different parts of the app.

Sources/SwiftExtensions/NumberFormatter+Extension.swift
```diff
@@ -24,4 +24,29 @@ extension NumberFormatter {
...
+    static func localizedCount<T: BinaryInteger>(_ count: T) -> String {
+        if let signedCount = Int64(exactly: count) {
+            return localizedCount(NSNumber(value: signedCount))
+        }
+
+        if let unsignedCount = UInt64(exactly: count) {
+            return localizedCount(NSNumber(value: unsignedCount))
+        }
+
+        return String(describing: count)
```
alex-dev-neo, line 50: **Unreachable fallback**

After `Int64(exactly:)` and `UInt64(exactly:)`, this line is unreachable for every integer type used in the codebase. If it ever were reached it would return unformatted digits with no grouping and no localized numerals — silently defeating the purpose of the helper, with no test that would surface it.

Sources/Controllers/MyPlaces/OsmEditsListViewController.swift
```diff
@@ -676,10 +677,12 @@ final class OsmEditsListViewController: UIViewController, MyPlacesScrollResettab
         let selectedPoints = Array(selectionManager.selectedItems)
         if !selectedPoints.isEmpty {
             let count = selectedPoints.count
-            let message = String(format: localizedString("osm_edits_delete_items_confirmation"), count)
+            let formattedCount = NumberFormatter.localizedCount(count)
+            let messageKey = count == 1 ? "osm_edits_delete_item_confirmation" : "osm_edits_delete_items_confirmation"
```
alex-dev-neo: **`count == 1` can only model a two-form plural**

This ternary is an improvement over always using the plural key, but it still assumes languages have exactly two forms. This same PR demonstrates the right mechanism — it adds `folder_tracks_count` to all 82 `.stringsdict` files with `one`/`few`/`many`/`other`.

Concretely, Czech with 5 selected edits:

```
"osm_edits_delete_items_confirmation" = "Přejete si skutečně odstranit %@ úpravy?";
```

`úpravy` is the *few* (2–4) form, so the user sees “…odstranit 5 úpravy?” instead of “…5 úprav?”. Same for ru/pl/uk (many form) and ar (six forms).

A `NSStringPluralRuleType` entry, keyed exactly the way `folder_tracks_count` now is, would collapse the two keys into one and pick the right form everywhere.

Sources/Controllers/MyPlaces/OsmEditsListViewController.swift
```diff
@@ -523,10 +523,11 @@ final class OsmEditsListViewController: UIViewController, MyPlacesScrollResettab
     
     private func delete(_ point: OAOsmPoint) {
         let count = 1
```
alex-dev-neo: **This message-building block is now duplicated verbatim**

`delete(_:)` and `deleteButtonPressed(_:)` contain the same eight lines: format the message, wrap in `NSMutableAttributedString`, `range(of: formattedCount)`, apply the bold font.

They have already drifted apart within this PR — the singular/plural key selection was added at line 681 but not here. Also, `let count = 1` exists only to be formatted back into the literal `"1"`.

Extracting `private func deleteConfirmationMessage(count: Int) -> NSAttributedString` would collapse both call sites and drop the placeholder constant.

Sources/Controllers/MyPlaces/TracksViewController.swift
```diff
@@ -675,9 +671,9 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
                 
                 let totalTracks = totalSelectedTracks + tracksInSelectedFolders
                 let itemText = localizedString("shared_string_item").lowercased()
```
alex-dev-neo: **Selection title always uses the singular noun**

`shared_string_item` is used unconditionally, so selecting 5 tracks gives “5 item”. The equivalent code in `OsmEditsListViewController.updateNavigationBarTitle` (line 443) already handles this:

```swift
let itemText = localizedString(totalSelectedPoints > 1 ? "shared_string_items" : "shared_string_item").lowercased()
```

The changed line localizes the digits but leaves the noun ungrammatical in every language.

alex-dev-neo (commented):
Automated review of the count-localization changes (source files only; the 82 locale resource files were checked separately and are consistent — see the note at the end).

13 comments below, roughly most-severe first. The top one is a genuine behaviour regression; several others are pre-existing issues on lines this PR touches, flagged because they sit squarely in the area being fixed.

---

**Checked and clean**, for the record:

- All 82 `.stringsdict` files received the new `folder_tracks_count` entry, with positional `%2$@` placeholders and `ld` value types throughout — no locale is left with a `%ld`/`%d` specifier that would now receive an `NSString`.
- Every locale that defines `ltr_or_rtl_combine_via_of`, `remove_folder_with_files_descr`, `osm_edits_delete_*`, `items_delete_confirmation_title` or `import_tracks_descr_*` was converted to `%@`; none were missed, and none reorder the placeholders.
- The two-argument plural pattern (`count` for selection, formatted string for display) resolves correctly for both the 2-arg and 4-arg keys, verified against `NSBundle`.
- `Scripts/add_translations.swift` will not clobber the hand-edited translated `.strings`: none of these keys exist in the Android `strings.xml` it syncs from.
- The Objective-C call sites compile — the extension is `@objc` and all three `.mm` files already import `OsmAnd_Maps-Swift.h`.

Sources/Controllers/MyPlaces/Import/ImportTracksViewController.swift
```diff
@@ -570,9 +570,10 @@ private extension ImportTracksViewController {
     }
 
     func makeImportTracksDescription(tracksCount: Int) -> NSAttributedString {
+        let formattedTracksCount = NumberFormatter.localizedCount(tracksCount)
         let text = tracksCount == 1
```
alex-dev-neo: **Manual `one`/`other` key split instead of a plural rule**

Selecting between `import_tracks_descr_one` and `import_tracks_descr_other` on `tracksCount == 1` is the same two-form assumption as elsewhere in this PR. Russian import of a file with 3 tracks gets the `_other` string, which translators wrote as the 5+ (many) form.

Separately: `_one` was made positional (`%1$@ … %2$@`) while `_other` stayed non-positional (`%@ … %@`). Both work today, but since both placeholders are now `%@` and no longer distinguishable by type, a translator swapping them would silently produce “The file 3 contains myfile.gpx separate tracks.” The old `%@` + `%d` pairing made that mistake impossible.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+Actions.swift
```diff
@@ -537,11 +537,11 @@ extension FavoriteListViewController {
...
 
         if foldersCount > 0 && pointsCount == 0 {
-            return String.localizedStringWithFormat(NSLocalizedString("folders_delete_confirmation_title", comment: ""), foldersCount)
+            return String.localizedStringWithFormat(NSLocalizedString("folders_delete_confirmation_title", comment: ""), foldersCount, NumberFormatter.localizedCount(foldersCount))
         } else if pointsCount > 0 && foldersCount == 0 {
-            return String.localizedStringWithFormat(NSLocalizedString("favorites_delete_confirmation_title", comment: ""), pointsCount)
+            return String.localizedStringWithFormat(NSLocalizedString("favorites_delete_confirmation_title", comment: ""), pointsCount, NumberFormatter.localizedCount(pointsCount))
         } else {
-            return String(format: localizedString("items_delete_confirmation_title"), pointsCount + foldersCount)
+            return String(format: localizedString("items_delete_confirmation_title"), NumberFormatter.localizedCount(pointsCount + foldersCount))
```
alex-dev-neo: **This branch has no plural rule, unlike the two above it**

The two branches directly above use `String.localizedStringWithFormat` against `.stringsdict` plural entries; this one uses plain `String(format:)` against `items_delete_confirmation_title` (“Delete %@ items?”), whose noun is hard-coded plural.

Czech with 1 folder + 4 points (5 items) renders “Odstranit 5 položky?” — the 2–4 form instead of `položek`. The branch is also reached when the selection contains neither type, giving “Delete 0 items?”.

Sources/Controllers/MyPlaces/TracksFilterDetailsViewController.swift
```diff
@@ -353,7 +353,7 @@ final class TracksFilterDetailsViewController: OABaseNavbarViewController {
                         filteredTracks.contains(where: { $0.path == trackItem.path })
```
alex-dev-neo: **O(n×m) membership scan, repeated per folder**

`folderTracks.filter { filteredTracks.contains(where: { $0.path == trackItem.path }) }` scans the whole filtered list for every folder track, and `displayFolder` re-runs it for each folder including recursive subfolders (the same pattern is at line 370).

With 2,000 tracks across 20 folders that is tens of millions of string comparisons on the main thread each time `generateData()` runs, and the filter sheet visibly stalls.

```swift
let filteredPaths = Set(filteredTracks.map(\.path))
let matchingTracksCount = folderTracks.count { filteredPaths.contains($0.path) }
```

Hoisting the `Set` out of the loop makes it linear.

Sources/Controllers/MyPlaces/TracksFilterDetailsViewController.swift
```diff
@@ -353,7 +353,7 @@ final class TracksFilterDetailsViewController: OABaseNavbarViewController {
...
                     }.count
                     let totalTracksCount = folderTracks.count
-                    allFoldersRow.descr = "\(matchingTracksCount)/\(totalTracksCount)"
+                    allFoldersRow.descr = "\(NumberFormatter.localizedCount(matchingTracksCount))/\(NumberFormatter.localizedCount(totalTracksCount))"
```
alex-dev-neo: **Hard-coded `/` bypasses `ltr_or_rtl_combine_via_slash`**

`SelectPointsViewController.swift:329` — changed in this same PR — builds the identical construct through the RTL-aware helper:

```swift
String(format: localizedString("ltr_or_rtl_combine_via_slash"),
       NumberFormatter.localizedCount(…), NumberFormatter.localizedCount(…))
```

Here (and at line 373, and `ImportTracksViewController.swift:408`) the slash is a literal, so in Hebrew/Arabic these rows order the counts differently from the visually identical rows in the point-selection screen. `ltr_or_rtl_combine_via_slash` is already the codebase convention — it is used in `VehicleMetricsPlugin`, `TrackChartHelper` and `PlanRouteAnalyzeViewController`.

Sources/Controllers/MyPlaces/TracksChangeAppearanceViewController.swift
```diff
@@ -557,7 +557,7 @@ final class TracksChangeAppearanceViewController: OABaseNavbarViewController {
     }
     
     private func applyExistingTracksTitle() -> String {
-        String(format: localizedString("ltr_or_rtl_combine_via_space"), localizedString("apply_to_existing"), "(\(tracks.count))")
+        String(format: localizedString("ltr_or_rtl_combine_via_space"), localizedString("apply_to_existing"), "(\(NumberFormatter.localizedCount(tracks.count)))")
```
alex-dev-neo: **Hand-built parentheses bypass `ltr_or_rtl_combine_with_brackets`**

The string key `"ltr_or_rtl_combine_with_brackets" = "%@ (%@)"` already exists for exactly this and is used by `AstronomyPlugin`, `GradientEditorViewController` and `OBDRemainingFuelWidget`.

The same hand-built `" (…)"` appears at `TracksFiltersViewController.swift:142`, `TracksViewController.swift:804` and `TracksViewController.swift:675`. In RTL locales the literal parentheses are not reordered consistently with the surrounding title, and the four sites now have to be fixed individually.

Sources/Controllers/MyPlaces/TracksSortModeHelper.swift
```diff
@@ -137,22 +137,26 @@ protocol SortableFolder {
...
+    }
     
     @objc static func getTrackDescription(track: GpxDataItem, sortMode: TracksSortMode, includeFolderInfo: Bool = false) -> NSAttributedString {
         let date = TracksSortModeHelper.dateFormatter.string(from: track.lastModifiedTime)
         let creationDate = TracksSortModeHelper.dateFormatter.string(from: track.creationDate)
         let distance = OAOsmAndFormatter.getFormattedDistance(track.totalDistance) ?? localizedString("shared_string_not_available")
         let duration = track.getAnalysis()?.getDurationInSeconds() ?? 0
         let time = OAOsmAndFormatter.getFormattedTimeInterval(TimeInterval(duration), shortFormat: true) ?? localizedString("shared_string_not_available")
-        let waypointCount = "\(track.wptPoints)"
+        let waypointCount = NumberFormatter.localizedCount(track.wptPoints)
```
alex-dev-neo, line 159: **Grouping style diverges from the app's existing number formatter**

`localizedCount` groups from 1,000 upward using the locale's default separator. The app's established formatter, `OAOsmAndFormatter.formatValue:` (`Sources/Helpers/OAOsmAndFormatter.mm:315-337`), forces a **space** separator and only groups at values ≥ 10,000.

`getTrackDescription` puts both in one label:

```swift
let detailsText = "\(distance) • \(time) • \(waypointCount)"
```

so a track footnote can read `12 345 km • 1:23 • 12,345` — and in German the two become `12 345` and `12.345` side by side. Worth deciding on one convention rather than two.

Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1878,7 +1878,7 @@
 
 "select_tracks" = "Select tracks";
 "selected_tracks_count" = "Tracks - %d";
-"folder_tracks_count" = "%d tracks";
+"folder_tracks_count" = "%@ tracks";
```
alex-dev-neo: **This edit has no effect — the `.stringsdict` entry shadows it**

`folder_tracks_count` is now defined in **both** `Localizable.strings` and `Localizable.stringsdict`. `.stringsdict` takes precedence in `NSBundle` lookup, so changing this value from `%d tracks` to `%@ tracks` is dead in all 82 locales. Verified against `NSBundle.localizedString(forKey:)`, which returns the raw `%1$#@tracks@`.

The hazard is for the next caller: reading this file suggests `String(format: localizedString("folder_tracks_count"), count)` is correct, but that renders the literal `%1$#@tracks@` token in the UI.

Also, `folder_one_track` on the next line lost its last reference when `TracksViewController.formattedTracksCount` was deleted, and is now orphaned in all 82 locale files.

Sources/SwiftExtensions/NumberFormatter+Extension.swift
DmitrySvetlichny: After preserving the full regional locale, these formatters are no longer configured identically. localizedNumberFormatter is an existing parser used by PlanRouteAnalyzeViewController, while countFormatter formats display counts using Locale.autoupdatingCurrent. Reusing the former would either restore the region bug or change existing parsing behavior outside this task.

Sources/SwiftExtensions/NumberFormatter+Extension.swift
DmitrySvetlichny: The fallback is unreachable for the fixed-width integer types currently used by the project, but not for the generic BinaryInteger contract: a custom wider implementation may fit neither Int64 nor UInt64. Keeping a non-trapping fallback is intentional. All current callers use one of the localized NSNumber paths, so this fallback does not affect current UI formatting.

Sources/Controllers/MyPlaces/TracksSortModeHelper.swift
DmitrySvetlichny: This difference is intentional for this task. localizedCount formats user-facing counts using the locale’s decimal grouping rules, while OAOsmAndFormatter.formatValue is the existing measurement formatter with OsmAnd-specific grouping. The task explicitly requires keeping distance and size formatting unchanged and forbids hard-coding a grouping separator for counts. Unifying both conventions would require a separate app-wide formatting change, so I suggest keeping this code as is.

PR #5709, opened 2026-08-31
Migrate reverse geocoder to async

vitaliy-sova-ios:


Sources/Router/OATargetPointsHelper.mm
```diff
@@ -22,6 +22,153 @@
...
+
+@interface OAAddressLookupRequest : NSObject
+@property(nonatomic) NSUInteger generation;
+@property(nonatomic, copy) NSString *coordinateKey;
+@end
+
+@implementation OAAddressLookupRequest
+@end
+
+@interface OAAddressLookupCoordinator ()
```
tigrim: Do we really need a separate `OAAddressLookupCoordinator` here? `OAReverseGeocoder` already handles async lookup, caching, and deduplication. Could we just validate that the target/coordinates are still current in each completion handler, as we do in `OADestinationsLayer`, and avoid this additional coordination layer?

Sources/Plugins/Astronomy/contextmenu/AstroVisibilityCardController.swift
```diff
@@ -123,8 +121,6 @@ final class AstroVisibilityCardController {
     func cancelPendingWork() {
         computeWorkItem?.cancel()
         computeWorkItem = nil
-        locationWorkItem?.cancel()
```
tigrim, line 126: Previously cancelPendingWork() also cancelled the pending location resolution. After this change the reverse-geocoding callback can still update locationText and call onDataChanged after cancelPendingWork() if lastLocationKey has not changed. Should we invalidate the current location request/key here as well?

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -554,208 +545,197 @@ - (void) setWorkPoint:(CLLocation *) latLon description:(OAPointDescription *)na
...
-    CLLocation *location = destination.point;
-    BOOL isNameNotValid = [destination isSearchingAddress];
-    
+
+    const double lookupLat = point.getLatitude;
+    const double lookupLon = point.getLongitude;
+    const BOOL isNameNotValid = [point isSearchingAddress];
     __weak __typeof(self) weakSelf = self;
-    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+    [self getLocationName:point.point completion:^(NSString *address) {
```
tigrim, line 683: Removing the per-point in-flight guards allows `lookupAllAddresses` to register multiple completions for the same target while reverse geocoding is still running. `OAReverseGeocoder` coalesces the lookup itself, but retains every completion, so all of them will persist the same point and notify listeners when the result arrives. Home/work lookups may also write favorites multiple times. Please keep a per-target request token/flag or otherwise invalidate duplicate completions.

Sources/Controllers/Map/Layers/OADestinationsLayer.mm
```diff
@@ -714,20 +714,43 @@ - (BOOL)isObjectMovable:(id)object
...
+                                            objectId:0
+                                          completion:^(NSString *address) {
+        __strong __typeof(weakSelf) strongSelf = weakSelf;
+        if (!strongSelf || address.length == 0)
+            return;
+
+        if ([strongSelf.app.data.destinations indexOfObjectIdenticalTo:destCopy] == NSNotFound)
+            return;
+        
+        if (![OAUtilities doublesEqualUpToDigits:5 source:destCopy.latitude destination:position.latitude] ||
```
tigrim: Do we still need this coordinate check? `destCopy` is created for this exact move and its coordinates are assigned from `position` before starting the lookup, so these values should always match. It looks like the identity check above is the actual stale-result protection here.

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -554,208 +552,211 @@ - (void) setWorkPoint:(CLLocation *) latLon description:(OAPointDescription *)na
...
+        return;
+
+    const double lookupLat = point.getLatitude;
+    const double lookupLon = point.getLongitude;
+    NSUUID *lookupToken = [NSUUID UUID];
+    _homeLookupToken = lookupToken;
+    __weak __typeof(self) weakSelf = self;
+    [self getLocationName:point.point completion:^(NSString *address) {
+        __strong __typeof(weakSelf) strongSelf = weakSelf;
+        if (!strongSelf || ![strongSelf->_homeLookupToken isEqual:lookupToken])
```
tigrim: isEqualToString

vitaliy-sova-ios: done

Sources/Router/OATargetPointsHelper.mm
```diff
@@ -554,208 +552,211 @@ - (void) setWorkPoint:(CLLocation *) latLon description:(OAPointDescription *)na
...
-                __strong __typeof(weakSelf) strongSelf = weakSelf;
-                if (!strongSelf)
-                    return;
-                
-                if ([strongSelf->_intermediatePoints indexOfObjectIdenticalTo:targetPoint] == NSNotFound)
-                    return;
+    const double lookupLat = point.getLatitude;
+    const double lookupLon = point.getLongitude;
+    NSString *lookupToken = [NSUUID UUID].UUIDString;
+    [_intermediateLookupTokens setObject:lookupToken forKey:point];
```
alex-dev-neo, line 719: `_intermediateLookupTokens` is created with `+[NSMapTable weakToStrongObjectsMapTable]`, which uses **object personality** — keys are hashed with `-hash` and compared with `-isEqual:`, not by pointer identity.

`OARTargetPoint.hash` mixes in the mutable `index` property:

```objc
- (NSUInteger) hash
{
    NSUInteger result = self.point.hash;
    result = 31 * result + self.index;
    ...
}
```

So mutating `index` after the token has been stored moves the key into a different bucket and strands the entry.

Concrete failure: intermediate points A (index 0) and B (index 1). `lookupAddressForIntermediatePoint:` stores B's token while `B.index == 1`. Before the geocode returns, the user deletes A — `deleteIntermediatePoint:` reindexes the remainder (`tp.index = ind++`), so `B.index` becomes 0 and B's hash changes. When the completion fires, `[_intermediateLookupTokens objectForKey:point]` returns `nil`, `![nil isEqualToString:lookupToken]` is `YES`, and the block returns early. B keeps the "Looking up address…" placeholder name and an empty `pointDescription.address` until some later `readFromSettings` happens to re-run the lookup. The old entry also stays stranded under the stale hash.

Keying by pointer identity avoids this:

```objc
_intermediateLookupTokens = [[NSMapTable alloc]
    initWithKeyOptions:NSPointerFunctionsWeakMemory | NSPointerFunctionsObjectPointerPersonality
          valueOptions:NSPointerFunctionsStrongMemory
              capacity:0];
```

Alternatively, store the token directly on the `OARTargetPoint`, matching how the other five lookups use a plain ivar.

PR #5712, opened 2026-08-31
Fix overlapping context menu presentations

tigrim:
## Related issue / task

Fixes # https://github.com/osmandapp/OsmAnd-Issues/issues/3307

## Summary

* Fixed a race condition when rapidly switching between POI, GPX, and multi-context menus.
* Added a shared context-menu presentation coordinator.
* New menu presentations now wait until the current menu is fully dismissed.
* While a transition is in progress, only the latest presentation request is retained.
* Added cleanup for stale polygon highlighting and transport-route UI state.
* Preserved the existing GPX track navigation-history behavior.
* Added unit and UI test coverage for context-menu presentation transitions.

## Testing

**Tested on:**

* Device / Simulator: iPhone 16 Pro Simulator
* iOS: 18.6

### Scenarios

* Rapidly switch between POI and different GPX tracks.
* Open a multi-context menu while another context menu is being dismissed.
* Verify that only the latest queued context menu is presented.
* Verify that regular and multi-context menus do not overlap.
* Open a waypoint from the current track menu and use **Open track**.
* Verify that the current track menu is restored correctly.

## Relevant checks

* [ ] Portrait / Landscape tested
* [ ] Light / Dark mode tested
* [ ] Different profiles tested
* [ ] Localization / RTL checked
* [ ] VoiceOver / Accessibility checked
* [ ] CarPlay checked
* [ ] Android parity checked
* [x] Performance impact checked

## AI disclaimer

**Implementation:**

* Tool / Agent: Codex
* Model: GPT-5.6 Sol

1. Analyze the context-menu overlap race when rapidly switching between POI, GPX, and multi-context menus.
2. Fix presentation ordering so a new context menu is shown only after the previous one is fully dismissed.
3. Preserve existing GPX track navigation behavior and related menu cleanup.
4. Add automated coverage for the context-menu presentation race.
5. Add UI tests and configure them to run in CI.

* Introduced a shared `ContextMenuPresentationCoordinator` instead of duplicating transition handling across individual presentation paths.
* Used a latest-request-wins strategy while a menu dismissal is in progress.
* Added dedicated UI-test launch arguments and accessibility identifiers for deterministic test scenarios.
* Added unit tests for coordinator state transitions in addition to UI tests.

**Final review:**

* Tool / Agent: ChatGPT
* Model: GPT-5.6 Sol
* [x] Final diff reviewed

**Significant findings:**

* Reviewed navigation-history restoration when replacing GPX track menus.
* Reviewed dismissal completion ordering and transport-route cleanup.
* Reviewed stale polygon-highlight handling.
* Reviewed UI-test coverage for queued menu presentation and current-track waypoint flow.

### DIFF
--- a/Sources/Controllers/Base/OABaseScrollableHudViewController.mm
+++ b/Sources/Controllers/Base/OABaseScrollableHudViewController.mm
@@ -35,6 +35,7 @@
     
     BOOL _isDragging;
     BOOL _isHiding;
+    void (^_forceHideCompletion)(void);
     CGFloat _initialTouchPoint;
     CGFloat _tableViewContentOffsetY;
 }
@@ -488,6 +489,12 @@
     [self hide];
 }
 
+- (void)forceHideWithCompletion:(void (^)(void))onComplete
+{
+    _forceHideCompletion = [onComplete copy];
+    [self forceHide];
+}
+
 - (void) hide:(BOOL)animated duration:(NSTimeInterval)duration onComplete:(void (^)(void))onComplete
 {
     _isHiding = YES;
@@ -502,14 +509,28 @@
             [_mapPanel hideScrollableHudViewController];
             _scrollableView.frame = frame;
         } completion:^(BOOL finished) {
-            [self dismissViewControllerAnimated:NO completion:onComplete];
+            [self dismissViewControllerAnimated:NO completion:^{
+                if (onComplete)
+                    onComplete();
+                void (^forceHideCompletion)(void) = _forceHideCompletion;
+                _forceHideCompletion = nil;
+                if (forceHideCompletion)
+                    forceHideCompletion();
+            }];
         }];
     }
     else
     {
         [_mapPanel hideScrollableHudViewController];
         _scrollableView.frame = frame;
-        [self dismissViewControllerAnimated:YES completion:onComplete];
+        [self dismissViewControllerAnimated:YES completion:^{
+            if (onComplete)
+                onComplete();
+            void (^forceHideCompletion)(void) = _forceHideCompletion;
+            _forceHideCompletion = nil;
+            if (forceHideCompletion)
+                forceHideCompletion();
+        }];
     }
 }
 
--- a/Sources/Controllers/Panels/OAMapPanelViewController.mm
+++ b/Sources/Controllers/Panels/OAMapPanelViewController.mm
@@ -2827,6 +2827,9 @@ typedef enum
         [self restoreFromContextMenuMode];
     
     [self.targetMenuView hide:YES duration:animationDuration onComplete:^{
+        if (onComplete)
+            onComplete();
+
         if (_activeTargetType != OATargetNone)
         {
             if (_activeTargetActive || _activeTargetChildPushed)
@@ -2849,9 +2852,6 @@ typedef enum
         {
             [_hudViewController updateDependentButtonsVisibility];
         }
-
-        if (onComplete)
-            onComplete();
     }];
     
     [_hudViewController updateControlsLayout:YES];
--- a/Sources/Controllers/Panels/OAMapPanelViewController.mm
+++ b/Sources/Controllers/Panels/OAMapPanelViewController.mm
@@ -149,17 +149,6 @@ typedef enum
 
 @property (strong, nonatomic) OARouteInfoView* routeInfoView;
 
-- (void)enqueueContextMenuPresentation:(void (^)(void))presentation;
-- (void)processPendingContextMenuPresentation;
-- (void)presentContextMenuWithPoints:(NSArray<OATargetPoint *> *)targetPoints
-                     selectedObjects:(nullable NSArray<SelectedMapObject *> *)selectedObjects
-                    touchPointLatLon:(nullable CLLocation *)touchPointLatLon;
-- (void)presentContextMenu:(OATargetPoint *)targetPoint selectedObject:(nullable SelectedMapObject *)selectedObject;
-- (void)presentContextMenu:(OATargetPoint *)targetPoint
-                 saveState:(BOOL)saveState
-             preferredZoom:(float)preferredZoom
-            selectedObject:(nullable SelectedMapObject *)selectedObject;
-
 @end
 
 @implementation OAMapPanelViewController
--- a/Sources/Controllers/Panels/OAMapPanelViewController.mm
+++ b/Sources/Controllers/Panels/OAMapPanelViewController.mm
@@ -1422,6 +1422,9 @@ typedef enum
 
 - (void) showContextMenuWithPoints:(NSArray<OATargetPoint *> *)targetPoints selectedObjects:(nullable NSArray<SelectedMapObject *> *)selectedObjects touchPointLatLon:(nullable CLLocation *)touchPointLatLon
 {
+    if (self.isNewContextMenuDisabled)
+        return;
+
     __weak __typeof(self) weakSelf = self;
     [self enqueueContextMenuPresentation:^{
         __strong __typeof(weakSelf) strongSelf = weakSelf;
@@ -1507,6 +1510,9 @@ typedef enum
 
 - (void)showContextMenu:(OATargetPoint *)targetPoint saveState:(BOOL)saveState preferredZoom:(float)preferredZoom
 {
+    if (self.isNewContextMenuDisabled)
+        return;
+
     __weak __typeof(self) weakSelf = self;
     [self enqueueContextMenuPresentation:^{
         __strong __typeof(weakSelf) strongSelf = weakSelf;
@@ -3168,13 +3174,14 @@ typedef enum
                         state:(OATrackMenuViewControllerState *)state
                      analysis:(nullable OASGpxTrackAnalysis *)analysis;
 {
+    BOOL shouldClearNavControllerHistory = _scrollableHudViewController && !state.openedFromTrackMenu;
     __weak __typeof(self) weakSelf = self;
     [self enqueueContextMenuPresentation:^{
         __strong __typeof(weakSelf) strongSelf = weakSelf;
         if (!strongSelf)
             return;
 
-        if (!state.openedFromTrackMenu)
+        if (shouldClearNavControllerHistory)
             state.navControllerHistory = nil;
 
         [strongSelf doShowGpxItem:item items:items routeKey:routeKey state:state trackHudMode:trackHudMode analysis:analysis];
@@ -3196,7 +3203,7 @@ typedef enum
             if (!strongSelf || !strongSelf->_scrollableHudViewController)
                 return NO;
 
-            [strongSelf->_scrollableHudViewController forceHideWithCompletion:completion];
+            [strongSelf->_scrollableHudViewController hide:YES duration:0.2 onComplete:completion];
             return YES;
         },
         ^BOOL(dispatch_block_t completion) {
@@ -3212,7 +3219,8 @@ typedef enum
             if (!strongSelf || !strongSelf.targetMenuView.superview)
                 return NO;
 
-            [strongSelf hideTargetPointMenu:0.2 onComplete:completion];
+            strongSelf->_prevScrollableHudViewController = nil;
+            [strongSelf hideTargetPointMenu:0.2 onComplete:completion hideActiveTarget:YES mapGestureAction:NO];
             return YES;
         },
     ]];
--- a/Sources/Controllers/Panels/OAMapPanelViewController.mm
+++ b/Sources/Controllers/Panels/OAMapPanelViewController.mm
@@ -201,7 +201,6 @@ typedef enum
 
     MBProgressHUD *_gpxProgress;
     ContextMenuPresentationCoordinator *_contextMenuPresentationCoordinator;
-    BOOL _contextMenuPresentationUITestFixtureStarted;
     UIView *_contextMenuPresentationUITestStateView;
 }
 
@@ -287,7 +286,7 @@ typedef enum
 
     // Setup target point menu
     self.targetMenuView = [[OATargetPointView alloc] initWithFrame:CGRectMake(0.0, 0.0, DeviceScreenWidth, DeviceScreenHeight)];
-    self.targetMenuView.accessibilityIdentifier = @"context_menu_container";
+    self.targetMenuView.accessibilityIdentifier = UITestAccessibilityIdentifier.contextMenuContainer;
     self.targetMenuView.isAccessibilityElement = [self isContextMenuPresentationUITestingEnabled];
     self.targetMenuView.menuViewDelegate = self;
     [self.targetMenuView setMapViewInstance:_mapViewController.view];
@@ -298,7 +297,7 @@ typedef enum
 
     // Setup target multi menu
     self.targetMultiMenuView = [[OATargetMultiView alloc] initWithFrame:CGRectMake(0.0, 0.0, DeviceScreenWidth, 140.0)];
-    self.targetMultiMenuView.accessibilityIdentifier = @"multi_context_menu_container";
+    self.targetMultiMenuView.accessibilityIdentifier = UITestAccessibilityIdentifier.multiContextMenuContainer;
     self.targetMultiMenuView.isAccessibilityElement = [self isContextMenuPresentationUITestingEnabled];
 
     [self setupContextMenuPresentationUITestStateViewIfNeeded];
@@ -327,12 +326,13 @@ typedef enum
     BOOL isCarPlayConnected = UIApplication.sharedApplication.isCarPlayConnected;
     if ([_mapViewController parentViewController] != self && !isCarPlayConnected)
         [self doMapRestore];
-    
+
     if (isCarPlayConnected)
         [self onCarPlayConnected];
-    
+
+    [self runUITestsIfNeeded];
+
     [[OADiscountHelper instance] checkAndDisplay];
-    [self runContextMenuPresentationUITestFixtureIfNeeded];
 }
 
 - (void) viewWillLayoutSubviews
@@ -4733,9 +4733,20 @@ typedef enum
 
 #pragma mark - UI Tests
 
+- (void)runUITestsIfNeeded
+{
+    [self runContextMenuPresentationUITestFixtureIfNeeded];
+    [self runGpxWaypointOpenTrackUITestFixtureIfNeeded];
+}
+
 - (BOOL)isContextMenuPresentationUITestingEnabled
 {
-    return [[[NSProcessInfo processInfo] arguments] containsObject:@"-ui-testing-context-menu-presentation-race"];
+    return UITestState.isContextMenuPresentationRaceEnabled;
+}
+
+- (BOOL)isGpxWaypointOpenTrackUITestingEnabled
+{
+    return UITestState.isGpxWaypointOpenTrackEnabled;
 }
 
 - (void)setupContextMenuPresentationUITestStateViewIfNeeded
@@ -4746,7 +4757,7 @@ typedef enum
     _contextMenuPresentationUITestStateView = [[UIButton alloc] initWithFrame:CGRectMake(0.0, 0.0, 44.0, 44.0)];
     _contextMenuPresentationUITestStateView.backgroundColor = UIColor.clearColor;
     _contextMenuPresentationUITestStateView.isAccessibilityElement = YES;
-    _contextMenuPresentationUITestStateView.accessibilityIdentifier = @"context_menu_presentation_test_state";
+    _contextMenuPresentationUITestStateView.accessibilityIdentifier = UITestAccessibilityIdentifier.contextMenuPresentationState;
     _contextMenuPresentationUITestStateView.accessibilityLabel = @"idle";
     _contextMenuPresentationUITestStateView.accessibilityValue = @"idle";
     [self.view addSubview:_contextMenuPresentationUITestStateView];
@@ -4771,10 +4782,8 @@ typedef enum
 
 - (void)runContextMenuPresentationUITestFixtureIfNeeded
 {
-    if (![self isContextMenuPresentationUITestingEnabled] || _contextMenuPresentationUITestFixtureStarted)
+    if (![UITestState shouldRunContextMenuPresentationRaceFixture])
         return;
-
-    _contextMenuPresentationUITestFixtureStarted = YES;
 
     __weak __typeof(self) weakSelf = self;
     dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.5 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
@@ -4841,6 +4850,46 @@ typedef enum
     [self.view bringSubviewToFront:_contextMenuPresentationUITestStateView];
 
     [self showTargetPointMenu:NO showFullMenu:NO onComplete:onComplete];
+}
+
+- (void)runGpxWaypointOpenTrackUITestFixtureIfNeeded
+{
+    if (![UITestState shouldRunGpxWaypointOpenTrackFixture])
+        return;
+
+    __weak __typeof(self) weakSelf = self;
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.5 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+        __strong __typeof(weakSelf) strongSelf = weakSelf;
+        if (!strongSelf)
+            return;
+
+        OASGpxFile *currentTrack = [OASavingTrackHelper sharedInstance].currentTrack;
+        NSString *waypointName = @"UITest Waypoint A";
+        BOOL hasFixtureWaypoint = NO;
+        for (OASWptPt *point in [currentTrack getPointsList])
+        {
+            if ([point.name isEqualToString:waypointName])
+            {
+                hasFixtureWaypoint = YES;
+                break;
+            }
+        }
+
+        if (!hasFixtureWaypoint)
+        {
+            OASWptPt *waypoint = [[OASWptPt alloc] initWithLat:52.379189 lon:4.899431];
+            waypoint.name = waypointName;
+            waypoint.category = @"UITest";
+            [currentTrack addPointPoint:waypoint];
+            [currentTrack processPoints];
+        }
+
+        OASTrackItem *trackItem = [[OASTrackItem alloc] initWithGpxFile:currentTrack];
+        [strongSelf openTargetViewWithGPX:trackItem
+                              selectedTab:EOATrackMenuHudPointsTab
+                     selectedStatisticsTab:EOATrackMenuHudSegmentsStatisticsOverviewTab
+                            openedFromMap:NO];
+    });
 }
 
 @end
--- a/Sources/Controllers/Panels/OAMapPanelViewController.mm
+++ b/Sources/Controllers/Panels/OAMapPanelViewController.mm
@@ -3214,7 +3214,7 @@ typedef enum
             if (!strongSelf || !strongSelf->_scrollableHudViewController)
                 return NO;
 
-            [strongSelf->_scrollableHudViewController hide:YES duration:0.2 onComplete:completion];
+            [strongSelf->_scrollableHudViewController forceHideWithCompletion:completion];
             return YES;
         }],
         [[ContextMenuDismissHandler alloc] initWithHandler:^BOOL(dispatch_block_t completion) {
--- a/Sources/Controllers/Panels/OAMapPanelViewController.mm
+++ b/Sources/Controllers/Panels/OAMapPanelViewController.mm
@@ -200,8 +200,9 @@ typedef enum
     BOOL _isNewContextMenuStillEnabled;
 
     MBProgressHUD *_gpxProgress;
-    BOOL _contextMenuTransitionInProgress;
-    void (^_pendingContextMenuPresentation)(void);
+    ContextMenuPresentationCoordinator *_contextMenuPresentationCoordinator;
+    BOOL _contextMenuPresentationUITestFixtureStarted;
+    UIView *_contextMenuPresentationUITestStateView;
 }
 
 - (instancetype) init
@@ -226,6 +227,7 @@ typedef enum
     _destinationsHelper = [OADestinationsHelper instance];
     _mapWidgetRegistry = [OAMapWidgetRegistry sharedInstance];
     _weatherToolbarStateChangeObservable = [[OAObservable alloc] init];
+    _contextMenuPresentationCoordinator = [ContextMenuPresentationCoordinator new];
 
     _addonsSwitchObserver = [[OAAutoObserverProxy alloc] initWith:self
                                                       withHandler:@selector(onAddonsSwitch:withKey:andValue:)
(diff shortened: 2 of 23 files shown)


Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3136,17 +3157,78 @@ - (void)openTargetViewWithGPX:(OASTrackItem *)item
...
                      analysis:(nullable OASGpxTrackAnalysis *)analysis;
 {
+    __weak __typeof(self) weakSelf = self;
+    [self enqueueContextMenuPresentation:^{
+        __strong __typeof(weakSelf) strongSelf = weakSelf;
+        if (!strongSelf)
+            return;
+
+        if (!state.openedFromTrackMenu)
+            state.navControllerHistory = nil;
```
alex-dev-neo, line 3196: **`navControllerHistory` is now cleared unconditionally.**

Previously this line lived inside the `if (_scrollableHudViewController)` branch; when no HUD was open, `doShowGpxItem` was called directly and the history survived. Now it runs on every presentation where `openedFromTrackMenu` is false.

`openTargetViewWithGPXFromTracksList:navControllerHistory:fromTrackMenu:` is called with `fromTrackMenu: false` and a non-empty history from `MyPlacesNavigator.swift:35`, `TracksViewController.swift:2423`, `SavedArticlesTabViewController.swift:256`, `TravelExploreViewController.swift:684` and `MapSettingsGpxViewController.swift:457` — in all of those cases no scrollable HUD is open.

Repro: My Places → Tracks → tap a track. The history is nil'd here, so `OATrackMenuHudViewController.mm:217` never populates `_navControllerHistory` and the restore at `OATrackMenuHudViewController.mm:443` is skipped. Closing the track menu drops the user on the bare map instead of back on the Tracks list.

This is the opposite of the PR description's *"Preserved the existing track navigation-history behavior for menus opened outside another track menu."* Suggest keeping the clear conditional on a HUD actually having been dismissed.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3136,17 +3157,78 @@ - (void)openTargetViewWithGPX:(OASTrackItem *)item
...
+            [strongSelf processPendingContextMenuPresentation];
+        }];
+        return;
+    }
+
+    if (self.targetMenuView.superview)
+    {
+        _contextMenuTransitionInProgress = YES;
+        __weak __typeof(self) weakSelf = self;
+        [self hideTargetPointMenu:0.2 onComplete:^{
```
alex-dev-neo: **`restoreActiveTargetMenu` re-enters the queue from inside the completion you're awaiting, clobbering the pending presentation.**

`hideTargetPointMenu:`'s completion runs `restoreActiveTargetMenu` (line 2811), which for `_activeTargetType == OATargetGPX` calls the **public** `openTargetViewWithGPX:` (line 2062) — which now goes through `enqueueContextMenuPresentation` and overwrites `_pendingContextMenuPresentation`.

Sequence with a track menu as the active target and a POI menu on screen, user taps track B:
1. `enqueueContextMenuPresentation(B)` stores B.
2. `targetMenuView.superview` branch → `_contextMenuTransitionInProgress = YES`, `hideTargetPointMenu:0.2`.
3. Inside `[self.targetMenuView hide:...]`'s completion, `restoreActiveTargetMenu` → `openTargetViewWithGPX(_activeTargetObj)` → **`_pendingContextMenuPresentation` is replaced with A (the old track)**. Its nested `processPendingContextMenuPresentation` returns immediately because the flag is still YES.
4. The coordinator's own `onComplete` (line 2825) then clears the flag and re-processes → **runs A, not B**.

Net: tapping a different track re-opens the previous track's menu — the same class of bug this PR is fixing. The single-slot last-write-wins queue is what makes this destructive rather than merely redundant.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3136,17 +3157,78 @@ - (void)openTargetViewWithGPX:(OASTrackItem *)item
...
+    _pendingContextMenuPresentation = [presentation copy];
+    [self processPendingContextMenuPresentation];
+}
+
+- (void)processPendingContextMenuPresentation
+{
+    if (_contextMenuTransitionInProgress || !_pendingContextMenuPresentation)
+        return;
+
     if (_scrollableHudViewController)
```
alex-dev-neo: **The coordinator tears down the visible HUD *before* the presentation gets to run its `isNewContextMenuDisabled` guard.**

`isNewContextMenuDisabled` (line 1489) returns YES for `OATargetRoutePlanning`, `OATargetRouteLineAppearance`, `OATargetWeatherLayerSettings`, `OATargetTerrainParametersSettings`, `OATargetMapModeParametersSettings` and others — all presented as `_scrollableHudViewController` (`PlanRouteScrollableViewController`, `OARouteLineAppearanceHudViewController`, `OAWeatherLayerSettingsViewController`, `OAMapSettingsTerrainParametersViewController`, `MapSettingsMapModeParametersViewController`).

Before this PR the guard was the *first* thing `showContextMenuWithPoints:` / `showContextMenu:saveState:` did, so a map tap in those modes was a no-op. Now the guard sits inside `presentContextMenuWithPoints:` (line 1426) and `presentContextMenu:saveState:` (line 1511), which only run **after** this `forceHideWithCompletion:`.

So: tap the map while Plan Route / Terrain parameters / Weather settings / Route line appearance is open → the sheet is fully torn down (losing in-progress route points or unsaved parameter edits) and then the presentation bails out, leaving nothing in its place.

It's worse than a dismissal, because `forceHide` routes through each controller's `hide` — the *user-pressed-back* path:
- `MapSettingsMapModeParametersViewController.swift:58` → `mapPanel.mapSettingsButtonClick("")` (re-opens Map Settings)
- `ProfileAppearanceIconSizeViewController.swift:119` → restores a settings nav stack
- `OARouteLineAppearanceHudViewController.mm:527` → pushes Route info + Route preferences

An unrelated screen pops open on a plain map tap. The disabled-check should be evaluated before anything is dismissed.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3136,17 +3157,78 @@ - (void)openTargetViewWithGPX:(OASTrackItem *)item
...
     {
-        [_scrollableHudViewController hide:YES duration:0.2 onComplete:^{
-            if (!state.openedFromTrackMenu)
-                state.navControllerHistory = nil;
+        _contextMenuTransitionInProgress = YES;
+        __weak __typeof(self) weakSelf = self;
+        [_scrollableHudViewController forceHideWithCompletion:^{
```
alex-dev-neo: **`forceHide` is not equivalent to the `hide:duration:onComplete:` this replaced — it collapses the navigation stack.**

The old code used `[_scrollableHudViewController hide:YES duration:0.2 onComplete:]`, leaving `_forceHiding` NO, so `restoreNavControllerHistoryIfNeeded` (`OATrackMenuHudViewController.mm:435`, called from `viewWillDisappear`) took the `openedFromTracksList` branch and did `setViewControllers:_navControllerHistory`.

`forceHideWithCompletion:` → `[self forceHide]` → `OATrackMenuHudViewController.mm:417` sets `_forceHiding = YES` → `restoreForceHidingScrollableHud()` (`UIViewController+Extension.swift:150`) does `setViewControllers([rootViewControllerToKeep], animated: true)`, discarding the entire pushed stack.

Repro: open a track from My Places, then tap a second track on the map — the My Places stack is destroyed rather than preserved. Same divergence in `PlanRouteScrollableViewController.swift:313`, where `isForceHiding` now diverts `restorePreviousScreenIfNeeded` away from `restoreTrackMenu` / `setViewControllers(navControllerHistory)`.

Combined with the `state.navControllerHistory = nil` change at line 3167, the back-stack is lost twice over.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3136,17 +3157,78 @@ - (void)openTargetViewWithGPX:(OASTrackItem *)item
...
+    }];
+}
+
+- (void)enqueueContextMenuPresentation:(void (^)(void))presentation
+{
+    _pendingContextMenuPresentation = [presentation copy];
+    [self processPendingContextMenuPresentation];
+}
+
+- (void)processPendingContextMenuPresentation
```
alex-dev-neo, line 3208: **No failsafe for `_contextMenuTransitionInProgress`: one dropped completion permanently disables all context menus.**

The flag is cleared *only* from the three dismissal completion blocks, and each of those APIs has a path that silently drops its completion:

- `hideTargetPointMenu:onComplete:hideActiveTarget:mapGestureAction:` (line 2770) returns without calling `onComplete` when `![self.targetMenuView preHide]`, and again when `self.targetMultiMenuView.superview` is set.
- `OATargetPointView.hide:duration:onComplete:` (`OATargetPointView.mm:1005`) drops `onComplete` entirely in its `else { _hiding = NO; }` branch when `superview` is nil.
- `OATargetMultiView.hide:duration:onComplete:` (`OATargetMultiView.mm:236`) is a complete no-op, completion included, when `superview` is nil.

The `superview` checks above each branch cover these today, and `preHide` currently has no overrides so it always returns YES — but the invariant is undocumented and one override away from breaking. Once the flag sticks at YES, **every** subsequent tap on a POI, track or multi-object area does nothing until the app restarts, with no log and no recovery.

Worth adding a watchdog (reset on `viewDidDisappear`, or a timeout that logs and clears), given the failure mode is a fully dead map UI.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -2791,9 +2812,6 @@ - (void) hideTargetPointMenu:(CGFloat)animationDuration onComplete:(void (^)(voi
             }
         }
         
-        if (onComplete)
-            onComplete();
-
         if (_prevScrollableHudViewController)
```
alex-dev-neo, line 2846: **The previous HUD is restored here and then immediately force-hidden by the coordinator.**

`showTargetPointMenu:` (line 2526) stashes any open scrollable HUD into `_prevScrollableHudViewController` whenever a target menu goes on top. So with a track menu underneath a POI menu, `_scrollableHudViewController` is nil and `processPendingContextMenuPresentation` takes the `targetMenuView.superview` branch.

That branch's completion runs *this* line first — re-mounting the old track menu — and only then clears the flag and re-enters `processPendingContextMenuPresentation`, which now finds `_scrollableHudViewController` non-nil and force-hides it.

Result: the previous track menu visibly flashes in and out before the new menu appears, and for non-track HUDs in that slot it runs their full back-button teardown (see the comment on line 3184).

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -2803,6 +2821,9 @@ - (void) hideTargetPointMenu:(CGFloat)animationDuration onComplete:(void (^)(voi
         {
             [_hudViewController updateDependentButtonsVisibility];
         }
+
+        if (onComplete)
```
alex-dev-neo: **Moving `onComplete()` after the restore blocks changes ordering for every existing caller, not just the new coordinator.**

`showRouteInfo` (line 1117) does `[self hideTargetPointMenu:.2 onComplete:^{ [self showRouteInfoInternal:fullScreen]; }]`. Previously the route-info panel was built first and the restores followed. Now `restoreActiveTargetMenu` (which for `OATargetGPX` re-opens an entire track menu through the presentation queue) and `showScrollableHudViewController:_prevScrollableHudViewController` run first, leaving a restored HUD mounted underneath the route-info panel.

`showTargetPointMenu:` line 2540 (the `hideActiveTarget:YES` restart path) is affected the same way: the previous HUD is re-shown, then the immediately following `showTargetPointMenu:` stashes and hides it again.

The reorder is only needed for the coordinator's own call at line 3218 — consider scoping it there (e.g. a separate completion parameter) rather than changing the shared method's contract.

Sources/Controllers/TargetMenu/Transport/OATransportRouteController.mm
```diff
@@ -291,8 +291,9 @@ + (void) hideToolbar
     toolbarController = nil;
 }
 
-- (void) onMenuSwipedOff
+- (void)onMenuDismissed
```
alex-dev-neo, line 294: **`onMenuDismissed` doesn't cover "every dismissal path" — it's skipped on the non-animated hide.**

`OATargetPointView.hide:animated:duration:onComplete:` (`OATargetPointView.mm:935`) calls `[self clearCustomControllerIfNeeded]` at line 997 in the `duration <= 0` branch, and `clearCustomControllerIfNeeded` (line 652) sets `_customController = nil`. The trailing `if (self.customController) [self.customController onMenuDismissed];` at line 1010 is then skipped entirely.

`targetHideMenu:0 backButtonClicked:YES` (`OAGPXWptViewController.mm:265`) already reaches a duration-0 hide, so this branch is live. With the `onMenuSwipedOff` override deleted, that path now leaves the transport toolbar and the `transportStopsLayer` route on screen with nothing to remove them.

Two further behavior changes worth checking:
- On the animated branch `onMenuDismissed` fires **synchronously at the start** of the hide, whereas `onMenuSwipedOff` ran in the completion — the toolbar now disappears before the menu finishes sliding away.
- It also fires on hides that immediately re-show the menu (`OAMapPanelViewController.mm:2540` restart path), stripping the toolbar and route line from a menu that is about to come back.

alex-dev-neo (commented):
Automated review of the context-menu presentation coordinator. 15 findings, ranked most severe first in the inline comments.

The recurring theme in the top findings is that the coordinator sits *above* the guards it needs to respect: `processPendingContextMenuPresentation` tears down whatever is on screen before the presentation block checks `isNewContextMenuDisabled`, before it knows whether `restoreActiveTargetMenu` will re-enter the queue, and without distinguishing `forceHide` (which trips `_forceHiding` and collapses the nav stack) from the `hide:` the old code used.

Highest-impact items:
1. `OAMapPanelViewController.mm:3167` — `state.navControllerHistory` is now cleared unconditionally, which inverts the behavior the PR description says it preserves.
2. `OAMapPanelViewController.mm:3218` — `restoreActiveTargetMenu` overwrites the pending presentation from inside the completion being awaited, so the *previous* track menu re-opens.
3. `OAMapPanelViewController.mm:3184` — map taps now destroy Plan Route / Terrain / Weather / Route-line-appearance sheets that `isNewContextMenuDisabled` used to protect.

Findings were derived by reading the diff together with the enclosing call graph; the runtime scenarios are reasoned from the code, not observed on device, so they're worth confirming against the repro steps in each comment.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3158,8 +3240,7 @@ - (void)doShowGpxItem:(OASTrackItem *)item
 {
     BOOL showCurrentTrack = item.isShowCurrentTrack;
 
-    [self hideMultiMenuIfNeeded];
-    [self hideTargetPointMenu];
+    [_mapViewController hidePolygonHighlight];
```
alex-dev-neo, line 3249: **The polygon-highlight clear is a leaf special case; the multi-menu path still shows a stale highlight.**

`setSelectedObject:` (line 1578) highlights via `contextMenuLayer highlightPolygon:` for POI/details objects. The clear now exists in exactly two places: `presentContextMenu:saveState:` line 1539, and here.

`presentContextMenuWithPoints:` line 1471 routes multi-object taps to `showMultiContextMenu:points:selectedObjects:`, which never clears it — so tapping a POI polygon and then an area with several selectable objects leaves the old highlight painted under the multi-menu. That's the same symptom listed in the PR description, just on a different path.

Since every presentation now funnels through `processPendingContextMenuPresentation`, that's the one place this belongs — it would also cover any future presentation path for free.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3136,17 +3157,78 @@ - (void)openTargetViewWithGPX:(OASTrackItem *)item
...
+        if (!state.openedFromTrackMenu)
+            state.navControllerHistory = nil;
+
+        [strongSelf doShowGpxItem:item items:items routeKey:routeKey state:state trackHudMode:trackHudMode analysis:analysis];
+    }];
+}
+
+- (void)enqueueContextMenuPresentation:(void (^)(void))presentation
+{
+    _pendingContextMenuPresentation = [presentation copy];
```
alex-dev-neo: **Unconditional overwrite: queued presentations are silently discarded.**

The queue is one slot with last-write-wins semantics and no notification to the dropped caller. For a rapid double-tap that's the intent, but programmatic presentations share the same door: a search result (`OAQuickSearchTableController.mm:538`), a deep link, `OASaveTrackBottomSheetViewController.mm:101`, `OAEditPointViewController.mm:975`. Any of those landing inside a 0.2–0.3 s transition window vanishes, and the app just does nothing in response to an explicit navigation action.

This is also what turns the `restoreActiveTargetMenu` re-entrancy (see line 3218) from redundant work into the wrong menu opening.

Sources/Controllers/Base/OABaseScrollableHudViewController.mm
```diff
@@ -488,6 +489,12 @@ - (void)forceHide
     [self hide];
 }
 
+- (void)forceHideWithCompletion:(void (^)(void))onComplete
```
alex-dev-neo, line 492: **Single-slot `_forceHideCompletion` with no in-flight check.**

Two issues:

1. A second `forceHideWithCompletion:` before the first completes overwrites `_forceHideCompletion`, so the first block is never invoked.
2. `hide:duration:onComplete:` has no `_isHiding` re-entrancy guard — it just starts another animation. If the HUD is already mid-hide (user swiped it down, or `hide` was called elsewhere) when the coordinator calls `forceHideWithCompletion:`, the **first** hide's `dismissViewControllerAnimated:` completion reads and fires the completion belonging to the second call. The coordinator then clears `_contextMenuTransitionInProgress` and presents the new menu while the old HUD is still animating out — reintroducing the overlap this PR is meant to prevent.

Bailing out (or coalescing) when `_forceHideCompletion != nil` would close both.

Sources/Controllers/Base/OABaseScrollableHudViewController.mm
```diff
@@ -502,14 +509,28 @@ - (void) hide:(BOOL)animated duration:(NSTimeInterval)duration onComplete:(void
             [_mapPanel hideScrollableHudViewController];
             _scrollableView.frame = frame;
         } completion:^(BOOL finished) {
-            [self dismissViewControllerAnimated:NO completion:onComplete];
+            [self dismissViewControllerAnimated:NO completion:^{
```
alex-dev-neo, line 512: **The completion block is duplicated verbatim in both branches.**

Lines 512–519 and 526–533 are character-for-character identical. Any future fix — nil-ing `_forceHideCompletion` before invoking it, adding the re-entrancy guard — has to be made twice and will drift. Hoisting it into a local built once before `if (animated)` and passing it to both `dismissViewControllerAnimated:completion:` calls removes the duplication.

More broadly, the `_forceHideCompletion` ivar looks avoidable: `forceHide` could simply take the completion and thread it into `hide:duration:onComplete:` — which is exactly what `PlanRouteScrollableViewController.swift:268` already does.

Sources/Controllers/Base/OABaseScrollableHudViewController.mm
```diff
@@ -35,6 +35,7 @@ @implementation OABaseScrollableHudViewController
     
     BOOL _isDragging;
     BOOL _isHiding;
+    void (^_forceHideCompletion)(void);
```
alex-dev-neo, line 38: **Long-lived ivar retaining a closure and everything it captured.**

`_forceHideCompletion` and `_pendingContextMenuPresentation` both hold blocks that strongly capture their arguments. The enqueued blocks at `OAMapPanelViewController.mm:3161` capture `item`, `items`, `routeKey`, `state` and `analysis` (an `OASGpxTrackAnalysis`, which can be sizeable for long tracks); the blocks at lines 1415, 1499 and 1654 capture `targetPoints`, `selectedObjects` and `touchPointLatLon`.

`_pendingContextMenuPresentation` is only nil'd on the success path (line 3228) and never on any abort path, so the last presentation's whole argument graph stays resident on the map panel — a process-lifetime singleton — for as long as the app runs. Nil the slot when a transition is abandoned, or hold the arguments in a small state object instead of a closure.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -3136,17 +3157,78 @@ - (void)openTargetViewWithGPX:(OASTrackItem *)item
...
+    [self processPendingContextMenuPresentation];
+}
+
+- (void)processPendingContextMenuPresentation
+{
+    if (_contextMenuTransitionInProgress || !_pendingContextMenuPresentation)
+        return;
```
alex-dev-neo: **Three copy-paste transition blocks.**

Lines 3184–3196, 3198–3210 and 3212–3224 are the same nine-line shape (set flag → weak/strong dance → dismiss → clear flag → re-process), differing only in the dismissal call. That's ~30 lines where ~12 would do, and it's three places the missing watchdog would have to be added.

A single helper would collapse each branch to one line and give one place to put the failsafe:

```objc
- (void)beginContextMenuTransition:(void (^)(void (^done)(void)))dismiss;
```

Sources/Controllers/PlanRoute/PlanRouteScrollableViewController.swift
```diff
@@ -23,6 +23,18 @@ final class PlanRouteScrollableViewController: OABaseScrollableHudViewController
         .lightContent
     }
 
+    override var currentState: EOADraggableMenuState {
```
alex-dev-neo, line 26: **Unrelated churn in a concurrency fix.**

Of the 23 added / 20 deleted lines in this file, only the four-line `forceHide(completion:)` override relates to the PR. The rest is relocation and reformatting: `currentState` moved here from line 296, `mapViewportBounds` from line 92, `deinit` down to line 1161, `@discardableResult` joined to its declaration, and parentheses dropped from `guard (force || ...)`.

That inflates the diff of a subtle race-condition fix roughly 4x, buries the one functional change, and makes `git blame` on these declarations point at a concurrency commit. Worth splitting into its own commit or dropping.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -1411,9 +1422,19 @@ - (void) showContextMenuWithPoints:(NSArray<OATargetPoint *> *)targetPoints
 
 - (void) showContextMenuWithPoints:(NSArray<OATargetPoint *> *)targetPoints selectedObjects:(nullable NSArray<SelectedMapObject *> *)selectedObjects touchPointLatLon:(nullable CLLocation *)touchPointLatLon
 {
-    if (_activeTargetType == OATargetGPX && _scrollableHudViewController)
-        [_scrollableHudViewController forceHide];
+    if (self.isNewContextMenuDisabled)
```
alex-dev-neo: **Regression: with a track menu open, tapping a POI (or a multi-object area) is now a no-op.**

This guard moved to the top of the entry point in `8463a3429` — before that it lived inside `presentContextMenuWithPoints:`. It correctly closes the earlier problem for Plan Route / Terrain / Weather, but it over-corrects: `isNewContextMenuDisabled` (line 1501) also returns YES for `_activeTargetType == OATargetGPX`, and `doShowGpxItem` sets exactly that at line 3251 and leaves it set for the whole life of the track menu. So this returns before anything is ever enqueued.

The pre-PR code worked *because* of the ordering it removed:

```objc
if (_activeTargetType == OATargetGPX && _scrollableHudViewController)
    [_scrollableHudViewController forceHide];   // ran first
if (self.isNewContextMenuDisabled)
    return;
```

`forceHide` -> `hide:duration:onComplete:` -> `[UIView animateWithDuration:0.3 animations:^{ [_mapPanel hideScrollableHudViewController]; ... }]`. The animations block runs synchronously, so `hideScrollableHudViewController` -> `resetActiveTargetMenu` -> `_activeTargetType = OATargetNone` had already happened by the time the guard was evaluated.

Repro: open any track menu, then tap a POI / favorite / waypoint on the map. Nothing happens. Same for tapping an area with several selectable objects. Map taps still reach here (`OAMapViewController.mm:1777` -> `OAContextMenuLayer.mm:566` / `:618`); `enterContextMenuMode` only affects the HUD, not gesture delivery.

This is steps 3-4 of the reproduction scenario in the PR description.

Suggestion: exclude `OATargetGPX` from the early check and let the coordinator's first dismiss handler take the GPX HUD down — that handler already does the equivalent, non-destructive `hide:` instead of `forceHide`.

Sources/Controllers/Panels/OAMapPanelViewController.mm
```diff
@@ -1637,6 +1665,16 @@ - (void) showContextMenu:(OATargetPoint *)targetPoint
 }
 
 - (void) showContextMenu:(OATargetPoint *)targetPoint selectedObject:(SelectedMapObject *)selectedObject
+{
+    __weak __typeof(self) weakSelf = self;
+    [self enqueueContextMenuPresentation:^{
```
alex-dev-neo, line 1678: **This entry point still enqueues without the guard, so a long press now tears down Plan Route / Terrain / Weather / Route-line-appearance sheets.**

`8463a3429` added the `isNewContextMenuDisabled` check to `showContextMenuWithPoints:` (line 1425) and `showContextMenu:saveState:preferredZoom:` (line 1513) but not here, so this path still dismisses first and checks afterwards.

Path: long press on empty map -> `OAContextMenuLayer.mm:533` `[mapPanel showContextMenu:unknownTargetPoint]` -> `showContextMenu:selectedObject:` -> `enqueueContextMenuPresentation` (no guard) -> dismiss handler 1 hides `_scrollableHudViewController` -> `hideScrollableHudViewController` resets `_activeTargetType` to `OATargetNone` -> the guard inside `presentContextMenu:saveState:preferredZoom:selectedObject:` (line 1521) now passes -> the menu is presented.

Before this PR, `showContextMenu:saveState:preferredZoom:selectedObject:` hit the guard and returned, and the sheet survived.

Impact: long-pressing the map while Plan Route is open dismisses the route-planning sheet along with its in-progress points; same for the terrain / weather / route-line-appearance parameter sheets. Note that `routePlanningLayer onMapPointSelected:` (`OAMapViewController.mm:1779`) still runs immediately after, so in Plan Route mode the user both adds a point and loses the sheet.

Together with line 1425, the guard is now evaluated in three different places with three different behaviours. One evaluation, in `enqueueContextMenuPresentation`, with `OATargetGPX` excluded, would cover all of them consistently.

Sources/Controllers/Base/OABaseScrollableHudViewController.mm
alex-dev-neo, line 492: **`forceHideWithCompletion:` no longer has any callers — this whole addition is now dead code.**

It was live through `38370a810`, where `processPendingContextMenuPresentation` called `[strongSelf->_scrollableHudViewController forceHideWithCompletion:completion]`. `8463a3429` switched that to `hide:YES duration:0.2 onComplete:completion` — which is the right call, since `forceHide` trips `_forceHiding` and `restoreForceHidingScrollableHud` collapses the pushed nav stack — but the plumbing added for the old approach was left behind.

`grep -rn forceHideWithCompletion Sources/` now returns only the declaration and this definition. Unreachable as a result:

- the `_forceHideCompletion` ivar (line 38);
- the two verbatim-identical completion blocks at lines 512-519 and 526-533 in `hide:duration:onComplete:`;
- `- (void)forceHideWithCompletion:` in the header (line 44);
- `PlanRouteScrollableViewController.forceHide(completion:)` (line 268) — which is also the only PR-related change in that file's 43 changed lines.

The PR description's "Added completion support for force-closing scrollable HUD controllers" now describes code nothing calls. Suggest dropping it, along with the `PlanRouteScrollableViewController` override and the reordering churn that came with it.

alex-dev-neo (commented):
Follow-up review of the two grooming rounds since 29fac9120. Six of the earlier findings are addressed; three items remain, two of them newly introduced in 8463a3429.

Findings 1 and 2 are two halves of the same problem: `isNewContextMenuDisabled` is now evaluated in three places with three different behaviours. Hoisting it in `showContextMenuWithPoints:` fixed the Plan Route case but re-broke POI taps over a track menu, and `showContextMenu:selectedObject:` never got the check at all.

Derived by reading the diff together with the enclosing call graph. I was not able to run this on a simulator (`TracksSearchFilter.swift` fails against my local OsmAnd-shared build, unrelated to this PR), so the runtime scenarios are reasoned from the code and worth confirming against the repro steps in each comment.

Sources/Controllers/TargetMenu/Transport/OATransportRouteController.mm
alex-dev-neo, line 294: **`onMenuDismissed` does not cover "every dismissal path" — it is skipped entirely on non-animated hides, and on the animated ones it fires at the wrong time.**

The dispatch this now relies on is `OATargetPointView.mm:1010-1011`:

```objc
    if (self.customController)
        [self.customController onMenuDismissed];
```

Three problems with it as the home for the transport cleanup:

**1. Skipped on `duration <= 0` hides.** That branch calls `[self clearCustomControllerIfNeeded]` at `OATargetPointView.mm:997`, and `clearCustomControllerIfNeeded` (line 631) sets `_customController = nil`. By the time line 1010 is reached `self.customController` is nil, so `onMenuDismissed` never runs. `targetHideMenu:0 backButtonClicked:YES` reaches exactly this branch, so with the `onMenuSwipedOff` override removed there is now nothing left to take down the toolbar or the route line — `hideToolbar` and `[transportStopsLayer hideRoute]` simply never fire, and both stay painted on the map.

Note the animated branch nils it in the animation *completion* (line 978), so this asymmetry is easy to miss when reading the method.

**2. Fires at the start of the animated hide, not the end.** Line 1010 sits outside the `if (self.superview)` block, so on the animated path it runs synchronously right after the animation is kicked off. `onMenuSwipedOff` ran inside the completion block (`OATargetPointView.mm:2655-2658`). The transport toolbar and route line now disappear while the menu is still sliding away.

**3. Fires on hides that immediately re-show the same menu.** `showTargetPointMenu:showFullMenu:onComplete:afterComplete:` takes the restart path `[self hideTargetPointMenu:.1 onComplete:^{ [self showTargetPointMenu:...] }]` when `_activeTargetActive` is set (`OAMapPanelViewController.mm:2546`). That is an animated hide, so `onMenuDismissed` fires, strips the toolbar and the route line, and then the menu comes straight back without them.

Repro for (1): open a transport route context menu, then trigger any `targetHideMenu:0` dismissal. The toolbar and the highlighted route remain on the map with no way to clear them short of another dismissal that happens to be animated.

`onMenuDismissed` is the right hook conceptually — but the dispatch at line 1010 needs to move inside both branches, after `restoreTargetType` and before `clearCustomControllerIfNeeded` nils the controller (or capture the controller into a local before the clear, the way `OATargetPointView.mm:2654` already does with `OATargetMenuViewController __block *customController = self.customController;`). Otherwise the base-class `restoreViewport` in `OATargetMenuViewController.mm:1166` is being skipped on that path too, along with every other `onMenuDismissed` override.

This is unchanged since my previous review round and is the last item I would hold the PR on — findings on `canPresentNewContextMenu` are addressed in `fd47bd795`, thanks.

alex-dev-neo (commented):
One remaining item. `fd47bd795` resolves the two context-menu guard findings cleanly — `canPresentNewContextMenu` reduces to exactly "blocked unless the reason is a track menu", and all six entry points now share it.

Still open, and the only user-visible defect I have left: the transport cleanup moved onto `onMenuDismissed`, which is skipped on non-animated hides. Detail inline.

(The now-callerless `forceHideWithCompletion:` in `OABaseScrollableHudViewController` is still there too, but that one is dead code rather than a behaviour bug.)

As before, reasoned from the code rather than observed on device.

PR #5715, opened 2026-09-01
Unify coordinate format quick search

### LINKED ISSUE #5512: Unify coordinate format picker across Settings, Search and Widget

vitaliy-sova-ios:


Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -1249,13 +1379,32 @@ - (void) tagHintTapped:(id)sender
...
+{
+    OAApplicationMode *mode = OAAppSettings.sharedManager.applicationMode.get;
+    NSArray<NSString *> *excluded = [OAAppSettings.sharedManager.coordinateFormatSettingsStorage preferredIds:mode];
+    
+    __weak __typeof(self) weakSelf = self;
+    [CoordinateFormatSelectorRouter presentAddFrom:self excludedIds:excluded onSelected:^(NSString *formatId) {
+        __strong __typeof(weakSelf) self = weakSelf;
+        if (!self) return;
+        
+        [OAAppSettings.sharedManager.coordinateFormatSettingsStorage addRecentId:formatId];
```
tigrim: CoordinateFormatSelectorRouter.presentAdd already adds the selected format to recent formats before calling onSelected. Do we need to add it here again? It looks like duplicated responsibility

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -166,17 +171,24 @@ - (void) viewDidLoad
...
     self.tableView.dataSource = self;
+    self.tableView.sectionHeaderTopPadding = 0;
     self.tableView.separatorColor = [SeparatorAppearance color];
 
     self.navigationItem.title = OALocalizedString(@"coords_search");
     
     _toolbarView.frame = CGRectMake(0, DeviceScreenHeight, self.view.frame.size.width, kHintBarHeight);
     
-    _currentFormat = [OAAppSettings.sharedManager.settingGeoFormat get];
+    _currentFormatId = [CoordinateFormatBridge primaryFormatId];
```
tigrim, line 181: Should Quick Search always reset to the profile's primary coordinate format when reopened? Previously the selected geo format came from settings. With the new selector, a user can select a recent/EPSG format, but reopening coordinate search immediately falls back to Primary. Is this the intended UX, or should the last search format be preserved?

Sources/Controllers/Settings/Coordinates/CoordinateFormatBridge.swift
```diff
@@ -7,28 +7,71 @@
 //
 
 import Foundation
+import CoreLocation
+
+@objc enum CoordinateSearchInputMode: Int {
+    case latLon = 0
```
tigrim: case latLon, utm...

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -188,19 +200,21 @@ - (void) configureNavigationBar
...
     UINavigationBarAppearance *blurAppearance = [[UINavigationBarAppearance alloc] init];
+    blurAppearance.shadowColor = nil;
     
     self.navigationController.navigationBar.standardAppearance = blurAppearance;
     self.navigationController.navigationBar.scrollEdgeAppearance = appearance;
     self.navigationController.navigationBar.tintColor = [UIColor colorNamed:ACColorNameIconColorActive];
     self.navigationController.navigationBar.prefersLargeTitles = NO;
     
-    UIBarButtonItem *backButton = [[UIBarButtonItem alloc] initWithTitle:OALocalizedString(@"shared_string_back") style:UIBarButtonItemStylePlain target:self action:@selector(onLeftNavbarButtonPressed)];
+    UIBarButtonItem *backButton = [[UIBarButtonItem alloc] initWithImage:[UIImage systemImageNamed:@"chevron.left"] style:UIBarButtonItemStylePlain target:self action:@selector(onLeftNavbarButtonPressed)];
```
tigrim: ?

Sources/Controllers/Settings/Coordinates/CoordinateFormatBridge.swift
```diff
@@ -46,4 +89,155 @@ final class CoordinateFormatBridge: NSObject {
...
+    static func formatCoordinates(lat: Double, lon: Double, formatId: String?) -> String {
+        let normalized = CoordinateFormatIds.normalize(formatId) ?? CoordinateFormatIds.builtinDdd
+        let format = CoordinateFormatHelper.resolve([normalized]).first
+            ?? BuiltInCoordinateFormat.ddd.toCoordinateFormat()
+        return CoordinateFormatHelper.format(format, lat: lat, lon: lon)
+    }
+
+    static func prefillFields(lat: Double, lon: Double, formatId: String?) -> [String: String] {
+        let info = resolveSearchFormat(formatId)
+        var result: [String: String] = [
```
tigrim: Could we use a typed model for these fields instead of [String: String] with string keys? The current API relies on magic strings such as "easting", "northing" and "maidenhead", so missing or mistyped keys are only detected at runtime. A small CoordinateInputFields struct would make this safer and easier to maintain

Sources/Controllers/Settings/Coordinates/CoordinateFormatBridge.swift
```diff
@@ -46,4 +89,155 @@ final class CoordinateFormatBridge: NSObject {
...
+                              olc: String?,
+                              mgrs: String?,
+                              maidenhead: String?) -> CLLocation? {
+        let info = resolveSearchFormat(formatId)
+        switch info.inputMode {
+        case .latLon:
+            let latV = OALocationConvert.convert(lat ?? "")
+            let lonV = OALocationConvert.convert(lon ?? "")
+            guard !latV.isNaN, !lonV.isNaN else { return nil }
+            return CLLocation(latitude: latV, longitude: lonV)
```
tigrim: https://developer.apple.com/documentation/corelocation/cllocationcoordinate2d

Sources/Controllers/Settings/Coordinates/CoordinateFormatBridge.swift
```diff
@@ -46,4 +89,155 @@ final class CoordinateFormatBridge: NSObject {
...
+                return OAEpsgCoordinateTransformer.sharedInstance()
+                    .toLonLat(withCode: info.epsgCode, easting: e, northing: n)
+            }
+            return nil
+        default:
+            return nil
+        }
+    }
+    
+    private static func parseMetric(_ raw: String?) -> Double {
```
tigrim: private static func parseMetric(_ raw: String?) -> Double? {
    guard let raw else { return nil }

    let value = raw
        .trimmingCharacters(in: .whitespacesAndNewlines)
        .replacingOccurrences(of: " ", with: "")
        .replacingOccurrences(of: ",", with: ".")

    return Double(value)
}

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+import UIKit
+
+@objc protocol CoordinateFormatSelectorDelegate: AnyObject {
+    func coordinateFormatSelector(_ selector: CoordinateFormatSelectorViewController,
+                                  didSelectFormatId formatId: String)
+    func coordinateFormatSelectorDidRequestOtherFormat(_ selector: CoordinateFormatSelectorViewController)
+}
+
+@objcMembers
+final class CoordinateFormatSelectorViewController: OABaseNavbarViewController {
```
tigrim, line 18: Picker?

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+
+    private static let formatIdKey = "formatId"
```
tigrim: enum Key

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+                              delegate: CoordinateFormatSelectorDelegate?) {
+        let vc = CoordinateFormatSelectorViewController(selectedFormatId: selectedFormatId)
+        vc.selectorDelegate = delegate
+        let nav = UINavigationController(rootViewController: vc)
+        nav.modalPresentationStyle = .pageSheet
+        if let sheet = nav.sheetPresentationController {
+            sheet.detents = [.medium(), .large()]
+            sheet.selectedDetentIdentifier = .medium
+            sheet.prefersGrabberVisible = true
+            sheet.preferredCornerRadius = 38
```
tigrim: 20?

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+        fatalError("init(coder:) has not been implemented")
+    }
+    
+    @objc static func present(from presenter: UIViewController,
+                              selectedFormatId: String?,
+                              delegate: CoordinateFormatSelectorDelegate?) {
+        let vc = CoordinateFormatSelectorViewController(selectedFormatId: selectedFormatId)
+        vc.selectorDelegate = delegate
+        let nav = UINavigationController(rootViewController: vc)
+        nav.modalPresentationStyle = .pageSheet
```
tigrim: showMediumToLargeSheetViewController

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+        if showSelectOther {
+            let otherSection = tableData.createNewSection()
+            let row = otherSection.createNewRow()
+            row.cellType = OASimpleTableViewCell.reuseIdentifier
+            row.title = localizedString("coordinate_format_select_other")
+            row.key = Self.selectOtherKey
+        }
+    }
+
+    override func getRow(_ indexPath: IndexPath?) -> UITableViewCell? {
```
tigrim, line 132: voice over?

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -188,19 +200,21 @@ - (void) configureNavigationBar
     UINavigationBarAppearance *appearance = [[UINavigationBarAppearance alloc] init];
     [appearance configureWithOpaqueBackground];
     appearance.backgroundColor = self.tableView.backgroundColor;
-    appearance.shadowColor = [SeparatorAppearance color];
+    appearance.shadowColor = nil;
```
tigrim, line 203: ?

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -1249,13 +1379,32 @@ - (void) tagHintTapped:(id)sender
...
 }
 
+- (void)coordinateFormatSelectorDidRequestOtherFormat:(CoordinateFormatSelectorViewController *)selector
+{
+    OAApplicationMode *mode = OAAppSettings.sharedManager.applicationMode.get;
+    NSArray<NSString *> *excluded = [OAAppSettings.sharedManager.coordinateFormatSettingsStorage preferredIds:mode];
+    
+    __weak __typeof(self) weakSelf = self;
+    [CoordinateFormatSelectorRouter presentAddFrom:self excludedIds:excluded onSelected:^(NSString *formatId) {
+        __strong __typeof(weakSelf) self = weakSelf;
```
tigrim: strongSelf

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
vitaliy-sova-ios: Resetting to “Primary” when reopening the app is intended behavior that matches that of Android and older versions of iOS. The specific format of the most recent search is not saved.

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
vitaliy-sova-ios: Selector — parity with Android and user interface type a list, not a scroll wheel.

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
vitaliy-sova-ios: Figma - 38

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
vitaliy-sova-ios: Figma

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -217,67 +231,89 @@ - (void) updateControllsSectionCells
...
-            @"value" : _mgrsStr,
+            @"value" : _mgrsStr ?: @"",
             @"tag" : @(EOAQuickSearchCoordinatesTextFieldMgrs),
         }];
     }
+    else if (_inputMode == CoordinateSearchInputModeMaidenhead)
+    {
+        [result addObject:@{
+            @"type" : [OAInputTableViewCell getCellIdentifier],
+            @"title" : OALocalizedString(@"navigate_point_format_maidenhead"),
```
tigrim, line 280: Could we use an alphanumeric keyboard for the Maidenhead field? The input-cell configuration only gives OLC and MGRS a non-numeric keyboard, so Maidenhead falls through to `UIKeyboardTypeASCIICapableNumberPad`. Its hint bar is also empty. As a result, users cannot type a locator such as `KO50GK`; they can only paste one or edit the prefilled value using digits.

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -217,67 +231,89 @@ - (void) updateControllsSectionCells
...
+            @"value" : _maidenheadStr ?: @"",
+            @"tag" : @(EOAQuickSearchCoordinatesTextFieldMaidenhead),
+        }];
+    }
+    else if (_inputMode == CoordinateSearchInputModeEastingNorthing)
+    {
+        [result addObject:@{
+            @"type" : [OAInputTableViewCell getCellIdentifier],
+            @"title" : [NSString stringWithFormat:@"%@ / X", OALocalizedString(@"navigate_point_easting")],
```
tigrim, line 289: These fields also use `UIKeyboardTypeASCIICapableNumberPad`, which does not provide a minus sign or decimal separator. That is sufficient for the legacy UTM inputs, but arbitrary projected CRS values can be negative (for example, EPSG:3857 west of Greenwich) and may contain fractions. `parseMetric` accepts both forms, but the UI cannot enter them after a field is cleared. Please choose the keyboard/accessory based on the EPSG input mode.

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
```diff
@@ -0,0 +1,273 @@
...
+    }
+}
+
+@objcMembers
+final class CoordinateFormatSelectorRouter: NSObject {
+    @objc static func presentAdd(from presenter: UIViewController,
+                                 excludedIds: [String],
+                                 onSelected: @escaping (String) -> Void) {
+        let mode = OAAppSettings.sharedManager().applicationMode.get()
+        let addVC = CoordinatesFormatAddViewController(appMode: mode, excludedIds: excludedIds)
```
tigrim: The issue specifies that **Select other format** should open Add Format with the search field focused. This route presents `CoordinatesFormatAddViewController`, but that controller only installs its `UISearchController`; it never activates the search controller or makes the search bar first responder. Consequently the screen opens in the General list with no keyboard instead of directly in search. Could this route opt into an auto-focus mode and activate the search after presentation?

tigrim (commented):
Static review focused on the new coordinate-format input flows. I found three user-facing issues around keyboard selection and the Select Other search flow.

Sources/Controllers/TargetMenu/Search/OAQuickSearchCoordinatesViewController.mm
```diff
@@ -213,7 +213,7 @@ - (void) configureNavigationBar
     self.navigationController.navigationBar.tintColor = [UIColor colorNamed:ACColorNameIconColorActive];
     self.navigationController.navigationBar.prefersLargeTitles = NO;
     
-    UIBarButtonItem *backButton = [[UIBarButtonItem alloc] initWithImage:[UIImage systemImageNamed:@"chevron.left"] style:UIBarButtonItemStylePlain target:self action:@selector(onLeftNavbarButtonPressed)];
+    UIBarButtonItem *backButton = [[UIBarButtonItem alloc] initWithImage:[UIImage systemImageNamed:@"xmark"] style:UIBarButtonItemStylePlain target:self action:@selector(onLeftNavbarButtonPressed)];
```
tigrim: UIBarButtonItem(barButtonSystemItem: .close

tigrim (approved):

PR #5716, opened 2026-09-01
Grouped list fix

### LINKED ISSUE #5667: Fix grouped list
### ToDo
List of places were fix is needed, next part of https://github.com/osmandapp/OsmAnd-Issues/issues/3256

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th></th>
      </tr>
      <tr>
        <td> </td>
      </tr>
    </table>
</details>  

- [x] Simulate navigation

- - - - - -

<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th></th>
      </tr>
      <tr>
        <td> </td>
      </tr>
    </table>
</details>  

- [x] External sensors info

- - - - - -
 
<details>
    <summary>Screenshot</summary>
    <table style="width:100%">
      <tr>
        <th></th>
      </tr>
      <tr>
        <td> </td>
      </tr>
    </table>
</details>  

- [x] Add 16pt padding for separator from the ride side for our base tabell cell

vitaliy-sova-ios:


Sources/Controllers/RoutePlanning/OAOpenAddTrackViewController.mm
```diff
@@ -146,18 +148,17 @@ - (void)generateData
...
     NSArray *gpxList = [NSMutableArray arrayWithArray:[self sortData:filteredData]];
     
-    [existingTracksSection addObject:@{
+    [foldersSection addObject:@{
         @"type" : [OAFoldersCell getCellIdentifier],
         @"selectedValue" : @(_selectedFolderIndex),
         @"values" : [self getFoldersList]
     }];
     
-    [existingTracksSection addObject:@{ @"type" : [OADividerCell getCellIdentifier] }];
```
tigrim, line 159: Why was OADividerCell removed?

Sources/Controllers/Settings/NavigationSettings/OASimulationNavigationSettingViewController.mm
```diff
@@ -132,18 +123,14 @@ - (void)generateData
...
-            @"footerTitle" : selectedModeDescription
-        }];
         [paramsSection addObject:@{
             @"type" : kUICellKey,
             @"key" : [OASimulationMode toKey:EOASimulationModePreview],
             @"title" : [OASimulationMode toTitle:EOASimulationModePreview],
             @"descr" : @"",
-            @"selected" : @([_selectedMode isEqualToString:[OASimulationMode toKey:EOASimulationModePreview]])
+            @"selected" : @([_selectedMode isEqualToString:[OASimulationMode toKey:EOASimulationModePreview]]),
+            @"headerTitle" : OALocalizedString(@"speed_mode"),
```
tigrim, line 132: ?

Sources/Controllers/Settings/OSMEditing/OAOsmAccountSettingsViewController.m
```diff
@@ -163,45 +150,6 @@ - (UITableViewCell *)getRow:(NSIndexPath *)indexPath
         }
         outCell = cell;
     }
-    else if ([type isEqualToString:[OADividerCell getCellIdentifier]])
```
tigrim, line 166: Why was this logic removed?

Sources/Plugins/ExternalSensorsPlugin/BLE/BLEDescription/BLEDescriptionViewController.swift
```diff
@@ -52,35 +52,38 @@ final class BLEDescriptionViewController: OABaseNavbarViewController {
...
         registerObservers()
     }
     
-    override func setupTableHeaderView() {
-        configureHeader()
+    override func hideFirstHeader() -> Bool {
+        true
     }
     
-    private func configureHeader() {
```
tigrim, line 70: Why was the header moved into a cell? What happens when the table is updated frequently but the header itself doesn't need to be updated?

Sources/Controllers/Cells/DescriptionDeviceCell.swift
```diff
@@ -0,0 +1,46 @@
...
+        header.backgroundColor = .clear
+        header.hideBottomDivider()
+        contentView.addSubview(header)
+
+        NSLayoutConstraint.activate([
+            header.topAnchor.constraint(equalTo: contentView.topAnchor),
+            header.leadingAnchor.constraint(equalTo: contentView.leadingAnchor),
+            header.trailingAnchor.constraint(equalTo: contentView.trailingAnchor),
+            header.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
+            header.heightAnchor.constraint(equalToConstant: 156)
```
tigrim: Do we need to keep the header height fixed at `156` here? Since the header is now embedded in a self-sizing table cell, it would be better to let its content constraints determine the height if possible. A fixed height may truncate the content with larger Dynamic Type sizes

Sources/Controllers/RoutePlanning/OAOpenAddTrackViewController.mm
vitaliy-sova-ios: Removed intentionally for insetGrouped. Section bottom spacing comes from the table style now; OADividerCell was only a spacer for the old layout.

tigrim (approved):

Sources/Controllers/Settings/NavigationSettings/OASimulationNavigationSettingViewController.mm
```diff
@@ -132,18 +123,14 @@ - (void)generateData
     {
         BOOL isConstantMode = [_selectedMode isEqualToString:[OASimulationMode toKey:EOASimulationModeConstant]];
         NSMutableArray *paramsSection = [NSMutableArray array];
-        [paramsSection addObject:@{
-            @"type" : [OADividerCell getCellIdentifier],
-            @"inset" : zeroInset,
-            @"headerTitle" : OALocalizedString(@"speed_mode"),
-            @"footerTitle" : selectedModeDescription
-        }];
         [paramsSection addObject:@{
```
alex-dev-neo, line 126: **Row-index fallout: `updateSliderValue:` now reloads the slider instead of the speed row.**

Removing the leading `OADividerCell` shifted every row in `paramsSection` up by one. With Constant mode selected the section is now:

```
0 preview
1 divider
2 constant   <- displays the formatted speed in @"descr"
3 slider
4 divider
5 realistic
```

But `updateSliderValue:` (line 410) still does:

```objc
[self.tableView reloadRowsAtIndexPaths:@[[NSIndexPath indexPathForRow:3 inSection:1]] ...];
```

Row 3 used to be the constant-mode row; it is now the slider cell itself.

**Scenario:** Navigation settings → Simulate navigation → enable → pick "Constant speed", then drag the slider. The reload recreates the slider cell mid-gesture (resetting it to the regenerated value and interrupting the continuous drag), while the "Constant: 50 km/h" label at row 2 — the row that actually renders `_selectedSpeed` — is never refreshed, so the displayed speed stops tracking the slider.

The index should be `2`.

Sources/Controllers/RoutePlanning/OAOpenAddTrackViewController.mm
```diff
@@ -192,7 +193,8 @@ - (void)generateData
             @"key" : @"gpx_route"
         }];
     }
-    [existingTracksSection addObject:@{ @"type" : [OADividerCell getCellIdentifier] }];
+
+    [data addObject:foldersSection];
     [data addObject:existingTracksSection];
```
alex-dev-neo, line 198: **Off-by-one left behind by removing the trailing divider row.**

The `[existingTracksSection addObject:@{ @"type" : [OADividerCell getCellIdentifier] }];` that used to terminate this section is gone, but the last-row test in `getRow:` (line 338) was not updated:

```objc
cell.separatorView.hidden = indexPath.row == [self.tableView numberOfRowsInSection:indexPath.section] - 2;
```

Previously the section ended with the divider, so the last `OAGPXTrackCell` sat at `count - 2` and had its separator hidden. Now the last track is at `count - 1`.

**Scenario:** open "Open existing track" / "Follow track" with two or more tracks. The separator is hidden under the *second-to-last* track and drawn under the last one — a stray separator at the bottom of the list plus a missing one in the middle.

Should be `- 1`.

Sources/Controllers/Cells/DescriptionDeviceCell.swift
```diff
@@ -0,0 +1,45 @@
...
+    override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+    }
+
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+    }
+    
+    func configure(header: DescriptionDeviceHeader) {
+        guard deviceHeader !== header else { return }
```
alex-dev-neo: **The identity guard can permanently strand the header.**

```swift
guard deviceHeader !== header else { return }
```

This bails out on identity alone, without checking the header is still in *this* cell's `contentView`. Both `BLEDescriptionViewController` and `VehicleMetricsDescriptionViewController` hold a single lazy `headerView` and hand the same instance to whichever `DescriptionDeviceCell` is dequeued, and `setupView` re-parents it with `contentView.addSubview(header)`.

**Scenario:** if two `DescriptionDeviceCell` instances are ever alive at once (an animated reload, or an extra self-sizing dequeue while the old cell is still retained), cell B's `setupView` moves the header out of cell A — but `A.deviceHeader` still points at it. When A is later dequeued and reconfigured with the same header, the guard returns immediately, A adds nothing to its `contentView`, and the device-info row renders as an empty ~156pt gap.

Suggest tightening it to also require attachment:

```swift
guard deviceHeader !== header || header.superview !== contentView else { return }
```

alex-dev-neo (commented):
Automated correctness review of the divider-cell → grouped-section refactor. Three findings below (two of them are row-index fallout from removing divider rows). The pbxproj hunks are just Xcode re-sorting — no build entries lost.

PR #5717, opened 2026-09-02
Fix: Unable to download online tiles

### LINKED ISSUE #5710: Online tiles cannot be downloaded — download buttons are misplaced at the top of the screen
## Description
Downloading online tiles is not possible because the download controls are misplaced. Instead of appearing in the toolbar at the bottom of the screen, the buttons are rendered at the top, where they overlap other UI elements and cannot be used.

Related to https://github.com/osmandapp/OsmAnd-iOS/issues/2956

### How to reproduce
* Set any online overlay as the **Map type** in **Configure Map**
* Long tap on the map → **Actions** → **Download map**

### Actual result
<table style="width:100%">
  <tr>
    <td> </td>
  </tr>
</table>   

* The buttons are placed at the top of the screen

### Expected result
* The buttons are shown in the toolbar at the bottom of the screen

**Environment**
OsmAnd Version: 5454
iOS version: 27 Beta

vitaliy-sova-ios:


Sources/DownloadMapSource/OADownloadMapViewController.mm
```diff
@@ -199,6 +199,7 @@ - (void) viewDidLoad
     self.tableView.dataSource = self;
     self.tableView.delegate = self;
     self.bottomToolBarView.backgroundColor = [UIColor colorNamed:ACColorNameGroupBg];
+    self.bottomToolBarView.translatesAutoresizingMaskIntoConstraints = YES;
```
tigrim: Could we fix this in the XIB instead? `bottomToolBarView` uses a fixed frame/autoresizing layout, but `translatesAutoresizingMaskIntoConstraints` is disabled there. It seems cleaner to keep the layout configuration in one place rather than overriding the XIB setting in `viewDidLoad`.

PR #5720, opened 2026-09-03
keep screen on setting

### LINKED ISSUE #1777: [Feature request] Screen turns off
When you turn on the application, the device screen always stays on. It would be nice if the screen stays on when navigation is active and when viewing a map or other actions in the application, the screen turns off after a while.

Vladyslav-lys:


Sources/Helpers/OAAppSettings.m
```diff
@@ -6885,6 +6886,9 @@ - (instancetype) init
         
         _mapScreenOrientation = [OACommonInteger withKey:mapScreenOrientationKey defValue:EOAScreenOrientationSystem];
         [_profilePreferences setObject:_mapScreenOrientation forKey:@"map_screen_orientation"];
+
+        _keepScreenOn = [[OACommonInteger withKey:keepScreenOnKey defValue:EOAKeepScreenOnModeSystemDefault] makeProfile];
```
tigrim: Could we define a profile-specific default or an explicit migration here? With `SystemDefault` as the default, existing installations have no stored value and will start allowing the screen to lock, including during active navigation. Before this PR, OsmAnd disabled the idle timer whenever the app was in the foreground. If the intended behavior from #1777 is system timeout while browsing but keeping the screen awake while navigating, navigation profiles should default to `DuringNavigation` (or existing users should be migrated explicitly).

Sources/Controllers/Settings/GeneralSettings/OAProfileGeneralSettingsViewController.m
```diff
@@ -282,6 +300,13 @@ - (void)generateData
             @"key" : @"screenOrientation"
         }];
     }
+    [appearanceArr addObject:@{
+        @"type" : OAValueTableViewCell.reuseIdentifier,
+        @"title" : OALocalizedString(@"keep_screen_on"),
+        @"value" : keepScreenOnValue,
+        @"icon" : ACImageNameIcCustomMapModeAppThemeFilled,
```
tigrim: This uses the filled App Theme icon for the `Keep screen on` row, so General Settings shows a theme-related glyph for an unrelated setting. Could we use a screen/lock-specific asset (for example, `ic_custom_touch_screen_lock`) or add the intended design asset?

tigrim:
Managing idleTimerDisabled is not the responsibility of OsmAndAppImpl. This logic should be moved to a dedicated service

tigrim:
Could we clarify the expected behavior for the following cases?

1. **CarPlay navigation:** Should `During navigation` keep the connected iPhone screen awake when navigation is active on CarPlay but the main iPhone scene is in the background? The current implementation only treats the main iPhone scene as foreground.
2. **Route simulation:** Should route simulation also keep the screen awake in `During navigation` mode? Currently the decision depends only on `isFollowingMode`.
3. **Copied Browse Map profile:** A copy of the Browse Map profile is no longer equal to `OAApplicationMode.DEFAULT`, so it exposes the `During navigation` option. Is this the intended behavior?
4. **Default value for a copied Browse Map profile:** Should a newly copied profile inherit `System default` from Browse Map, or should it use `During navigation` like navigation profiles?
5. **Meaning of Always with CarPlay:** Should `Always` mean “while OsmAnd is visible on the iPhone,” or should an active CarPlay scene also count? Currently the phone screen is allowed to turn off when the main iPhone scene is not in the foreground.
6. **Navigation states:** Which states should count as active navigation: regular following, paused navigation, route simulation, route preview, following a track, and CarPlay-only navigation?

Vladyslav-lys:
> Could we clarify the expected behavior for the following cases?
> 
> 1. **CarPlay navigation:** Should `During navigation` keep the connected iPhone screen awake when navigation is active on CarPlay but the main iPhone scene is in the background? The current implementation only treats the main iPhone scene as foreground.
> 2. **Route simulation:** Should route simulation also keep the screen awake in `During navigation` mode? Currently the decision depends only on `isFollowingMode`.
> 3. **Copied Browse Map profile:** A copy of the Browse Map profile is no longer equal to `OAApplicationMode.DEFAULT`, so it exposes the `During navigation` option. Is this the intended behavior?
> 4. **Default value for a copied Browse Map profile:** Should a newly copied profile inherit `System default` from Browse Map, or should it use `During navigation` like navigation profiles?
> 5. **Meaning of Always with CarPlay:** Should `Always` mean “while OsmAnd is visible on the iPhone,” or should an active CarPlay scene also count? Currently the phone screen is allowed to turn off when the main iPhone scene is not in the foreground.
> 6. **Navigation states:** Which states should count as active navigation: regular following, paused navigation, route simulation, route preview, following a track, and CarPlay-only navigation?

1. CarPlay navigation: With During navigation, the iPhone screen stays awake while OsmAnd is in the foreground on the iPhone, including when navigation was started from CarPlay. If another app is open on the iPhone, normal auto-lock applies.

2. Route simulation: Active simulation counts as navigation. The implementation now checks the actual simulation state in addition to isFollowingMode, including standalone location simulation.

3. Copied Browse Map profile: During navigation is available when the profile has Navigation settings. The original Browse Map profile does not expose this option; its navigation-capable copies do.

4. Default value: I changed the default to During navigation for navigation-capable profiles, including copies of Browse Map. Profiles without Navigation settings use System default. Explicitly saved or imported values are preserved; resetting the setting applies the new default.

5. Always with CarPlay: Always keeps the screen awake while OsmAnd is in the foreground on the iPhone. An active CarPlay scene alone does not affect this behavior. This setting does not control the vehicle’s display.

6. Navigation states: Regular guidance, active simulation, navigation along a GPX track, and CarPlay navigation count as active navigation. Paused navigation and route preview without Start do not. Pause takes precedence even if simulation is still running.

Sources/Helpers/OAAppSettings.m
```diff
@@ -2095,14 +2096,36 @@ - (void) set:(int)integer mode:(OAApplicationMode *)mode
     [self setValue:@(integer) mode:mode];
 }
 
+- (void)setValue:(NSObject *)value mode:(OAApplicationMode *)mode
```
tigrim: ?

tigrim:
I found two behavior gaps in the current implementation:

1. `migrateKeepScreenOnMode()` writes `.always` for every existing profile. As a result, users upgrading to this version will still have an always-on screen while browsing the map, until they change each profile manually. The linked issue (#1777) asks for the screen to follow the system timeout outside active navigation. Is preserving the old behavior for all existing users intentional? If the new behavior should apply on upgrade, the migration needs to use `SystemDefault` for Browse map and `DuringNavigation` for navigation profiles.

2. `ScreenAwakeService.updateIdleTimer()` treats route animation as active navigation, but selects `settings.applicationMode` whenever `isFollowingMode` is false. Route simulation can run without following mode, so if the current map profile uses `SystemDefault` while the route profile uses `DuringNavigation`, the screen can lock during simulation. Please use the routing helper's profile when route animation is active as well.

tigrim:
I reviewed the current head (14dc6b8) by code inspection, without a build. Two issues remain:

1. `ScreenAwakeService.updateIdleTimer()` checks only `UIApplication.shared.mainScene`. The app supports multiple scenes, but `mainScene` returns the first scene with the default configuration. If that scene is in the background while another main window is active, the service allows the screen to lock during navigation in the visible window. Please check whether any main application scene is foregrounded.

2. `OACommonKeepScreenOnMode` redeclares `get` and `set` with `EOAKeepScreenOnMode` (`NSInteger`), while its `OACommonInteger` superclass declares the same selectors with `int`. These have different widths on 64-bit iOS. Please keep the override signatures compatible with the superclass and convert to the enum at call sites or use a matching base type.

tigrim (approved):

PR #5721, opened 2026-09-03
Follow up my places redesign favorite folder

### LINKED ISSUE #5612: Follow up for My Places redesign

DmitrySvetlichny:


Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+ContextMenu.swift
```diff
@@ -8,57 +8,57 @@
...
         }
         let pinAction = UIAction(title: localizedString(folder.isPinned ? "unpin_folder" : "pin_folder"), image: folder.isPinned ? .icCustomDrawingPinDisable : .icCustomDrawingPin) { [weak self] _ in
-            guard let self else { return }
-            OAFavoritesHelperBridge.shared().setFavoriteGroupPinned(folder.bridgeItem.groupName, pinned: !folder.isPinned)
+            guard let self, let group = folder.group else { return }
+            OAFavoritesHelperBridge.shared().setFavoriteGroupPinned(group.groupName, pinned: !folder.isPinned)
             self.applySnapshot(animatingDifferences: true)
         }
-        let firstButtonsSection = UIMenu(title: "", options: .displayInline, children: [showHideAction, pinAction])
+        let firstButtonsSection = UIMenu(title: "", options: .displayInline, children: folder.group == nil ? [] : [showHideAction, pinAction])
```
tigrim: s it intentional that Show/Hide and Pin actions are unavailable for virtual folders? Also, could we avoid creating an empty inline menu when folder.group == nil?

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+ContextMenu.swift
```diff
@@ -8,57 +8,57 @@
...
 
         let renameAction = UIAction(title: localizedString("shared_string_rename"), image: .icCustomEdit) { [weak self] _ in
             guard let self else { return }
             self.showRenameAlert(for: folder)
```
tigrim: self?.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+ContextMenu.swift
```diff
@@ -8,57 +8,57 @@
...
                 try? FileManager.default.removeItem(at: favoritesUrl)
             })
         }
         let moveAction = UIAction(title: localizedString("shared_string_move"), image: .icCustomFolderMoveOutlined) { [weak self] _ in
             guard let self else { return }
-            self.openFavoriteItemsMove([folder.bridgeItem])
+            self.openFavoriteItemsMove(folderFavoriteItem)
         }
-        let thirdButtons: [UIMenuElement] = (hasFavoritePoints ? [shareAction] : []) + (folder.bridgeItem.groupName.isEmpty ? [] : [moveAction])
+        let thirdButtons: [UIMenuElement] = (hasFavoritePoints ? [shareAction] : []) + (folder.fullPath.isEmpty ? [] : [moveAction])
```
tigrim: Could we simplify this logic? Using multiple ternary expressions and concatenating arrays makes the conditions harder to read. Building thirdButtons with conditional append calls would be clearer and easier to extend

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -292,6 +295,9 @@ - (BOOL)addFavoriteGroup:(NSString *)name
       backgroundIconName:(nullable NSString *)backgroundIconName
 {
     NSString *trimmedName = [(name ?: @"") stringByTrimmingCharactersInSet:NSCharacterSet.whitespaceAndNewlineCharacterSet];
+    if ([trimmedName containsString:@"/"] || [trimmedName containsString:@"_%_"])
```
tigrim: kSubfolderPlaceholder

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -306,31 +312,28 @@ - (BOOL)addFavoriteGroup:(NSString *)name
...
-- (void)renameFavoriteGroup:(NSString *)groupName newName:(NSString *)newName
+- (BOOL)renameFavoriteGroup:(NSString *)groupName newName:(NSString *)newName
 {
     OAFavoriteGroup *group = [self favoriteGroupWithName:groupName];
-    NSString *trimmedName = [newName stringByTrimmingCharactersInSet:NSCharacterSet.whitespaceAndNewlineCharacterSet];
-    if (!group || trimmedName.length == 0)
-        return;
+    NSString *targetGroupName = newName ?: @"";
+    for (NSString *segment in [targetGroupName componentsSeparatedByString:@"/"])
+        if (segment.length == 0 || [segment containsString:@"_%_"])
```
tigrim: kSubfolderPlaceholder

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -415,26 +430,22 @@ - (void)moveFavoriteItems:(NSArray *)favoriteItems toGroupName:(NSString *)targe
         if (movedGroups)
             [OAFavoritesHelper saveCurrentPointsIntoFile];
     }
+
+    return movedFolderPaths.count > 0 || movedPoints;
 }
 
 - (NSArray<NSString *> *)favoriteGroupNamesForMovingFavoriteItems:(NSArray *)favoriteItems
```
tigrim, line 436: Why is NSArray used here without specifying the element type?

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -470,11 +478,11 @@ - (void)changeFavoriteItems:(NSArray *)favoriteItems colorIndex:(NSInteger)color
 
     for (id item in favoriteItems)
     {
-        if (![item isKindOfClass:OAFavoriteFolderBridgeItem.class])
+        if (![item isKindOfClass:NSString.class])
```
tigrim: [NSString class]

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -470,11 +478,11 @@ - (void)changeFavoriteItems:(NSArray *)favoriteItems colorIndex:(NSInteger)color
...
             continue;
 
-        OAFavoriteFolderBridgeItem *folderItem = (OAFavoriteFolderBridgeItem *) item;
-        OAFavoriteGroup *group = [self favoriteGroupWithName:folderItem.groupName];
+        NSString *folderPath = (NSString *) item;
```
tigrim: NSString *)item;

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -537,15 +545,11 @@ - (NSURL *)shareFavoriteItems:(NSArray *)favoriteItems
...
+        if (![item isKindOfClass:NSString.class])
             continue;
 
-        OAFavoriteFolderBridgeItem *folderItem = (OAFavoriteFolderBridgeItem *) item;
-        OAFavoriteGroup *group = [self favoriteGroupWithName:folderItem.groupName];
-        if (!group)
-            continue;
-
-        for (OAFavoriteGroup *groupToShare in [self favoriteGroupsInsideOrEqualToGroupName:group.name])
+        NSString *folderPath = (NSString *) item;
```
tigrim: up

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -1149,6 +1145,47 @@ - (NSString *)lastComponentForGroupName:(NSString *)groupName
...
+        {
+            if (segment.length == 0)
+                continue;
+
+            currentPath = currentPath.length == 0 ? segment : [NSString stringWithFormat:@"%@/%@", currentPath, segment];
+            [folderPaths addObject:currentPath];
+        }
+    }
+
+    return folderPaths.copy;
```
tigrim: olderPaths copy];

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -1149,6 +1145,47 @@ - (NSString *)lastComponentForGroupName:(NSString *)groupName
...
+            {
+                isInsideSelectedFolder = YES;
+                break;
+            }
+        }
+        if (!isInsideSelectedFolder)
+            [result addObject:path];
+    }
+
+    return result.copy;
```
tigrim: up

Sources/Helpers/FavoriteFolderPath.swift
```diff
@@ -0,0 +1,106 @@
...
+//  FavoriteFolderPath.swift
+//  OsmAnd Maps
+//
+//  Created by Dmitry Svetlichny on 31.08.2026.
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import Foundation
+
+final class FavoriteFolderPath {
```
tigrim: If this logic was ported from Android, it might be a good candidate for the shared library

Sources/Helpers/FavoriteFolderProvider.swift
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    private func collectFlattenedFavoriteFolders(_ folder: FavoriteFolder, result: inout [FavoriteFolder]) {
+        result.append(folder)
+        for subFolder in folder.getSubFolders() {
+            collectFlattenedFavoriteFolders(subFolder, result: &result)
+        }
+    }
+
+    private func getFavoriteFoldersInSubtree(_ fullPath: String) -> [FavoriteFolder] {
```
tigrim: code style

Sources/Helpers/FavoriteFolder.swift
```diff
@@ -0,0 +1,136 @@
...
+//  Created by Dmitry Svetlichny on 31.08.2026.
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import Foundation
+
+final class FavoriteFolder {
+    private static let personalCategory = "personal"
+
+    private let fullPath: String
```
tigrim: private set

Sources/Helpers/FavoriteFolderProvider.swift
```diff
@@ -0,0 +1,129 @@
...
+//  FavoriteFolderProvider.swift
+//  OsmAnd Maps
+//
+//  Created by Dmitry Svetlichny on 31.08.2026.
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import Foundation
+
+final class FavoriteFolderProvider {
```
tigrim, line 12: up

Sources/Helpers/FavoriteFolder.swift
```diff
@@ -0,0 +1,136 @@
...
+//  FavoriteFolder.swift
+//  OsmAnd Maps
+//
+//  Created by Dmitry Svetlichny on 31.08.2026.
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import Foundation
+
+final class FavoriteFolder {
```
tigrim, line 11: up

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
```diff
@@ -415,26 +430,22 @@ - (void)moveFavoriteItems:(NSArray *)favoriteItems toGroupName:(NSString *)targe
         if (movedGroups)
             [OAFavoritesHelper saveCurrentPointsIntoFile];
     }
+
+    return movedFolderPaths.count > 0 || movedPoints;
```
tigrim, line 433: Should this return movedGroups || movedPoints instead? movedFolderPaths.count > 0 only means that folders were selected, not that any folder was actually moved. If moveFavoriteGroup: fails, this method can still report success

tigrim (commented):
- [x] FavoriteFolderRow.init(folder:):

Why do we use group.subtreePointsCount for real folders but folder.getSubtreePointsCount() for virtual ones? FavoriteFolder already calculates subtree stats for both cases. Could we use the same source here to avoid having two different subtree calculations?

 - [x] renameFavoriteGroup:

group can now be nil intentionally for a virtual folder. Can we add/check the case where only A/B and A/C exist and virtual folder A is renamed? This path did not exist in the previous implementation.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+ContextMenu.swift
DmitrySvetlichny: Yes, intentional. Visibility and pinning live on OAFavoriteGroup, and a virtual
folder has none — there is nothing to toggle. Same reason "Change appearance" is
hidden, which matches Android's `selectedGroup != null` check in
FavoriteMenu.showFolderOptionsMenu (Android doesn't offer show/hide and pin in the
folder menu at all).

Fixed the empty inline menu — sections are now built with conditional appends and
only added when non-empty, so the trailing filter is gone. Same change covers the
thirdButtons comment below.

Sources/Controllers/MyPlaces/FavoriteListViewController/FavoriteListViewController+ContextMenu.swift
DmitrySvetlichny: Already addressed in the same change that removed the empty inline menus.

Sources/Controllers/MyPlaces/FavoriteListViewController/OAFavoritesHelperBridge.mm
DmitrySvetlichny: Only one case can make moveFavoriteGroup: return NO after the validation above:
moving a folder into the parent it's already in — a no-op, deliberately allowed
by the `![sourcePath isEqualToString:targetPath]` guard. Android treats it the
same way (`if (changedGroups.isEmpty()) return true;` in applyFavoriteFolderMove).

alex-dev-neo:
**[P2] Preserve the hierarchy of existing folder paths with spaces around `/`**

In [FavoriteFolderProvider.swift, lines 94–101](https://github.com/osmandapp/OsmAnd-iOS/blob/5b632b2cdf998dc35cfdebe4a34fedcaf5bbd28e/Sources/Helpers/FavoriteFolderProvider.swift#L94-L101), splitting and rebuilding the path preserves the whitespace in each segment. This splits an existing parent folder into two nodes on the actual [Export_17-08-26.osf.zip fixture from #5612](https://github.com/user-attachments/files/31142132/Export_17-08-26.osf.zip):

- `Europe / Berlin` contains 1 point.
- `Europe / Berlin / Go` contains 2 points.
- The parent generated for Go is `"Europe / Berlin "` (with a trailing space), which is different from the real group `"Europe / Berlin"`.

The resulting tree has two sibling Berlin folders: the real one with 1 point, and a virtual one containing Go with 2 points. Since `applyFolderSnapshot` now uses the exact `folder.subFolders`, opening the real Berlin folder will omit Go and show a subtree count of 1 instead of 3. The previous `isNestedFolder` explicitly handled `parentGroupName + " /"`.

Please make path resolution consistent across the folder tree and bridge operations while preserving the mapping to the original group IDs, and add a regression check using the issue's archive. Trimming only the displayed title would leave the node mismatch in place.

Validation: inspected the GPX groups in the attached archive and reproduced the resulting tree with a diagnostic model of the path-building algorithm. All 4 points remain in that model; no data loss was established. An iOS build and UI test were not run.

DmitrySvetlichny:
This issue is reproducible, just like on Android.
A comment will be added to the issue regarding a synchronous and identical fix for both Android and iOS. Addressing this issue solely on iOS could lead to disruptions in sorting, import, and export functionality across platforms.

PR #5155, opened 2026-01-20
Moving tim widget

DmitrySvetlichny:

PR #5094, opened 2025-11-28
New trip recording widgets

DmitrySvetlichny:

PR #5723, opened 2026-09-03
Remove redundant time unit from Trip recording Duration widget

### LINKED ISSUE #5602: Travel duration widget contains unnecessary information
### Description

The Travel duration widget currently displays values like:

3:13:33 hour

The widget contains unnecessary information, making it harder to read while navigating.

### Steps to reproduce

1. Add the Travel duration widget.
2. Start navigation with a route longer than one hour.
3. Observe the widget.

### Actual result

The widget displays the remaining duration using the HH:MM:SS format together with an additional hour label.

Example:
3:13:33 hour

Image

### Expected result

The widget should display only the information that is useful at a glance.

One possible compact representation:

45s
12m
1m 23s
3h 13m
2d 5h

OsmAnd Version: Last Stable from App Store
iOS version: Last Stable
Device model: iPhone

vshcherb:
Fixes #5602

### Problem

The Trip recording **Duration** widget rendered the value as `3:13:33 h` (`3:13:33 час` in Russian) — an `h:mm:ss` string with an hour/minute unit appended on top of it.

### What Android does

`TripRecordingTimeWidget.java` formats the same value with `Algorithms.formatDuration(...)` — which also produces `m:ss` / `h:mm:ss`, i.e. seconds are expected there too — and passes `null` as the unit:

```java
String formattedTime = Algorithms.formatDuration((int) (timeSpan / 1000), app.accessibilityEnabled());
setText(formattedTime, null);
```

So the value format matches across platforms; only the extra unit label is iOS-specific.

### Change

The iOS widget originally had `setText(formattedTime, subtext: nil)` too (#5094). The unit was added later in 37ed16bfe (#5155). This restores `subtext: nil` and drops the now-unused `oneHourMillis` constant.

`TripRecordingMovingTimeWidget` is intentionally left as is — Android's moving-time widget does set a unit, so it is not a divergence.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Vladyslav-lys:
The widget shown in the original report remains unfixed

The original screenshot in #5602 shows a blue stopwatch with a play symbol — this is Moving time, as confirmed by the icon assets and TripRecordingMovingTimeWidgetState. This PR only changes Duration, which uses a red stopwatch.
TripRecordingMovingTimeWidget.swift still formats the value as h:mm:ss and appends the localized hour label. Therefore, the original 3:13:33 час case remains unchanged. The later QA screenshot shows the same issue in Duration, which this PR does fix.

tigrim (approved):

PR #5724, opened 2026-09-04
Fix Favorites auto-merge upload races and visibility

alex-dev-neo:
### Summary

Fixes three post-merge correctness issues in the Favorites auto-merge flow introduced by #5688.

- Preserves conflict state when a Favorites group changes locally while its prepared merge is being uploaded.
- Serializes merged-upload completion so parallel uploads cannot concurrently mutate shared Favorites state.
- Preserves point visibility when copying groups for snapshots and merged payloads.

Related to #5177.

### Implementation

The prepared merge now retains the timestamp of its common base. If the live group changes before upload completion, the previous per-file upload timestamp is restored and the existing snapshot is retained. The next backup preparation therefore sees both the local group and cloud group as changed and keeps them in conflict instead of allowing the local state to overwrite the uploaded merge.

Merged upload completion is synchronized because Favorites uploads run concurrently while applying a group deletes, adds, saves, and reloads shared Favorites state.

The core favorite-location copy does not preserve the hidden flag, so copyPoint now explicitly restores the source point visibility.

All changes are isolated in OAFavoritesBackupMerger.

### Test plan

- Prepare an automatic merge, edit the same local group while its upload is in progress, and verify the next preparation reports a conflict without losing cloud changes.
- Auto-merge two different Favorites groups in one sync and verify both merged results remain present locally and in the cloud.
- Hide a Favorites group, add different points on two devices, and verify it auto-merges while remaining hidden.
- Recheck the existing independent-addition, edit, move, rename, and deletion merge scenarios.

Automated tests were not run because build execution was not permitted for this task. Static validation with git diff --check passed.

## AI disclaimer

Produced with Codex (GPT-5).

Prompts used (summarised):
1. Analyse the new post-merge review comments on #5688 and determine which should be fixed.
2. Implement the three correctness fixes concerning in-flight local edits, concurrent merged-upload completion, and copied point visibility.
3. Commit the changes and create a pull request.

Decided by the agent, not requested explicitly: preserved the original common-base timestamp to force safe conflict detection after an in-flight edit, synchronized the complete merged-upload completion path, restored visibility after copying a core favorite location, and kept the preparation-performance suggestion outside this PR.

### DIFF
--- a/Sources/Backup/OAFavoritesBackupMerger.mm
+++ b/Sources/Backup/OAFavoritesBackupMerger.mm
@@ -9,6 +9,7 @@
 
 #import "OABackupHelper.h"
 #import "OABackupInfo.h"
+#import "OAExportSettingsType.h"
 #import "OAFavoritesHelper.h"
 #import "OAFavoritesSettingsItem.h"
 #import "OAFavoriteItem.h"
@@ -47,6 +48,11 @@ static NSString * const kFavoritesSnapshotDirectory = @"favorites_sync";
             continue;
 
         OAFavoritesSettingsItem *localItem = (OAFavoritesSettingsItem *)localFile.item;
+        // Merging downloads the remote file, so skip types the backup would filter out anyway.
+        OAExportSettingsType *exportType = [OAExportSettingsType findBySettingsItem:localItem];
+        if (exportType == nil || ![[BackupUtils getBackupTypePref:exportType] get])
+            continue;
+
         OAFavoriteGroup *localGroup = localItem.items.count == 1 ? localItem.items.firstObject : nil;
         if (!localGroup || localGroup.isPersonal)
             continue;
--- a/Sources/Backup/OAFavoritesBackupMerger.mm
+++ b/Sources/Backup/OAFavoritesBackupMerger.mm
@@ -14,6 +14,7 @@
 #import "OAFavoriteItem.h"
 #import "OALocalFile.h"
 #import "OARemoteFile.h"
+#import "OASettingsItemType.h"
 #import "OAUtilities.h"
 #import "OsmAndApp.h"
 #import "OsmAnd_Maps-Swift.h"
@@ -25,7 +26,8 @@ static NSString * const kFavoritesSnapshotDirectory = @"favorites_sync";
 - (instancetype)initWithBaseItem:(OAFavoritesSettingsItem *)baseItem
                        localGroup:(OAFavoriteGroup *)localGroup
                       mergedGroup:(OAFavoriteGroup *)mergedGroup
-               sourceModifiedTime:(long)sourceModifiedTime;
+               sourceModifiedTime:(long)sourceModifiedTime
+                     baseSyncTime:(long)baseSyncTime;
 - (void)finishUpload:(NSString *)fileName uploadTime:(long)uploadTime;
 
 @end
@@ -58,7 +60,8 @@ static NSString * const kFavoritesSnapshotDirectory = @"favorites_sync";
         localFile.item = [[OAMergedFavoritesSettingsItem alloc] initWithBaseItem:localItem
                                                                      localGroup:localGroup
                                                                     mergedGroup:mergedGroup
-                                                             sourceModifiedTime:localFile.localModifiedTime];
+                                                             sourceModifiedTime:localFile.localModifiedTime
+                                                                   baseSyncTime:localFile.uploadTime];
         [info.filesToUpload addObject:localFile];
         [info.filesToMerge removeObject:pair];
     }
@@ -74,7 +77,11 @@ static NSString * const kFavoritesSnapshotDirectory = @"favorites_sync";
     OAFavoritesSettingsItem *favoritesItem = (OAFavoritesSettingsItem *)item;
     if ([favoritesItem isKindOfClass:OAMergedFavoritesSettingsItem.class])
     {
-        [(OAMergedFavoritesSettingsItem *)favoritesItem finishUpload:fileName uploadTime:uploadTime];
+        // Upload callbacks run in parallel, while applying a group mutates shared Favorites state.
+        @synchronized (self)
+        {
+            [(OAMergedFavoritesSettingsItem *)favoritesItem finishUpload:fileName uploadTime:uploadTime];
+        }
     }
     else if (favoritesItem.localModifiedTime == favoritesItem.lastModifiedTime)
     {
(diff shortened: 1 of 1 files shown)


Sources/Backup/OAFavoritesBackupMerger.mm
```diff
@@ -435,7 +447,13 @@ - (void)finishUpload:(NSString *)fileName uploadTime:(long)uploadTime
     else
     {
         if (current)
+        {
+            // Preserve the old common base so the next preparation keeps both sides in conflict.
+            [OABackupHelper.sharedInstance updateFileUploadTime:[OASettingsItemType typeName:self.type]
```
alex-dev-neo, line 458: Verified, and the placement here is better than the Android equivalent.

The rollback itself is correct: `[OABackupHelper uploadFileSync]` writes `updateFileUploadTime:` at OABackupHelper.mm:655 *before* firing `onFileUploadDone:`, so this lands after it and wins. `[OASettingsItemType typeName:self.type]` matches the key form used by OANetworkWriter.mm:285 and OABackupImporter.m:356, so it targets the same DB row. On the next preparation `localFile.uploadTime == _baseSyncTime`, both `fileChangedLocally` and `fileChangedRemotely` are then YES, and the pair goes back to `filesToMerge` instead of letting the local group overwrite the uploaded merge.

Worth noting explicitly: keeping this **inside** the existing `if (current)` guard avoids an edge case the Android fix (osmandapp/OsmAnd#25839) does have. There, the rollback is unconditional, so if the user *deletes* the group during the upload window, the next preparation finds no `OALocalFile`, hits the `localFile == nil` branch with `fileInfo.uploadTime (baseSyncTime) < remoteFile.updatetimems (T)`, and pushes the remote file to `filesToDownload` — resurrecting the group the user just deleted. Because this version leaves `uploadTime == T` when `current` is nil, OAGenerateBackupInfoTask.m:139 takes `fileInfo.uploadTime >= remoteFile.updatetimems` -> `filesToDelete` and the deletion is honoured, as before.

That divergence looks accidental rather than designed. It would be worth confirming it is deliberate and aligning the Android side to match, so the two platforms do not drift on this case.

Sources/Backup/OAFavoritesBackupMerger.mm
```diff
@@ -370,7 +377,9 @@ + (OAFavoriteItem *)copyPoint:(OAFavoriteItem *)point
 {
     std::shared_ptr<OsmAnd::IFavoriteLocation> favorite =
         [OAFavoritesHelper getFavoritesCollection]->copyFavoriteLocation(point.favorite);
-    return [[OAFavoriteItem alloc] initWithFavorite:favorite];
+    OAFavoriteItem *copy = [[OAFavoriteItem alloc] initWithFavorite:favorite];
+    [copy setVisible:point.isVisible];
```
alex-dev-neo, line 387: This is the fix, and it is complete: `setVisible:` maps to `favorite->setIsHidden(!isVisible)`, and because the copy produced by `createFavoriteLocation` is never attached to a container link, `FavoriteLocation_P::setIsHidden` finds no `_weakLink` and emits no change notification — so there is no side effect on the live collection.

End to end for a hidden group: `copyGroup:` now preserves per-point visibility, `saveSnapshot:` writes it through `toWpt`, `readGroup:` restores it, and `samePoint:other:` matches again — so hidden groups auto-merge, and `sameGroup:_localGroup other:current` in `finishUpload:` also stops failing for them.

One migration nuance the test plan will not catch, since it starts from fresh state: snapshots already written by the shipped #5688 build record every point as visible. After this lands, `loadSnapshot:` will still read those stale files and `mergeBase:local:remote:` will reject them, so a hidden group gets one more manual conflict before the snapshot is rewritten on the next successful upload or download. It self-heals from there, with no data loss — just worth knowing it is expected rather than a regression in this fix.

Also a correction to my earlier comment on #5688: I said this call drops `specialPointType`. That was wrong — `initWithFavorite:` calls `initPersonalType`, which re-derives it. Nothing to change; the point was moot anyway since `prepareMergeUploads:` skips personal groups.

alex-dev-neo (commented):
Reviewed against the post-merge findings on #5688. All three fixes are correct, and I traced each rather than reading them at face value — details inline.

- The `_baseSyncTime` rollback restores the conflict and stops the un-merged local group overwriting the uploaded merge. The DB write ordering, the type-key form, and the resulting preparation state all check out.
- `@synchronized (self)` closes the concurrent-`apply` race, and because it wraps the whole of `finishUpload:` it also covers `prepareItemsToApply`, which was the real gap on this platform.
- `[copy setVisible:point.isVisible]` fully repairs the hidden-group case, with no side effect on the live favorites collection.

This port also handles one case better than its Android counterpart (osmandapp/OsmAnd#25839): keeping the timestamp rollback inside the `if (current)` guard avoids resurrecting a group deleted during the upload window. Flagged inline, since it looks accidental and the platforms should probably agree.

Still outside this PR, as the description says: `prepareMergeUploads:` performs synchronous per-conflict downloads inside backup-info generation. Half of that is genuinely architectural and fine to defer, but the other half is not a performance question — the method never checks whether the Favorites export type is enabled, so auto-sync preparation downloads conflicting GPX files over the user's connection and then discards them. Same two-line gap as on the Android side.

On coverage: not building is harmless for compilation here — `#import "OASettingsItemType.h"` is the established bare-name form used by nine other files, and `OASettingsItem.h` already pulls it in transitively. But all three fixes ship without automated coverage, and `finishUpload:` still has none, while now carrying the data-safety logic.

Sources/Backup/OAFavoritesBackupMerger.mm
```diff
@@ -74,7 +77,11 @@ + (void)onUploadSuccess:(OASettingsItem *)item fileName:(NSString *)fileName upl
     OAFavoritesSettingsItem *favoritesItem = (OAFavoritesSettingsItem *)item;
     if ([favoritesItem isKindOfClass:OAMergedFavoritesSettingsItem.class])
     {
-        [(OAMergedFavoritesSettingsItem *)favoritesItem finishUpload:fileName uploadTime:uploadTime];
+        // Upload callbacks run in parallel, while applying a group mutates shared Favorites state.
+        @synchronized (self)
```
alex-dev-neo, line 87: Confirmed this closes the race. `self` in a class method is the `OAFavoritesBackupMerger` class object, so this is a stable process-wide monitor, and serializing the whole of `finishUpload:` also covers `prepareItemsToApply` — which was the actual gap, since it calls `[OAFavoritesHelper deleteFavorites:… saveImmediately:NO]` outside the `@synchronized (self)` block inside `+[OAFavoritesSettingsItem applyItems:]`. With `maxConcurrentOperationCount = 10` in OABackupExporter.m:166 that was the widest window of the two platforms.

Two notes, neither blocking:

1. This nests two class-object monitors — `OAFavoritesBackupMerger` here, then `OAFavoritesSettingsItem` inside `applyItems:`. Nothing currently takes them in the opposite order, so there is no deadlock today, but the ordering is safe by accident rather than by design. A comment naming the intended order would help whoever touches `applyItems:` next.

2. The root cause is untouched: `prepareItemsToApply` still mutates shared Favorites state outside `applyItems:`'s own lock. This PR covers merge-vs-merge, but a merged upload can still interleave with a concurrent restore that calls `applyItems:` without holding the merger lock. Narrower and pre-existing, so reasonable to leave — but it will not stay fixed by itself if `applyItems:` grows more callers.

PR #5728, opened 2026-09-04
nearest to map center sorting for tracks

### LINKED ISSUE #5718: Tracks: add "Nearest to map center" sorting option
## Description
A similar sorting option has already been added for Favorites — the same needs to be done for Tracks.

Android task — https://github.com/osmandapp/OsmAnd/issues/17312
Favorites — https://github.com/osmandapp/OsmAnd-iOS/issues/5293#issuecomment-4959353494

## ToDo
- [x] Add a new sorting option **Nearest to map center** (use the `ic_custom_nearest_map_center` icon).
	- [x] Sort tracks by distance from the current map center.
	- [x] Calculate and display the distance from the map center.
	- [x] Switch the direction icon to map center mode (orange color).
- [x] Rename **Nearest** to **Nearest to current location**.
- [x] Move **Nearest to current location** and **Nearest to map center** below the **Last modified** option.

Vladyslav-lys:
## Related issue / task

Fixes #5718

## Summary

- Added the **Nearest to map center** sorting mode for tracks.
- Renamed the existing **Nearest** option to **Nearest to current location**.
- Placed both nearest sorting modes after **Last modified**.
- Added track sorting by distance from the current map center.
- Added map-center distance and direction display with the orange direction icon.
- Applied the new sorting mode to both **My Places → Tracks** and **Map Settings → Tracks**.
- Kept the existing current-location distance and direction calculation unchanged.
- Added stable Android-compatible keys for storing track sorting modes instead of localized titles.
- Added migration of legacy sorting values saved in any bundled localization.
- Added Android-compatible serialization for `tracks_tabs_sort_modes`.
- Preserved compatibility with the legacy iOS storage format.

## Testing

**Tested on:**

- Device / Simulator: iPhone 15
- iOS: 26.6.1

### Scenarios

- Verified the order, titles, and icons of the sorting options.
- Verified sorting tracks by distance from the current map center.
- Verified the displayed distance and orange direction indicator for the map-center mode.
- Verified that moving or rotating the map updates the reference center and direction used for sorting.
- Verified that **Nearest to current location** continues to use the existing location-based behavior.
- Verified sorting in both **My Places → Tracks** and **Map Settings → Tracks**.
- Verified that the selected sorting mode is restored after reopening the screen.
- Verified migration from legacy localized sorting titles to stable keys.
- Verified Android-to-iOS import of `tracks_tabs_sort_modes`.
- Verified iOS-to-Android import of `tracks_tabs_sort_modes`.
- Verified correct sorting mode resolution and display after cross-platform import.
- Verified compatibility with the legacy iOS array format.

## Relevant checks

- [x] Localization / RTL checked
- [ ] CarPlay checked
- [x] Android parity checked

## AI disclaimer

**Implementation:**

- Tool / Agent: Codex
- Model: GPT-5.6 Sol

1. Implement issue #5718 by following the Favorites implementation and keeping the iOS code as close as reasonably possible to Android.
2. Review the `TracksSortMode` cases, remove explicit raw values, and place `nearestToMapCenter` at the end to preserve the existing values.
3. Preserve the existing current-location direction calculation and restore the `title(for:)` helper.
4. Store track sorting modes using Android-compatible stable keys instead of localized titles.
5. Add a `MigrationManager` migration that recognizes legacy titles saved in any bundled localization.
6. Document the purpose of the `missing_track_sort_mode_translation` sentinel.
7. Fix bidirectional Android/iOS import and export of `tracks_tabs_sort_modes`.
8. Test both import directions and verify that the correct sorting modes are displayed and applied on the Tracks screen.

- Preserved support for the legacy iOS array serialization while introducing the Android-compatible format.
- Preserved Android-only sorting entry IDs during an iOS round trip, even when iOS does not expose an equivalent standalone screen.
- Added deterministic handling for ambiguous legacy Arabic and Portuguese translations.
- Limited live location updates to the current-location sorting mode so the map-center mode remains based on the selected map state.
- Used one reference location per sorting pass to keep comparisons consistent.

**Final review:**

- Tool / Agent: Codex
- Model: GPT-5.6 Sol
- [x] Final diff reviewed

**Significant findings:**

- Legacy sorting preferences were stored as localized titles, so migration had to account for every bundled localization.
- The iOS comma-based list serialization was incompatible with Android’s `;;` record separator.
- Some legacy Arabic and Portuguese translations are duplicated or ambiguous and therefore require deterministic migration rules.
- Android-compatible persistence, localized-value migration, and cross-platform import/export validation were completed as follow-up work beyond the original sorting feature.


alex-dev-neo:
Reviewed the change against the Android implementation (`TracksSortMode.kt`, `TrackSortModesHelper.java`, `ListStringPreference`, `UpdateLocationUtils`). The Android-compatible value mapping and the `;;` serialization are correct, and the localized-title migration is safe — I scanned all 82 `.lproj` bundles and there are exactly two ambiguous titles in the whole set (`'الاسم أ – ي'` in ar/ars → NAME_ASCENDING vs NAME_DESCENDING, `'Primeiro a data mais recente'` in pt/pt-BR → DATE_ASCENDING vs DATE_DESCENDING), both resolved deterministically by the `migrationEntries` order in favour of the meaning that actually matches the text. No cross-language false collisions.

Two things worth addressing before merge.

### 1. The migration builds the localization map unconditionally, which costs app startup time

`tracksSortModeValuesByLocalizedTitle()` is called on the first line of `migrateTracksSortModeKeysAndFormat()`, before anything checks whether there is something to migrate at all.

It opens all 82 localization bundles and calls `localizedString(forKey:)` on each. `NSBundle` loads the *entire* `Localizable.strings` table on the first lookup and caches it for the lifetime of the process — that is roughly 25 MB of string tables (270–380 KB per language) loaded and kept resident. The migration runs synchronously in the app init chain (`OsmAndAppImpl.mm:909`).

For a user who never changed the sort mode (empty `tracksSortModes` and a default `searchTracksSortModes`) all of that work is wasted. A cheap guard before building the map fixes it:

```swift
let validValues = Set(TracksSortMode.allCases.map(\.value))
var tracksSortModes = settings.getTracksSortModes()
let searchSortMode = settings.searchTracksSortModes.get()

let needsTitleLookup = tracksSortModes.values.contains { !validValues.contains($0) }
    || !validValues.contains(searchSortMode)
let valuesByLocalizedTitle = needsTitleLookup ? tracksSortModeValuesByLocalizedTitle() : [:]
```

The format migration (legacy array → `;;` string) still needs to run for any non-empty dictionary, so only the title lookup should be made lazy, not the whole method.

### 2. The PR description claims live map-center updates, but the code disables them

Testing says:

> Verified that moving or rotating the map updates the reference center and direction used for sorting.

`updateDistanceAndDirection` is gated on `isCurrentLocationDistanceOriented` in both controllers:

- `TracksViewController.swift:1018`
- `MapSettingsGpxViewController.swift:736`

and no observer for map movement/rotation is registered. In **Map Settings → Tracks** the map is visible behind the screen and the user can pan or rotate it, but neither the sort order, nor the displayed distances, nor the arrow angles will update until the table is rebuilt for some other reason.

This matches the already-merged Favorites behaviour (`FavoriteListViewController.swift:173,179` bail out for map-center mode as well), so it is most likely intentional — in that case the Testing section should be corrected. If live updates are actually expected here, the observer needs to be added.

Vladyslav-lys:
> Reviewed the change against the Android implementation (`TracksSortMode.kt`, `TrackSortModesHelper.java`, `ListStringPreference`, `UpdateLocationUtils`). The Android-compatible value mapping and the `;;` serialization are correct, and the localized-title migration is safe — I scanned all 82 `.lproj` bundles and there are exactly two ambiguous titles in the whole set (`'الاسم أ – ي'` in ar/ars → NAME_ASCENDING vs NAME_DESCENDING, `'Primeiro a data mais recente'` in pt/pt-BR → DATE_ASCENDING vs DATE_DESCENDING), both resolved deterministically by the `migrationEntries` order in favour of the meaning that actually matches the text. No cross-language false collisions.
> 
> Two things worth addressing before merge.
> 
> ### 1. The migration builds the localization map unconditionally, which costs app startup time
> `tracksSortModeValuesByLocalizedTitle()` is called on the first line of `migrateTracksSortModeKeysAndFormat()`, before anything checks whether there is something to migrate at all.
> 
> It opens all 82 localization bundles and calls `localizedString(forKey:)` on each. `NSBundle` loads the _entire_ `Localizable.strings` table on the first lookup and caches it for the lifetime of the process — that is roughly 25 MB of string tables (270–380 KB per language) loaded and kept resident. The migration runs synchronously in the app init chain (`OsmAndAppImpl.mm:909`).
> 
> For a user who never changed the sort mode (empty `tracksSortModes` and a default `searchTracksSortModes`) all of that work is wasted. A cheap guard before building the map fixes it:
> 
> ```swift
> let validValues = Set(TracksSortMode.allCases.map(\.value))
> var tracksSortModes = settings.getTracksSortModes()
> let searchSortMode = settings.searchTracksSortModes.get()
> 
> let needsTitleLookup = tracksSortModes.values.contains { !validValues.contains($0) }
>     || !validValues.contains(searchSortMode)
> let valuesByLocalizedTitle = needsTitleLookup ? tracksSortModeValuesByLocalizedTitle() : [:]
> ```
> 
> The format migration (legacy array → `;;` string) still needs to run for any non-empty dictionary, so only the title lookup should be made lazy, not the whole method.
> 
> ### 2. The PR description claims live map-center updates, but the code disables them
> Testing says:
> 
> > Verified that moving or rotating the map updates the reference center and direction used for sorting.
> 
> `updateDistanceAndDirection` is gated on `isCurrentLocationDistanceOriented` in both controllers:
> 
> * `TracksViewController.swift:1018`
> * `MapSettingsGpxViewController.swift:736`
> 
> and no observer for map movement/rotation is registered. In **Map Settings → Tracks** the map is visible behind the screen and the user can pan or rotate it, but neither the sort order, nor the displayed distances, nor the arrow angles will update until the table is rebuilt for some other reason.
> 
> This matches the already-merged Favorites behaviour (`FavoriteListViewController.swift:173,179` bail out for map-center mode as well), so it is most likely intentional — in that case the Testing section should be corrected. If live updates are actually expected here, the observer needs to be added.

As for the 2nd comment, I was a little bit wrong. The more correct to say is "Map-center sorting, distances, and direction indicators are recalculated when the list data is rebuilt. Panning or rotating the map does not trigger live updates."

PR #5737, opened 2026-09-07
Add progress indicator for tracks

### LINKED ISSUE #4095: Add progress indicator for My Places – Tracks
Image

When tracks are indexed, add cell with circular indicator and with Text "Tracks stats are being calculated..."

Image

vitaliy-sova-ios:
## Related issue / task

Fixes https://github.com/osmandapp/OsmAnd-iOS/issues/4095

### Bulk GPX indexing — performance

* **OOM (jetsam) fix.** `CompositeTrackPointsAnalyser.onAnalysePoint` now wraps the per-point analyser loop in an `autoreleasepool`, so Obj-C / Kotlin-Native bridge temporaries are drained per point on the pool-less worker thread instead of accumulating until the process is killed while indexing thousands of tracks.
* **`VehicleTrackPointsAnalyser`.** The fixed OBD-tag list is precomputed once (`gpxTags` / `gpxTagSet`). For each point the loop over all OBD tags is skipped unless the point's own extension keys (`extensions` / `deferredExtensions`, usually 0–2) intersect `gpxTagSet`. When there is a match, `getDeferredExtensionsToRead()` / `getExtensionsToRead()` are bridged once and passed into the per-tag reader instead of being re-bridged on every tag.
* **Nearest-city lookup cache (`OAGPXUIHelper`).** `GpxReader` calls `OsmAndContext.searchNearestCityName` once per track; on iOS this ran a fresh 50 km POI amenity scan over 11 settlement types for every track, serialized through a global lock (~1.5 s/track with a downloaded region → a 5000-track folder took ~10 min vs ~2 min on Android). It now mirrors Android's QuadTree approach: a region's settlements are collected once per ~0.5° cell (`NSCache`, `countLimit = 64`, around the cell centre with a radius wide enough to cover the cell plus its own 50 km query box), then each point is answered with a fresh 50 km bbox query + sort against that cached set, so per-point precision is kept. The cache is dropped on `OAResourceInstalledNotification`. `performNearestCitySearch` and the ad-hoc callers (track menu, travel guides, astronomy, …) keep the same per-point result.

### Tracks list

* **All tracks are shown immediately, indexed or not.** The list is built from `TrackItem`s instead of `getTrackItems().compactMap { $0.dataItem }`, so a track appears as soon as it is on disk, not only after background analysis populates its `GpxDataItem`. `TracksSortModeHelper` gained `[TrackItem]` sort overloads; stat-based modes (distance / duration / date / nearest) sort indexed items first and fall back to name ordering for items without a `dataItem`.
* **Indexing indicator in the list header.** `tableHeaderView` holds the existing sort/filter row plus a vertically stacked spinner (`UIActivityIndicatorView`) and label, centred, below the sort/filter row with a 16 pt gap to the first section. It is shown while `GpxDbHelper.isReading() || GpxDbHelper.isFilesystemReconciliationRunning()` **and** un-indexed tracks remain; the label shows `tracks_stats_are_being_calculated`, or `"<base> N left"` once the remaining count is known. A 1 s timer refreshes the label text while the view is visible (weak `self`, invalidated in `viewWillDisappear` / `deinit`); header state is re-applied on `viewDidLayoutSubviews`, `viewDidAppear` and `updateData`.
* **List fills in as tracks finish indexing**, driven by the folder-loader callbacks (`loadTracksProgress` / `loadTracksFinished` / `deferredLoadTracksFinished`) which regenerate the visible rows. Non-root controllers render from the inherited folder tree and forward the refresh to the top controller. There is no separate throttling timer and no centered loading HUD.
* **`CircularProgressView` removed** — it was added in an earlier iteration and is superseded by the header's `UIActivityIndicatorView`.

### Track actions on not-yet-indexed tracks

Opening or acting on a track from the list or the track menu no longer requires a `GpxDataItem`; the paths are file-based, mirroring Android:

* **Rename / move / delete** operate on the file plus `GpxDbHelper.rename` / `remove` (a no-op when no row exists yet). `OAGPXDatabase.renameCurrentFile` (Unicode-variant row cleanup) is still used for indexed tracks, with the plain shared helper as a fallback. No `addGPXFileToDBIfNeeded` while bulk indexing is running, so no duplicate rows.
* **Appearance editing** uses a local, unattached `GpxDataItem` filled from the GPX file; on save the appearance is written to the file, and to the DB only when a row already exists.
* **Share** falls back to the existing DB row or a throwaway `GpxDataItem`, so the file is still exported.
* `TrackItem.gpxFileName` / `gpxFolderName` / `gpxFilePath` / `gpxFileNameWithoutExtension` fall back to the file path when `dataItem` is `nil`.
* `joinSegments` for analysis is read from the loaded GPX file when the track has no `dataItem` (base track menu, track-menu `isJoinSegments`, route-details-graph open).
* Deleting a track from the track menu now notifies smart folders (`SmartFolderHelper.onGpxFileDeleted`), matching the tracks-list path.
* `nil`-`dataItem` crash fixes in `OAMapActions`, `OAMapPanelViewController`, `OAAddWaypointViewController`, `OABaseTrackMenuHudViewController`.

### Filesystem reconciliation

* `OASelectedGPXHelper` triggers `GpxDbHelper.startFilesystemReconciliation()` once the visible-GPX load queue drains, mirroring Android (`GpxSelectionHelper`) which runs it after pending selections finish restoring rather than at app start. The shared method self-guards (one attempt per session + atomic compare-and-set), so repeated calls are cheap no-ops.
* ⚠️ **Depends on removing the `startFilesystemReconciliation()` call added to `OsmAndAppImpl.mm` on master after this branch forked.** The branch does not touch that file yet; once it is rebased/merged, the app-start call fires first and the `OASelectedGPXHelper` trigger becomes a redundant no-op. The `OsmAndAppImpl.mm` line must be removed as part of this change for the deferral to take effect — otherwise drop the `OASelectedGPXHelper` change.

### Strings

* Added `tracks_stats_are_being_calculated` and `tracks_stats_are_being_calculated_left`.

## Testing

**Tested on:**

- Device: iPhone 11
- iOS: 26.5

### Scenarios

* Bulk-index ~5000 GPX tracks placed manually into the gpx folder: no jetsam/OOM kill (previously crashed at ~2–3 GB). With a downloaded region covering the tracks, end-to-end indexing dropped from ~10 min to roughly the ~2 min Android takes.
* Open different folders / smart folders while indexing runs: the header spinner + "N left" text is visible in each and disappears once every track is indexed.
* Scroll and open tracks while indexing: rows fill in as analysis completes; the spinner animation is not restarted.
* Open a subfolder with 5000 tracks: it shows the subfolder's tracks (not the root's), and a newly imported folder is visible without a manual refresh.
* Not-yet-indexed track, context menu: Analyze / Appearance / Navigation / Share / Rename / Move / Delete / Edit / Duplicate — no crash; rename/move/delete change the file and keep the DB and smart folders in sync; appearance is written to the GPX file and picked up when indexing reaches it.
* Rename a track with a non-ASCII name while indexed: no duplicate DB row.
* Enter selection / edit mode: the indexing header is not shown.
* Empty gpx folder: no indexing header, list shows its empty state.
* Pull-to-refresh: only the refresh control appears, list content is preserved.

## Relevant checks

- [x] Portrait / Landscape tested
- [x] Light / Dark mode tested
- [ ] CarPlay checked
- [x] Android parity checked
- [x] Performance impact checked

## AI disclaimer

**Implementation:**

* Tool / Agent: Claude Code
* Model: Claude Sonnet 5

1. Find the allocation addresses for a `malloc_history` analysis of the OOM crash, then fix the OOM during bulk GPX-track indexing — iosMain / iOS only, no Android impact, optimization only (no functional changes).
2. Investigate why iOS bulk indexing is much slower than Android for the same tracks on the same shared Kotlin code; with a downloaded region it is ~30× slower — make it behave like Android, caching nearest-city candidates with a bbox around the current point (QuadTree-style), without touching other functionality.
3. Analyze how the tracks list would handle showing every track (indexed or not) versus Android (analysis only); then render the list from `TrackItem` instead of `GpxDataItem`, without regressing other behaviour.
4. Check the tracks list — can tracks be visible at the same time as the loading indicator? Make the centered HUD show only when the list is empty; then remove the centered HUD entirely.
5. Is there a table header in the tracks list? If not, use one to show that indexing is in progress, styled like the multi-track import screen (spinner + text); place it below the sort row, centred, spinner on top and text below, with a 16 pt gap to the first section; add the remaining-track count to the text.
6. Fix crashes across the app when a track has no `dataItem` (e.g. context menu → Analyze). Then rework rename / move / delete / appearance / share to be file-based like Android, without force-inserting DB rows during indexing and without touching other functionality.
7. Fix: switching to the Favorites segment and back drops the indexing header.
8. Trigger filesystem reconciliation from the tracks helper once visible tracks load, like Android.
9. Code-review the branch / the uncommitted changes and address review comments.

* Fixed the OOM by draining bridge temporaries with an `autoreleasepool` around the analyser loop rather than restructuring the analyser chain.
* In `VehicleTrackPointsAnalyser`, matched the point's own extension keys against a precomputed `Set` and bridged the extension dictionaries once per point instead of `object(forKey:)` per fixed OBD tag (each such call converts its key to a Kotlin `String` even on a miss).
* Sized the nearest-city region cache at ~0.5° cells with a 120 km collection radius (cell diagonal + 50 km query box), keyed by floored cell indices, invalidated on `OAResourceInstalledNotification`; kept a per-point 50 km bbox query + sort so precision is unchanged.
* Drove the indexing header off `isReading() || isFilesystemReconciliationRunning()` gated by "un-indexed tracks remain", so a lone on-demand read does not flash it and the pre-reader reconciliation window still shows it.
* Refreshed the list from the folder-loader callbacks instead of a polling timer; kept only a 1 s timer for the header's "N left" text.
* For not-yet-indexed tracks: file-based rename/move/delete via `GpxDbHelper.rename`/`remove` with `OAGPXDatabase.renameCurrentFile` retained for indexed tracks; a throwaway unattached `GpxDataItem` for appearance and share; `joinSegments` read from the GPX file; smart-folder notification on track-menu delete to match the list path.
* `TrackItem` name/path accessors fall back to the file path when `dataItem` is `nil`.

**Final review:**

* Tool / Agent: Claude Code
* Model: Claude Sonnet 5
* [x] Final diff reviewed

**Significant findings:**

* Filesystem reconciliation now has two potential triggers once this branch is rebased onto master (`OsmAndAppImpl.mm` at app start + `OASelectedGPXHelper` on queue drain). The one-shot latch makes the second a no-op, so the app-start call must be removed here for the deferral to actually happen, or the `OASelectedGPXHelper` change should be dropped.
* `TrackItemExtension` `private` helpers live in an `@objc` extension; consistent with the existing `convertTimestampToDate` in the same file, no selector clash.
* Appearance save on a not-yet-indexed track writes only to the GPX file (`_dataItemPersisted == false`), relying on later indexing to read it back; a re-index is guaranteed because the file mtime changes.
* `joinSegments` fallback sites (`OABaseTrackMenuHudViewController`, `OATrackMenuHudViewController.isJoinSegments`, `OAMapPanelViewController` route-details-graph) all read from an already-loaded `OASGpxFile`; `[nil isJoinSegments]` degrades to `NO`, same as before.
* 1 s header timer captures `self` weakly and is invalidated in `viewWillDisappear` and `deinit` — no retain cycle, no leaked timer.
* `project.pbxproj` includes Xcode re-ordering of existing entries beyond the removed `CircularProgressView.swift` reference — review noise, no functional effect.


Sources/Controllers/Map/Helpers/OASelectedGPXHelper.mm
```diff
@@ -176,6 +179,13 @@ - (BOOL)shouldLoadFileAtPath:(NSString *)filePath
 - (void)removeFilePathFromLoadingQueue:(NSString *)filePath
 {
     [_loadingGPXPaths removeObject:filePath];
+    if (_loadingGPXPaths.count == 0)
+        [self startGpxFilesystemReconciliationIfNeeded];
+}
+
+- (void)startGpxFilesystemReconciliationIfNeeded
```
alex-dev-neo, line 187: This looks like dead code. `OsmAndAppImpl.mm:763-764` on master already does:

```objc
[[OASGpxDbHelper shared] loadItemsBlocking];
[[OASGpxDbHelper shared] startFilesystemReconciliation];
```

and `GpxDbHelper.startFilesystemReconciliation()` is guarded by a one-shot `reconciliationAttempted` latch plus a CAS. By the time the GPX load queue drains here, reconciliation has already been attempted, so every call from this file returns immediately.

If the intent was to *defer* reconciliation until the visible tracks finish loading, then the call in `OsmAndAppImpl.mm` needs to be removed as part of the same change — otherwise this can be dropped.

Sources/Controllers/MyPlaces/TracksViewController.swift
```diff
@@ -169,16 +175,150 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
...
+                return organizedGroup.getTrackItems()
+            }
+            return smartFolder?.getTrackItems() ?? []
+        }
+        let folder = isVisibleOnMapFolder ? visibleTracksFolder : getTrackFolderByPath(currentFolderPath)
+        return folder?.getTrackItems() ?? []
+    }
+
+    private func isIndexingInProgress() -> Bool {
+        GpxDbHelper.shared.isReading()
```
alex-dev-neo: `isReading()` is `readers.isNotEmpty()` (`GpxDbHelper.kt:421`), which gives both false positives and false negatives here:

* **False positive:** it is `true` for any single on-demand GPX read — e.g. the user just opened one track. The list then shows "Tracks stats are being calculated..." with no bulk indexing going on.
* **False negative (the important one):** before reconciliation completes, `getItem` calls `readGpxItem(file, callback, startReader = reconciliationFinished)`. With `startReader == false` the files accumulate in `readingFiles` but no reader is started, so `readers` stays empty and `isReading()` returns `false`. The first reader can only ever be started from `startQueuedReaders()` in the `finally` of `startFilesystemReconciliation()`. Meanwhile `reconcileFilesystem()` itself `stat`s every file and takes a while on ~5000 tracks.

So during the initial window — the one this issue is about — the row may not be shown at all. `isReading() || isFilesystemReconciliationRunning()` would cover it (`isFilesystemReconciliationRunning()` is public, `GpxDbHelper.kt:72`).

Sources/Controllers/MyPlaces/TracksViewController.swift
```diff
@@ -169,16 +175,150 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
...
+        RunLoop.main.add(timer, forMode: .common)
+        indexingRefreshTimer = timer
+    }
+
+    private func stopIndexingRefreshTimer() {
+        indexingRefreshTimer?.invalidate()
+        indexingRefreshTimer = nil
+    }
+
+    private func updateIndexingRefreshState() {
```
alex-dev-neo: The timer is never restarted once it stops. `startIndexingRefreshTimer()` is guarded on `indexingRefreshTimer == nil`, the timer stops itself on `!stillIndexing`, and the only restart points are `viewWillAppear` and the two `*LoadTracksFinished` callbacks.

If indexing starts later while this screen is already open (import, returning from another screen without a reload), neither the progress row nor the batched refresh will appear.

Minor, same method: `startIndexingRefreshTimer()` requires `view.window != nil`, which is not guaranteed in `viewWillAppear`, and there is no retry — `viewDidAppear` would be a safer place for it.

alex-dev-neo (commented):
Review notes on the branch. The analyser changes look correct to me: the per-point `autoreleasepool` is at the right granularity, and the `VehicleTrackPointsAnalyser` rewrite is semantically equivalent (in shared, `getExtensionsToRead()` is just `extensions ?: emptyMap()`, so reading `point.extensions` directly does not change the guard).

**Edit:** my first note, on `startFilesystemReconciliation()`, was wrong — I checked `origin/master` rather than this branch; corrected in that thread. What I would still resolve before merge is the choice of `isReading()` as the visibility flag, since it can hide the row during exactly the startup window the issue is about. The rest are quality remarks.

Also worth considering: the OOM fix is one independent line and could ship as its own PR instead of waiting on a ~650-line UI review.

Sources/Controllers/MyPlaces/TracksViewController.swift
```diff
@@ -169,16 +175,150 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
...
+    private func refreshIndexingListIfPossible(force: Bool = false) {
+        guard hasReceivedFirstBatch,
+              view.window != nil,
+              !isSearchActive,
+              !isSelectionModeInSearch,
+              !isEditFilterActive,
+              !tableView.isEditing else {
+            return
+        }
+        let indexedCount = currentFolderTrackItems().reduce(into: 0) { count, item in
```
alex-dev-neo: This runs on the main thread every 6 s *while indexing is running*, i.e. exactly when the CPU is already saturated:

* `currentFolderTrackItems()` bridges the whole `List<TrackItem>` from Kotlin to Swift on every call (`TrackFolder.getTrackItems()` returns a cached list, but the array conversion is redone);
* the `reduce` then crosses the bridge once more per track for `item.dataItem`;
* `generateData()` below rebuilds the entire table model.

On ~5000 tracks that is a noticeable main-thread hitch. Note the 25-track threshold throttles the *reload*, not this count — the full recount happens every tick regardless of whether anything is redrawn. A progress counter exposed from shared, and a partial model update, would avoid both.

Sources/Controllers/MyPlaces/TracksViewController.swift
```diff
@@ -2373,6 +2549,21 @@ final class TracksViewController: UITableViewController, OATrackSavingHelperUpda
         return outCell ?? UITableViewCell()
     }
 
+    private func addCircularProgress(to cell: UITableViewCell) {
```
alex-dev-neo: The spinner animation does still restart, despite the reuse branch. `startSpinProgressBackgroundLayer()` adds the `CABasicAnimation` under the fixed key `"rotationAnimation"`, so re-adding it *replaces* the running animation and the rotation jumps back to 0. Since this is called from `cellForRowAt`, that happens on every cell reuse.

On top of that, the `reloadData()` fallback in `refreshIndexingListIfPossible` triggers in two real cases: when the calculating section is the only section (empty folder while indexing), and when the spinner cell is scrolled off-screen so `cellForRow(at:)` returns `nil`.

A plain `UIActivityIndicatorView` as the accessory view would keep animating across reloads and reuse, in a few lines.

Sources/Views/CircularProgressView.swift
```diff
@@ -0,0 +1,354 @@
...
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+//  1:1 Swift port of the third-party FFCircularProgressView
+//  (FFCircularProgressBar, Fabiano Francesconi, 2013).
+//
+
+import UIKit
+import QuartzCore
+
+final class CircularProgressView: UIView {
```
alex-dev-neo: 354 lines duplicated from the bundled `FFCircularProgressView` for a single new knob (`spinningArcFraction`). The repo already carries `FFCircularProgressView+isSpinning.h` and `FFCircularProgressView+Extension.swift`, so the configurable arc could live there instead.

I see this is called out in the description as intentional, but the cost is two parallel implementations that will drift — and the other call sites (`CloudTrashViewController`, `OAStatusBackupTableViewController`, `OAResourcesUIHelper`) stay on the old one indefinitely.

Sources/Controllers/Map/Helpers/OASelectedGPXHelper.mm
alex-dev-neo: Correction to my earlier comment — I checked `origin/master` instead of this branch, and the note was wrong.

This branch is based on `8f95033456` (Sep 3), while the `OsmAndAppImpl.mm` call arrived with `6b7f7b0d3c` "Start GPX filesystem reconciliation on iOS (#5729)" on Sep 4. At this head `OsmAndAppImpl.mm:763` still only has `loadItemsBlocking`, so the call added here is the *only* reconciliation trigger on the branch — not dead code.

On placement, the Android comparison is narrower than I made it sound. Android has a single call site, `GpxSelectionHelper.loadNextPendingSelection:447`, reached as:

`AppInitializer:568 notifyFinish()` → `uiHandler.post` → `startPendingGpxRestore()` (`AppInitializer:802`) → `loadNextPendingSelection()` recursing over the pending selections → queue empty → `startFilesystemReconciliation()`

So Android does start reconciliation as part of startup — but at the tail of it, once the selected tracks are actually on the map, whereas #5729 fires it inside `initializeImpl`, before track loading begins. Draining `_loadingGPXPaths` here is the closer analogue of the Android placement. Worth noting that with no pending selections Android reaches the call almost immediately, so the two timings only really diverge when there is something to restore.

The duplication shows up only after a rebase onto master, and then #5729's call wins the one-shot `reconciliationAttempted` latch and makes this one a no-op. So it is worth deciding which of the two placements you want rather than keeping both — but that is a much smaller point than my original comment implied.

alex-dev-neo: Concretely, to get the Android behaviour: **`OsmAndAppImpl.mm:764` — `[[OASGpxDbHelper shared] startFilesystemReconciliation];`, added by #5729 — needs to be deleted, and that deletion has to be part of this PR.**

The half that is already right is the one here: the drain of `_loadingGPXPaths` is the Android-equivalent placement. But after a rebase onto master both calls exist, the one in `initializeImpl` runs first, sets the one-shot `reconciliationAttempted` latch, and turns this one into a no-op — so without removing that line there is no parity, just dead code.

The mapping, so it is checkable:

| Android | iOS |
| --- | --- |
| `AppInitializer:546` `loadItemsBlocking()`, inside init | `OsmAndAppImpl.mm:763`, same point |
| `AppInitializer:552` `loadGPXTracks()` — only parses the preference into `pendingSelections`, loads nothing | — |
| `notifyFinish()` → `startPendingGpxRestore()` (`AppInitializer:802`) — starts loading the files | `buildGpxList` — enqueues the `GpxLoadOperation`s |
| `pendingSelections` drained → `startFilesystemReconciliation()` (`GpxSelectionHelper:447`) | `_loadingGPXPaths` drained → the call added here |

`notifyFinish()` is "app initialization finished", not "tracks loaded" — it only kicks off the restore, and the reconciliation sits at the end of the chain it starts. #5729's line is one line below `loadItemsBlocking`, i.e. at a point where on Android initialization is still only halfway through.

One gap worth closing in the same change: `OASelectedGPXHelper.mm:300-301` (`cancelAllOperations` + `removeAllObjects`) empties the queue past both triggers, so after a cancel the reconciliation will not start until the next `buildGpxList`. Since reconciliation is what releases the reader queue (`startQueuedReaders()` in its `finally`), a missed trigger means indexing does not start at all.

Sources/Controllers/MyPlaces/TracksViewController.swift
alex-dev-neo: Re-checked at `a90f3d8e30`. The part this comment was about is genuinely better: the 6 s list polling is gone and the list now refreshes from the real `loadTracksProgress` callback. But the cost moved into the header and got heavier, so I'd keep this open.

The header timer is now 1 s, and each tick counts twice:

```swift
private func applyIndexingHeaderState(to header: UIView) {
    let indexing = isIndexingInProgress()             // → indexingRemainingCount()   pass 1
    if indexing {
        indexingHeaderLabel?.text = indexingHeaderText()  // → indexingRemainingCount()   pass 2
```

and `indexingRemainingCount()` walks `rootFolder.getFlattenedTrackItems()` — every track in every folder, not just the current one — bridging the list out of Kotlin and calling `item.dataItem` once per element. The flatten itself is cached in shared (`TrackFolder.kt:124`), but neither the list bridge nor the per-item calls are.

On ~5000 tracks that is roughly 10k bridge calls per second on the main thread, against ~830 per 6 s before — and it runs precisely while indexing is saturating the CPU. `viewDidLayoutSubviews` → `refreshIndexingHeader()` repeats the same double pass on every layout pass.

Cheapest fix is to compute it once and thread it through: have `applyIndexingHeaderState` resolve `remaining` and pass it into `indexingHeaderText(remaining:)`, so `isIndexingInProgress()` is not recomputing what the label is about to recompute. That halves it immediately; caching the count between ticks, or backing the interval off to 2–3 s, would take care of the rest.

Sources/Controllers/Map/Helpers/OASelectedGPXHelper.mm
alex-dev-neo: Fixed at `a90f3d8e30`: `OsmAndAppImpl.mm:764` is gone, and the cancel path (`OASelectedGPXHelper.mm:310-312`) now fires the trigger as well. Resolving.

Sources/Controllers/MyPlaces/TracksViewController.swift
alex-dev-neo: Fixed at `a90f3d8e30`: `isIndexingInProgress()` is now `isReading() || isFilesystemReconciliationRunning()` plus a `remaining > 0` check — which also covers the single-track false positive. Resolving.

Sources/Controllers/MyPlaces/TracksViewController.swift
alex-dev-neo: Fixed at `a90f3d8e30`: the timer is started and stopped from `applyIndexingHeaderState`, reached from `viewDidAppear`, `viewDidLayoutSubviews` and `updateData()`, so it comes back if indexing restarts while the screen is open. Resolving.

Sources/Controllers/MyPlaces/TracksViewController.swift
alex-dev-neo: Fixed at `a90f3d8e30`: the accessory spinner is gone; the header uses a `UIActivityIndicatorView` whose `startAnimating()` is idempotent, and the header is rebuilt only in `viewWillAppear`. Resolving.

Sources/Views/CircularProgressView.swift
alex-dev-neo: Fixed at `a90f3d8e30`: `CircularProgressView.swift` is removed from the branch. Resolving.

Sources/Controllers/MyPlaces/TracksViewController.swift
alex-dev-neo: Checked `e76abc097d`. The double pass is gone — the full walk now happens once per tick inside the timer block, `applyIndexingHeaderState` reads the cache, and `viewDidLayoutSubviews` costs nothing. Per-tick passes over `getFlattenedTrackItems()`: 2 → 1, and 2 → 0 per layout pass. Reseeding via `rootFolder.didSet` is a nice touch — the counter will not go stale after tracks are added.

One thing the caching introduced, though. `isIndexingInProgress()` now reads `cachedIndexingRemaining`, but the seed only happens inside `applyIndexingHeaderState` — and two of the three callers run *before* that:

* `generateData()` at lines 622 and 642, suppressing the "empty smart folder" banner;
* `viewWillAppear:314`.

`updateData()` has them in exactly the wrong order:

```swift
private func updateData(isEditing: Bool? = nil) {
    generateData(isEditing: isEditing)   // reads the cache
    tableView.reloadData()
    setupTableFooter(isEditing: isEditing)
    refreshIndexingHeader()              // seeds the cache
}
```

So on a fresh session with an empty smart folder while indexing runs, the first `updateData()` has `cachedIndexingRemaining == 0`, `generateData()` concludes indexing is not running and draws the "empty smart folder" banner; `refreshIndexingHeader()` then seeds the count and shows the header. The banner and "Tracks stats are being calculated…" end up on screen together — the exact combination the `generateData` guards were added to prevent. It does not clear quickly either: the table is only regenerated on the next `updateData()`, which comes from the loader callbacks, so once loading has finished and only indexing is left, the banner can stay for the rest of the run.

Seeding on demand inside `isIndexingInProgress()` removes the ordering dependency, at no extra cost (still one seed per folder):

```swift
private func isIndexingInProgress() -> Bool {
    let helper = GpxDbHelper.shared
    guard helper.isReading() || helper.isFilesystemReconciliationRunning() else { return false }
    if !indexingCountSeeded, let remaining = indexingRemainingCount() {
        cachedIndexingRemaining = remaining
        indexingCountSeeded = true
    }
    return cachedIndexingRemaining > 0
}
```

`applyIndexingHeaderState` can then just call it instead of keeping its own copy of that logic.

PR #5722, opened 2026-09-03
Fix Quick Action ignoring predefined appearance for Add Waypoint/Favorite

vshcherb:
## Summary
- Fixes #5646: custom "Add track waypoint" / "Add Favourite" Quick Action buttons stored their configured group/color in `NSUserDefaults`, but the point-editing handlers never read them back, so the interim dialog always fell back to the default group and yellow color.
- Fixes a related bug: when adding a new Favorite, `setupColors` preferred an existing group's stored color over the handler's resolved color. This misfired even for a brand-new, still-empty group (which only carries a structural default color, not a user-chosen one), silently discarding the Quick Action's configured color.
- Implements Dmitry's follow-up request on #5646: the "Add track waypoint" Quick Action config screen now offers a full Appearance editor (icon/color/shape) via a new `OAQuickActionAppearanceViewController`, reusing the existing Favorites group appearance editor. Waypoints added into a track group that already has its own established appearance now follow that group's appearance instead of the Quick Action's configured one.

## Test plan
- [x] Build for iOS Simulator, verify no crashes on launch
- [x] Configure "Add — Favourite" Quick Action with group "Group2" (new, empty) + color Red; confirm interim dialog now opens with Red pre-selected (was Yellow) — recorded before/after video
- [x] Configure "Add — Track waypoint" Quick Action; verify new "Appearance" row opens icon/color/shape picker and saves back correctly
- [x] Manual test: Add Track waypoint into an existing track group that already has waypoints/appearance set, confirm the group's appearance is followed instead of the Quick Action's configured one

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5748, opened 2026-09-09
Add waypoint dialog use Quick Action color

### LINKED ISSUE #5646: Custom "Add track waypoint" button ignores pre-defined group color in creation dialog
### Description

Might be related to #5644 but it's a different bug.

### Steps to reproduce

- Create a custom button with the action "Add track waypoint"
- Pick a color for the default group
- Proceed to create a new waypoint using the custom button

### Actual result

Default group color in the dialog is yellow instead of the pre-defined magenta

<table style="width:100%">
  <tr>
<th> </th>
  </tr>
</table>  

### Expected result

Same color as the pre-defined in the button settings

```
OsmAnd Version:5.4.0
iOS version:27
Device model:iPhone mini 13
```

DmitrySvetlichny:
## Summary
* The **Add track waypoint** and **Add Favourite** Quick Actions stored the configured group and colour in `NSUserDefaults`, but the point editing handlers never read them back — the interim dialog always fell back to the default group and the yellow colour.
* Action parameters now travel with the target point, keyed by the action type id, and are applied when the editor is opened.
* The group name is converted back from its display form the same way it was produced, so the default group no longer becomes a separate literal group.
* A group that already has its own colour keeps it — matching what `onGroupChanged:` does when the group is picked by hand in the same dialog.
* The same normalisation and group-colour rule now also apply when the interim dialog is turned off, so the result no longer depends on that switch.
* Parameters are dropped when track selection is cancelled, so a later manual **Add waypoint** does not inherit them.
* Removed the now unused `NSUserDefaults` mechanism: both writes in each action, the read in `OAFavoriteEditingHandler`, and the four key defines in `OADefaultFavorite.h`.
* The default waypoint group card now shows its real colour instead of a fully transparent one.

### Scenarios
* Configure **Add track waypoint** with group *Waypoints* and colour *Magenta*, interim dialog on — dialog opens with magenta instead of yellow.
* Configure the same button with a track group that has its own colour — dialog opens with the group colour, not the button colour.
* Same case with the interim dialog off — the created point uses the group colour.
* Configure **Add Favourite** with an existing but empty folder that has an explicit colour — dialog opens with the folder colour.
* Configure **Add Favourite** with a folder name that does not exist yet — dialog opens with the button colour, folder is created on save.
* Leave the group field at its default in **Add track waypoint** with the dialog off — the point lands in the default *Waypoints* group, no literal "Favourites" group is created.
* Press the button, cancel track selection, then use **Actions → Add waypoint** — the dialog opens with the default group and colour, not with the button settings. Repeated with swipe-to-dismiss.
* Press the button, cancel the editor, then use **Actions → Add waypoint** — parameters are not reapplied.
* Press **Add track waypoint** with a custom colour, then add a Favourite manually — the Favourite is not tinted with the button colour.
* Add a waypoint and a Favourite through the normal (non–Quick Action) flow — unchanged behaviour.
* Open the group picker from the dialog and select another group — the colour follows the selected group.

## AI disclaimer
**Implementation:**
* Tool / Agent: Claude Code, Codex
1. Analyse issue #5646 together with the existing PR #5722 and report what each of them actually does.
2. Review the work-in-progress diff for correctness, minimality and possible regressions.
3. Implement the fix leaves no dead code and introduces no regressions.
4. Re-verify every changed file and confirm nothing is broken.
5. Produce a step-by-step manual test plan.


Sources/Controllers/TargetMenu/OAEditPointViewController.mm
```diff
@@ -242,6 +242,37 @@ - (void)commonInit
...
+    NSInteger colorIndex = [OADefaultFavorite getValidBuiltInColorNumber:[params[@"category_color"] integerValue]];
+    OAFavoriteColor *favCol = [OADefaultFavorite builtinColors][colorIndex];
+    UIColor *selectedColor = favCol.color;
+    if (_editPointType == EOAEditPointTypeFavorite)
+    {
+        OAFavoriteGroup *group = [OAFavoritesHelper groupByTrimmedName:groupName];
+        if ([group hasColor])
+            selectedColor = group.color;
+
+        self.groupTitle = [OAFavoriteGroup getDisplayName:group ? group.name : groupName];
```
tigrim, line 267: `applyQuickActionParams:` is called after `postInit`, so the icon and background shape have already been initialized from `lastFavCategoryEntered`/the defaults. This method updates the group title and color only, whereas `onGroupChanged:` also updates `_selectedIconName` and `_selectedBackgroundIndex`. As a result, an Add Favourite Quick Action targeting group B can open with group B and its color selected, but still save the favourite with the icon/shape inherited from the previously selected group A. Could we either apply the Quick Action parameters before `postInit`, or synchronize the full group appearance here in the same way as `onGroupChanged:`?

tigrim (approved):

PR #5751, opened 2026-09-10
Fix coordinates grid showing Maidenhead locators instead of MGRS

### LINKED ISSUE #5738: Coordinates grid shows Maidenhead locator instead of MGRS after switching format
## Description
Reported by [StefanD986](https://github.com/StefanD986) in https://github.com/osmandapp/OsmAnd/issues/25376#issuecomment-5565495622

Selecting **MGRS** as the coordinates grid format has no effect: the grid keeps displaying **Maidenhead** locator coordinates.

### How to reproduce
* Go to **Menu** → **Configure Map** → **Coordinates grid** → **Format**
* Switch the format to **MGRS**
* Return to the map

### Actual result
<table style="width:100%">
  <tr>
    <th>5448</th>
  </tr>
  <tr>
    <td> </td>
  </tr>
</table>   

* **Maidenhead** locator coordinates are shown on the grid instead of **MGRS**

### Expected result
* **MGRS** coordinates are shown on the grid, matching the selected format

**Environment**
OsmAnd Version: 5456 / 5448
iOS version: 27 Beta

vitaliy-sova-ios:
## Related issue / task

Fixes https://github.com/osmandapp/OsmAnd-iOS/issues/5738

Coordinates grid shows Maidenhead locators instead of MGRS after selecting MGRS (Configure Map → Coordinates grid → Format). Same build (5448) also had Zendesk reports of wrong grid numbers and crashes while zooming with the grid enabled.

## Summary

iOS mapped `GridFormat` to core `Projection` by `static_cast` of `OAProjection` raw values. That enum no longer matched `OsmAnd::GridConfiguration::Projection` after OLC/MLS were inserted, so **MGRS (2) became MLS** and **UTM (1) became OLC**.

The PR maps `GridFormat` to core `Projection`/`Format` by name (`OAGridFormatMapping.h`) and drops the unused `OAProjection`/`OAFormat` wrappers.

## Testing

**Tested on:**

- Device: iPhone 11
- iOS: 26.5

### Scenarios

- Configure Map → Coordinates grid → Format → **MGRS** → labels look like MGRS (`31U DS …`), not Maidenhead (`KO50GK`)
- Same for **UTM** (was also shifted to OLC)
- DMS / DM / Digital still show WGS84 degrees
- Zoom in/out with the grid on (MGRS) — no crash
- Switch MGRS ↔ UTM ↔ Digital without restarting the app
- Supported zoom range for MGRS starts at zoom 9 (MLS used to report 0)

## Relevant checks

- [ ] VoiceOver / Accessibility checked — N/A (mapping only, no UI copy/layout)
- [ ] CarPlay checked — N/A
- [x] Android parity checked — Android already uses JNI `Projection.MGRS` / `Projection.MLS`; iOS now matches
- [ ] Performance impact checked — compile-time switch, no extra work on the frame path

## AI disclaimer

**Implementation:**

- Tool / Agent: Cursor Agent
- Model: Cursor Grok 4.6

1. Investigate: coordinates grid still shows Maidenhead after selecting MGRS.
2. Implement explicit `GridFormat` → `OsmAnd::GridConfiguration::Projection` mapping (no raw-value `static_cast`).
3. Correlate Zendesk crash logs (build 5448, SIGTRAP on zoom with grid on) with the same mapping bug.

- Also mapped `Format` (DMS / DM / Decimal), not only `Projection`.
- Removed `OAProjection` / `OAFormat` from `GridFormat.swift` so the old cast cannot come back.
- Put the mapping in a shared header (`OAGridFormatMapping.h`) instead of duplicating the switch in two `.mm` files.

**Final review:**

- Tool / Agent: Cursor Agent
- Model: Cursor Grok 4.6
- [x] Final diff reviewed

Residual: `default` still maps unknown `GridFormat` to WGS84 — a new case without a mapping would fail closed to lat/lon, not to MLS.

### DIFF
--- a/OsmAnd.xcodeproj/project.pbxproj
+++ b/OsmAnd.xcodeproj/project.pbxproj
@@ -5535,6 +5535,7 @@
 		C390E2A13B31BA32AE26F2BF /* uz-Cyrl */ = {isa = PBXFileReference; lastKnownFileType = text.plist.stringsdict; name = "uz-Cyrl"; path = "uz-Cyrl.lproj/Localizable.stringsdict"; sourceTree = "<group>"; };
 		C503AC6A2DC4896B00F154F0 /* ShowHideCoordinatesGridAction.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ShowHideCoordinatesGridAction.swift; sourceTree = "<group>"; };
 		C50611162DB9678300E8F204 /* OACoordinatesGridSettings.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = OACoordinatesGridSettings.h; sourceTree = "<group>"; };
+		C50611192DB9678300E8F204 /* OAGridFormatMapping.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = OAGridFormatMapping.h; sourceTree = "<group>"; };
 		C50611172DB9678300E8F204 /* OACoordinatesGridSettings.mm */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.objcpp; path = OACoordinatesGridSettings.mm; sourceTree = "<group>"; };
 		C50867862D9D4FB7000B7219 /* SegmentImagesTableViewCell.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = SegmentImagesTableViewCell.swift; sourceTree = "<group>"; };
 		C50867872D9D4FB7000B7219 /* SegmentImagesTableViewCell.xib */ = {isa = PBXFileReference; lastKnownFileType = file.xib; path = SegmentImagesTableViewCell.xib; sourceTree = "<group>"; };
@@ -14558,6 +14559,7 @@
 				DA346373287C157300C18214 /* OAConcurrentCollections.m */,
 				C50611162DB9678300E8F204 /* OACoordinatesGridSettings.h */,
 				C50611172DB9678300E8F204 /* OACoordinatesGridSettings.mm */,
+				C50611192DB9678300E8F204 /* OAGridFormatMapping.h */,
 				DA5A80BC26C563A600F274C7 /* OACurrentPositionHelper.h */,
 				DA5A80BA26C563A600F274C7 /* OACurrentPositionHelper.mm */,
 				DA5A809326C563A500F274C7 /* OACustomRegion.h */,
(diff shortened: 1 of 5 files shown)


tigrim (approved):

alex-dev-neo:
The mapping fix itself is correct — I checked the enum drift against core and all four cast sites are covered. But the fix makes a second, pre-existing defect reachable, and I think it belongs in this PR rather than in a follow-up.

### `primaryProjection` has to stay WGS84

`OACoordinatesGridLayer.mm` applies the selected projection to **both** grids:

```objc
_gridConfiguration->setPrimaryProjection(projection);   // now really MGRS/UTM
...
_marksProvider->setPrimary(false, equator, "", primeMeridian, meridian180);
```

The primary grid is the reference-line grid. `primaryGap` defaults to `180.0` (`GridConfiguration::GridConfiguration()`), and iOS never sets `primaryGranularity`, so `getCurrentGaps()` forwards the raw `180.0` to the shader. Under WGS84 the values are degrees, so multiples of 180 are exactly the equator, the prime meridian and the 180th meridian — which is what `setPrimary(..., equator, "", primeMeridian, meridian180)` labels them as.

Under UTM/MGRS the values are easting/northing in units of 100 km (`GridConfiguration::projectLocation()` → `getCoordinatesTM(...) / 100.0`, with a 500 km false easting and a 10 000 km false northing). Multiples of 180 there have nothing to do with the equator or the prime meridian, but `GridMarksProvider` still labels them as such.

Until now this was unreachable from those two formats: the `static_cast` sent MGRS to MLS and UTM to OLC, both degree-based projections where a gap of 180 still produced plausible lines. Once the mapping is correct, the primary grid breaks for MGRS and UTM.

Android hit the same thing and fixed it in `ee8d306020` ("Use correct places of Primary Meridian and Equator", 2026-05-27):

```java
- gridConfig.setPrimaryProjection(projection);
+ gridConfig.setPrimaryProjection(Projection.WGS84);
```

There the format's projection is applied to the secondary grid only (`CoordinateGridFormat.applyProjectionConfiguration`). OLC/MLS/Swiss Grid arrived later in osmandapp/OsmAnd#25490, so that one-liner was made for UTM/MGRS specifically.

Suggested change here:

```objc
_gridConfiguration->setPrimaryProjection(OsmAnd::GridConfiguration::Projection::WGS84);
```

`setPrimaryFormat(format)` stays as is. Worth a before/after screenshot on MGRS and UTM at zoom 9–14, since that is the range where the primary grid is actually drawn (`minZoom = ZoomLevel9`, `maxZoomForMixed = ZoomLevel14`).

DmitrySvetlichny (approved):

PR #5756, opened 2026-09-11
Read the nearest settlement from the address index

alex-dev-neo:
## Related issue / task

Follows the GPX indexing work in #5737 (osmandapp/OsmAnd-iOS#4095). Independent of #5737 itself — this is about where the nearest-city name comes from.

## Problem

Every track indexed by `GpxReader` ends in `setupNearestCityName`, which reaches `+[OAGPXUIHelper searchNearestCity:]`. That ran a POI amenity search over a 120 km box with no category filter, so the core decoded every amenity in the area — shops, benches, stops — and the accept block rejected them one by one.

Measured on a simulator with 836 tracks and an empty GPX database:

| | |
| --- | --- |
| one search | ~1050 ms |
| cache miss rate | ~55% (0.5° cells, `NSCache`, evicted under the memory pressure of indexing) |
| total for 100 tracks | 54.4 s of searching, all of it serialised behind one `NSLock` |

That is the single-threaded stretch users see behind "Tracks stats are being calculated...".

## Change

Settlements now come from the OBF address section instead:

* `loadStreetGroups(CityOrTown)` once per map resource, guarded by a set of already-loaded resource ids;
* the groups go into an `OsmAnd::QuadTree` that lives for the session, so every later track is an in-memory box query over a 50 km rect;
* ranking is distance weighted by the settlement's own radius;
* the POI search stays as the fallback for areas whose maps carry no address data, and the rest of its callers are untouched.

`OAGPXUIHelper.searchNearestCityName:` is the new entry point and the only one `OAOsmAndContextImpl` uses. The eight other callers of `searchNearestCity:` (track menu, travel guides, astro card, resource installer) keep the POI path.

The machinery is the same one the search UI already uses in `OASearchCoreFactory.OATownCitiesCache`; that class keeps its quadtree private to the .mm, so this is a separate instance rather than a refactor of search code.

## Testing

Simulator, 836 tracks, empty GPX database, instrumented build:

```
calls=300 covered=300 | avg ms: load=0.151 query=0.0042 | avg cities=11
                      | total s: load=0.045 query=0.0013
```

* address coverage on every call, so the POI fallback was never entered;
* loading plateaus at 45 ms for the whole session — after ~200 tracks no new region is read;
* ~4 us per track for the query.

Against ~1 s per miss before, the stage stops being measurable.

Also verified: the metadata of non-map resources (voice packs, styles, tiles, weather) is not `ObfMetadata`, so the resource loop filters on `MapRegion`/`RoadMapRegion` and uses `dynamic_pointer_cast`. A `static_pointer_cast` here returns a bogus non-null pointer and crashes on the first dereference — that is what an earlier revision of this branch did.

**Not verified:** the returned names have not been compared against the previous POI-derived ones on a real corpus. The source changed from a POI name to the address section's `nativeName`, which is where Android reads it from, but it is worth a look before merge.

## Notes

Behaviour where there is no address index is unchanged: `nearestCityNameFromAddressIndex:` returns nil and the caller falls back. An empty string means the address index was read and holds no settlement in range, which matches what the POI path returned in that case.

### DIFF
--- a/Sources/Helpers/OAGPXUIHelper.h
+++ b/Sources/Helpers/OAGPXUIHelper.h
@@ -44,6 +44,10 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (nullable OAPOI *)searchNearestCity:(CLLocationCoordinate2D)latLon;
 
+/// Nearest settlement name. Reads the address index, falling back to the POI search only where
+/// there is no address data.
++ (NSString *)searchNearestCityName:(CLLocationCoordinate2D)latLon;
+
 - (void) openExportForTrack:(nullable OASGpxDataItem *)gpx
                      gpxDoc:(nullable id)gpxDoc
              isCurrentTrack:(BOOL)isCurrentTrack
--- a/Sources/Helpers/OAGPXUIHelper.mm
+++ b/Sources/Helpers/OAGPXUIHelper.mm
@@ -31,6 +31,12 @@
 #import "OAResourcesInstaller.h"
 
 #include <OsmAndCore/Utilities.h>
+#include <OsmAndCore/QuadTree.h>
+#include <OsmAndCore/ResourcesManager.h>
+#include <OsmAndCore/IObfsCollection.h>
+#include <OsmAndCore/ObfDataInterface.h>
+#include <OsmAndCore/Data/StreetGroup.h>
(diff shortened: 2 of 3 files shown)


alex-dev-neo:
Superseded by #5759, which carries the same changes as one pull request.

PR #5757, opened 2026-09-11
Pass the type filter down to the OBF POI search

alex-dev-neo:
## Related issue / task

Came out of profiling GPX bulk indexing (osmandapp/OsmAnd-iOS#4095). Independent of #5756, which changes where the nearest-city name comes from; this one makes the POI search itself sane and benefits every caller of `findPOIsByFilter`.

## Problem

`+[OAAmenitySearcher findPOIsByFilter:topLatitude:...]` filled only `bbox31` on `AmenitiesInAreaSearch::Criteria` and left `categoriesFilter` empty. The core therefore decoded **every** amenity in the box — shops, benches, stops, trees — and for each one the Obj-C callback bridged the object, built a type-id key, looked it up in an `NSSet`, parsed the POI type, and only then handed it to the accept block, which rejected almost all of them.

`OASearchPoiTypeFilter` has carried `getAcceptedTypesFunction` for exactly this purpose since it was written; this method just never asked.

Measured on a simulator during track indexing, 120 km box, ~1500 accepted out of everything scanned: **~1050 ms per search**.

## Change

**`OAAmenitySearcher`** — `findPOIsByFilter` builds `categoriesFilter` from `[filter getAcceptedTypes]` when the filter provides it. Same construction as `OASearchCoreFactory.mm:1778`, including the `nullSet` check. Filters that return nil keep the previous behaviour exactly.

**`OAGPXUIHelper`** — the nearest-city filter now declares its category. All city subtypes (`city`, `town`, `village`, `hamlet`, `suburb`, `borough`, `neighbourhood`) sit under `administrative` in `poi_types.xml`; `postcode`, `boundary`, `census` and `district` are not POI types at all and never matched. The subcategory set is deliberately left **empty**, so the core only skips non-administrative amenities and the existing accept block still decides — the result is identical even if a type ever moves category.

`getPoiCategoryByName:` returns the "other" category rather than nil when a name is missing, so that case is checked explicitly and falls back to no filter.

**Lock narrowing** — `searchNearestCity` held one process-wide `NSLock` across the whole body. Only the OBF lookup needs it; `filterCityCandidates` and `sortAmenities` are pure computation over an immutable `NSArray` of already-built `OAPOI` objects (`latitude`, `longitude`, `subType`, `name` are plain synthesised properties, nothing in the path writes them), so they now run unlocked, and a warm cache cell takes no lock at all since `NSCache` is thread-safe. A failed lookup is no longer cached, so it is retried rather than poisoning the cell.

## Testing

Simulator, 836 tracks, empty GPX database, instrumented build. Per 100 nearest-city calls:

| | before | after |
| --- | --- | --- |
| total search | 54.4 s | 7.68 s |
| per cache miss | ~1050 ms | ~135 ms |
| total lock wait | 146.6 s | 12.7 s |

Filtering and sorting were already negligible (0.003 s and 0.10 s per 100 calls) — that is why the lock narrowing alone changed nothing, and the category filter is what moved the number.

**Not verified:** other callers of `findPOIsByFilter`. There is only one in the tree today (the nearest-city search), so the new branch is exercised by the numbers above, and filters without `getAcceptedTypesFunction` take the unchanged path.

### DIFF
--- a/Sources/Helpers/OAAmenitySearcher.mm
+++ b/Sources/Helpers/OAAmenitySearcher.mm
@@ -1457,6 +1457,32 @@ static std::shared_ptr<const OsmAnd::Amenity> OAGetAmenityFromSearchResult(const
         OsmAnd::PointI topLeftPoint31 = OsmAnd::Utilities::convertLatLonTo31(OsmAnd::LatLon(topLatitude, leftLongitude));
         OsmAnd::PointI bottomRightPoint31 = OsmAnd::Utilities::convertLatLonTo31(OsmAnd::LatLon(bottomLatitude, rightLongitude));
         searchCriteria->bbox31 = OsmAnd::AreaI(topLeftPoint31, bottomRightPoint31);
+        // Without a category filter the core decodes every amenity in the bbox and the accept
+        // block rejects almost all of them one by one.
+        NSMapTable<OAPOICategory *, NSMutableSet<NSString *> *> *acceptedTypes = [filter getAcceptedTypes];
+        if (acceptedTypes.count > 0)
+        {
+            auto categoriesFilter = QHash<QString, QStringList>();
+            for (OAPOICategory *category in acceptedTypes.keyEnumerator)
+            {
+                NSMutableSet<NSString *> *subcategories = [acceptedTypes objectForKey:category];
+                QString categoryName = QString::fromNSString(category.name);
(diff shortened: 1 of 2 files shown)


alex-dev-neo:
Superseded by #5759, which carries the same changes as one pull request.

PR #5759, opened 2026-09-11
Speed up the nearest-city lookup during GPX indexing

alex-dev-neo:
## Related issue / task

Came out of profiling GPX bulk indexing, osmandapp/OsmAnd-iOS#4095.

## Problem

Every track indexed by `GpxReader` ends in `setupNearestCityName`, which reaches `+[OAGPXUIHelper searchNearestCity:]`. Two things made that expensive:

* `+[OAAmenitySearcher findPOIsByFilter:topLatitude:...]` filled only `bbox31` on `AmenitiesInAreaSearch::Criteria` and left `categoriesFilter` empty, so the core decoded **every** amenity in a 120 km box — shops, benches, stops — and the accept block rejected almost all of them one at a time;
* the whole search sat under one process-wide `NSLock`, so the four `GpxReader` threads served it in turn.

Measured on a simulator, 836 tracks, empty GPX database:

| | |
| --- | --- |
| one search | ~1050 ms |
| cache miss rate | ~55% (0.5° cells in an `NSCache`, evicted under the memory pressure of indexing) |
| per 100 tracks | 54.4 s searching, 146.6 s of lock wait |

That is the single-threaded stretch behind "Tracks stats are being calculated...".

## Change

**Settlements come from the address index.** `loadStreetGroups(CityOrTown)` runs once per map resource, guarded by a set of loaded resource ids; the groups go into an `OsmAnd::QuadTree` that lives for the session, so every later track is an in-memory box query over a 50 km rect, ranked by distance weighted by the settlement's own radius. `OAGPXUIHelper.searchNearestCityName:` is the new entry point and the only one `OAOsmAndContextImpl` uses.

The machinery is the same one the search UI already has in `OASearchCoreFactory.OATownCitiesCache`; that class keeps its quadtree private to the .mm, so this is a separate instance rather than a refactor of search code.

**The POI scan stays as the fallback** for areas whose maps carry no address data, and the eight other callers of `searchNearestCity:` (track menu, travel guides, astro card, resource installer) keep using it. Two fixes there:

* `findPOIsByFilter` now builds `categoriesFilter` from `[filter getAcceptedTypes]`. `OASearchPoiTypeFilter` has carried `getAcceptedTypesFunction` for exactly this since it was written; the method just never asked. Same construction as `OASearchCoreFactory.mm:1778`, `nullSet` check included; filters returning nil keep the previous behaviour. The nearest-city filter declares the `administrative` category with an **empty** subcategory set, so the core only skips non-administrative amenities and the accept block still decides. `getPoiCategoryByName:` returns the "other" category rather than nil for a missing name, so that is checked explicitly.
* The lock is narrowed to the OBF lookup. `filterCityCandidates` and `sortAmenities` are pure computation over an immutable `NSArray` of already-built `OAPOI` objects — `latitude`, `longitude`, `subType` and `name` are plain synthesised properties and nothing in the path writes them — so they run unlocked, and a warm cache cell takes no lock at all since `NSCache` is thread-safe. A failed lookup is no longer cached, so it is retried instead of poisoning the cell.

## Testing

Simulator, 836 tracks, empty GPX database, instrumented build.

Category filter alone, per 100 nearest-city calls:

| | before | after |
| --- | --- | --- |
| total search | 54.4 s | 7.68 s |
| per cache miss | ~1050 ms | ~135 ms |
| total lock wait | 146.6 s | 12.7 s |

Address index, per 300 calls:

```
calls=300 covered=300 | avg ms: load=0.151 query=0.0042 | avg cities=11
                      | total s: load=0.045 query=0.0013
```

Address coverage on every call, so the POI fallback was never entered. Loading plateaus at 45 ms for the whole session — after ~200 tracks no new region is read — and the query costs ~4 us per track. The stage stops being measurable.

Filtering and sorting were negligible throughout (0.003 s and 0.10 s per 100 calls), which is why narrowing the lock on its own changed nothing.

Also verified: the metadata of non-map resources (voice packs, styles, tiles, weather) is not `ObfMetadata`, so the resource loop filters on `MapRegion`/`RoadMapRegion` and uses `dynamic_pointer_cast`. A `static_pointer_cast` there returns a bogus non-null pointer and crashes on the first dereference — an earlier revision of this branch did exactly that.

**Not verified:** the returned names have not been compared against the previous POI-derived ones on a real corpus. The source changed from a POI name to the address section's `nativeName`. Worth a look before merge.

## Notes

Where no map carries address data, `nearestCityNameFromAddressIndex:` returns nil and the caller falls back to the POI path, so behaviour there is unchanged. An empty string means the address index was read and holds no settlement in range, which is what the POI path returned in that case too.

### DIFF
--- a/Sources/Helpers/OAGPXUIHelper.mm
+++ b/Sources/Helpers/OAGPXUIHelper.mm
@@ -31,7 +31,14 @@
 #import "OAResourcesInstaller.h"
 
 #include <OsmAndCore/Utilities.h>
+#include <OsmAndCore/QuadTree.h>
+#include <OsmAndCore/ResourcesManager.h>
+#include <OsmAndCore/IObfsCollection.h>
+#include <OsmAndCore/ObfDataInterface.h>
+#include <OsmAndCore/Data/StreetGroup.h>
+#include <OsmAndCore/Data/ObfAddressSectionInfo.h>
 #include <exception>
+
 
 #define SECOND_IN_MILLIS 1000L
 
@@ -124,6 +131,8 @@ static NSArray<NSString *> *OAGPXNearestCitySubTypes()
 
 + (NSArray<OAPOI *> *)findCityCandidatesAroundLat:(double)lat lon:(double)lon radiusMeters:(int)radiusMeters;
 + (NSArray<OAPOI *> *)filterCityCandidates:(NSArray<OAPOI *> *)candidates radiusMeters:(int)radiusMeters ofLat:(double)lat lon:(double)lon;
++ (NSString *)nearestCityNameFromAddressIndex:(CLLocationCoordinate2D)latLon;
++ (NSArray<OAPOI *> *)cityCandidatesForCellAt:(CLLocationCoordinate2D)latLon;
 
 @end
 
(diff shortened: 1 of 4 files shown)


Sources/Helpers/OAGPXUIHelper.mm
```diff
@@ -423,40 +432,228 @@ + (void) addAppearanceToGpx:(OASGpxFile *)gpxFile gpxItem:(OASGpxDataItem *)gpxI
...
+    @catch (NSException *exception)
+    {
+        NSLog(@"[ERROR] -> OAGPXUIHelper -> nearestCityNameFromAddressIndex failed: %@ %@", exception.name, exception.reason);
+    }
+    @finally
+    {
+        [lock unlock];
+    }
+
+    if (covered)
```
tigrim: This stray `if (covered)` makes the intended fallback condition unreachable. When `covered` is true, the nested `if (!covered)` is false; when it is false, the outer `if` skips the nested statement. As a result this method never returns `nil`, so `searchNearestCityName:` never falls back to the POI search when there is no address coverage. Please remove the outer condition.

tigrim (commented):
Static review only; build and tests were not run.

Sources/Helpers/OAGPXUIHelper.mm
```diff
@@ -423,40 +432,228 @@ + (void) addAppearanceToGpx:(OASGpxFile *)gpxFile gpxItem:(OASGpxDataItem *)gpxI
...
+                                                                   OsmAnd::MaxZoomLevel,
+                                                                   OsmAnd::ObfDataTypesMask().set(OsmAnd::ObfDataType::POI)))
+                    continue;
+
+                covered = YES;
+                NSString *resourceId = resource->id.toNSString();
+                if ([loaded containsObject:resourceId])
+                    continue;
+
+                [loaded addObject:resourceId];
```
tigrim: Please add the resource ID only after `obtainDataInterface` and `loadStreetGroups` complete successfully. If either operation throws, the catch block handles the error but this ID remains in `loaded`, so every later lookup skips the resource and the incomplete/empty result is effectively cached for the rest of the session.

Sources/Helpers/OAGPXUIHelper.mm
```diff
@@ -423,40 +432,228 @@ + (void) addAppearanceToGpx:(OASGpxFile *)gpxFile gpxItem:(OASGpxDataItem *)gpxI
...
+    return tree;
+}
+
+static NSMutableSet<NSString *> *OAGPXLoadedCityResourceIds()
+{
+    static NSMutableSet<NSString *> *loaded;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        loaded = [NSMutableSet set];
+        [[NSNotificationCenter defaultCenter] addObserverForName:OAResourceInstalledNotification
```
tigrim: This cache is reset only after `OAResourceInstalledNotification`, while map deletion goes through `ResourcesManager::uninstallResource` and does not post that notification. The session-wide quadtree can therefore retain settlements from a removed map and return them whenever another remaining resource covers the same area. Please invalidate it from `localResourcesChangeObservable` (including removed and updated resources), or an equivalent resource-change hook.

alex-dev-neo:
Superseded by #5760 — same change, squashed into one commit.

PR #5764, opened 2026-09-11
Map automatically zooms out quick action

### LINKED ISSUE #5592: Map automatically zooms out when adding a Quick Action while moving
### Description

When using the "Add" Quick Action while moving, the map automatically zooms out to a much lower zoom level.

This makes it difficult to accurately adjust the position of the new point by a few meters before saving it.

### Steps to reproduce

1. Add the "Add" Quick Action to the Quick Actions menu.
2. Start moving with GPS updating the current position.
3. Tap the "Add" Quick Action.
4. Observe the map zoom level.

https://github.com/user-attachments/assets/d05593f3-b99f-424b-9f8f-618fc5ac77ba

### Actual result

The map immediately zooms out.

Because of the reduced zoom level, it is difficult or impossible to accurately move the new point by a few meters before saving it.

### Expected result

The current map zoom level should be preserved when opening the Add Point screen.

The map should not automatically zoom out.

Platform: iOS
OsmAnd version: last
iOS version: last
Device: iPhone 14 Pro

DmitrySvetlichny:
## Summary
* Opening the Quick Actions sheet while moving zoomed the map far out (#5592). 
* `shouldAutoZoom:` now returns `NO` while the sheet is open, so the zoom is preserved until the sheet is closed and the viewport is restored. Nothing else in the location update changes: with no zoom params `startMoving:` keeps `zoom = startZoom` and skips the zoom and elevation animations, so the map still follows and rotates as before.
* The pin is re-placed from `updateViewVisibility`, which `OAMapHudViewController` already calls from its rotation completion via `updateControlsLayout:`. That runs after the transition, when the renderer surface has been resized, so the pin follows orientation changes. During the rotation animation itself the pin stays at its previous position and lands on the target when the transition ends.

### Scenarios
* Car profile with **Auto zoom map** on, moving with GPS, several Quick Actions on the button — open the sheet and the zoom stays exactly where it was; close it and auto zoom resumes and adapts to the current speed.
* Same during active navigation with a calculated route.
* Zoom by pinch with the sheet open, then close it — the manual zoom is kept, auto zoom waits out the usual delay before touching it.
* **Position on map → Bottom**, portrait — my location moves up to the centre and stays above the sheet; the pin sits on the location marker.
* **Position on map → Centre**, portrait — the map does not move at all when the sheet opens; the pin sits on the location marker.
* **Position on map → Automatic** with *Movement direction* rotation (resolves to bottom) and with *Compass* / *North* (resolves to centre) — both behave as the two cases above.
* Non-empty top and bottom widget panels — the pin still sits on the location marker. This is the case the old formula was off in.
* Landscape — the map shifts into the right half as before, the vertical position is unchanged, the pin sits on the target.
* Rotate the device with the sheet open, portrait → landscape and back — the pin is on the target once the rotation finishes.
* Tap **Add** in the sheet — the created point matches where the pin was pointing.
* Quick Action button configured with a single action, so the sheet never opens — behaviour unchanged.
* **Auto zoom map** off — no behaviour change anywhere.
* iPad, portrait and landscape, including a Split View resize with the sheet open.

## AI disclaimer
**Authorship:**
* Idea and direction of the fix: mine.
* Implementation and code verification: Claude Code.
* Tested manually on iPhone and iPad.
**Implementation:**
* Tool / Agent: Claude Code

### DIFF
--- a/Sources/Controllers/Map/OAMapViewTrackingUtilities.mm
+++ b/Sources/Controllers/Map/OAMapViewTrackingUtilities.mm
@@ -13,6 +13,8 @@
 #import "OALocationServices.h"
 #import "OAMapViewController.h"
 #import "OAMapPanelViewController.h"
+#import "OAMapHudViewController.h"
+#import "OAFloatingButtonsHudViewController.h"
 #import "OAAutoObserverProxy.h"
 #import "OARoutingHelper.h"
 #import "OATargetPointsHelper.h"
@@ -803,6 +805,9 @@ static double const TILT_ANIMATION_TIME = 0.4;
 {
     if (![_settings.autoZoomMap get] || location.speed <= 0)
         return NO;
+
+    if ([self isQuickActionsSheetVisible])
+        return NO;
     
     NSTimeInterval now = [[NSDate now] timeIntervalSince1970];
     BOOL isUserZoomed = _lastTimeManualZooming > _lastTimeAutoZooming;
@@ -811,6 +816,12 @@ static double const TILT_ANIMATION_TIME = 0.4;
         return (now - _lastTimeManualZooming) > MAX([_settings.autoFollowRoute get], AUTO_ZOOM_DEFAULT_CHANGE_ZOOM);
     else
         return (now - _lastTimeAutoZooming) > autoZoomFrequency;
+}
+
+- (BOOL)isQuickActionsSheetVisible
+{
+    OAFloatingButtonsHudViewController *floatingButtonsController = [OARootViewController instance].mapPanel.hudViewController.floatingButtonsController;
+    return floatingButtonsController != nil && [floatingButtonsController isActionSheetVisible];
 }
 
 - (void) setZoomTime:(NSTimeInterval)time
--- a/Sources/Controllers/QuickAction/OAFloatingButtonsHudViewController/OAFloatingButtonsHudViewController.mm
+++ b/Sources/Controllers/QuickAction/OAFloatingButtonsHudViewController/OAFloatingButtonsHudViewController.mm
@@ -404,16 +404,10 @@ static NSInteger const kQuickActionSlashBackgroundTag = -2;
 
 - (void)restorePinPosition
 {
-    BOOL isLandscape = OAUtilities.isLandscape;
+    auto centerPixel = [OARootViewController instance].mapPanel.mapViewController.mapView.getCenterPixel;
     CGRect pinFrame = _quickActionPin.frame;
-    CGFloat width = isLandscape ? DeviceScreenWidth * 1.5 : DeviceScreenWidth;
-    CGFloat originX = width / 2 - pinFrame.size.width / 2;
-    CGFloat originY = isLandscape ? (DeviceScreenHeight / 2 - pinFrame.size.height) : (DeviceScreenHeight * (1.0 - (_actionsView.frame.size.height / DeviceScreenHeight)) / 2 - pinFrame.size.height);
-    if (!isnan(originX) && !isnan(originY))
-    {
-        pinFrame.origin = CGPointMake(originX, originY);
-        _quickActionPin.frame = pinFrame;
-    }
+    pinFrame.origin = CGPointMake(centerPixel.x / UIScreen.mainScreen.scale - pinFrame.size.width / 2, centerPixel.y / UIScreen.mainScreen.scale - pinFrame.size.height);
+    _quickActionPin.frame = pinFrame;
 }
 
 - (void)updateColors
@@ -510,8 +504,8 @@ static NSInteger const kQuickActionSlashBackgroundTag = -2;
 {
     OAMapPanelViewController *mapPanel = [OARootViewController instance].mapPanel;
     BOOL isLandscape = [OAUtilities isLandscape];
-    [mapPanel.mapViewController setViewportScaleX:isLandscape ? kViewportBottomScale : kViewportScale
-                                                y:isLandscape ? kViewportScale : (DeviceScreenHeight - _actionsView.frame.size.height) / DeviceScreenHeight];
+    if (isLandscape || _cachedYViewPort > kViewportScale)
+        [mapPanel.mapViewController setViewportScaleX:isLandscape ? kViewportBottomScale : kViewportScale y:kViewportScale];
 }
 
 - (void)restoreMapViewPort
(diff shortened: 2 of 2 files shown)


Sources/Controllers/QuickAction/OAFloatingButtonsHudViewController/OAFloatingButtonsHudViewController.mm
```diff
@@ -510,8 +504,8 @@ - (void)adjustMapViewPort
 {
     OAMapPanelViewController *mapPanel = [OARootViewController instance].mapPanel;
     BOOL isLandscape = [OAUtilities isLandscape];
-    [mapPanel.mapViewController setViewportScaleX:isLandscape ? kViewportBottomScale : kViewportScale
-                                                y:isLandscape ? kViewportScale : (DeviceScreenHeight - _actionsView.frame.size.height) / DeviceScreenHeight];
+    if (isLandscape || _cachedYViewPort > kViewportScale)
```
tigrim, line 507: Reset the horizontal viewport when returning to portrait

With Position on map set to Centre, `_cachedYViewPort` is 1. Opening the sheet in landscape (or rotating to landscape with it open) sets the horizontal viewport scale to 1.5. When rotating back to portrait, both conditions here are false, so the horizontal scale remains 1.5: the map target and pin stay at three quarters of the screen width until the sheet closes. Please reset the X scale independently of the condition on the cached Y scale.

Sources/Controllers/Map/OAMapViewTrackingUtilities.mm
```diff
@@ -803,6 +805,9 @@ - (BOOL) shouldAutoZoom:(CLLocation *)location autoZoomFrequency:(int)autoZoomFr
 {
     if (![_settings.autoZoomMap get] || location.speed <= 0)
         return NO;
+
+    if ([self isQuickActionsSheetVisible])
```
tigrim, line 809: Cancel an active auto-zoom animation when opening the sheet

This guard prevents new auto-zoom calculations, but an animation started before the sheet opened can still change the zoom. With `zoomParams == nil`, `startMoving:` sets `animateZoom` to NO and clears its local `zoomAnimation` pointer before the cancellation block, leaving the existing location-services Zoom animation registered. `setZoom:` does not cancel it, and `resume()` lets it continue. Opening Quick Actions during a smooth auto-zoom transition therefore does not preserve the zoom at the moment the sheet opens. Please cancel the active location-services Zoom animation when opening the sheet, while keeping location following and rotation active.

Sources/Controllers/QuickAction/OAFloatingButtonsHudViewController/OAFloatingButtonsHudViewController.mm
```diff
@@ -510,8 +504,9 @@ - (void)adjustMapViewPort
 {
     OAMapPanelViewController *mapPanel = [OARootViewController instance].mapPanel;
     BOOL isLandscape = [OAUtilities isLandscape];
-    [mapPanel.mapViewController setViewportScaleX:isLandscape ? kViewportBottomScale : kViewportScale
-                                                y:isLandscape ? kViewportScale : (DeviceScreenHeight - _actionsView.frame.size.height) / DeviceScreenHeight];
+    [mapPanel.mapViewController setViewportScaleX:isLandscape ? kViewportBottomScale : kViewportScale];
+    if (isLandscape || _cachedYViewPort > kViewportScale)
+        [mapPanel.mapViewController setViewportScaleY:kViewportScale];
```
tigrim: [P2] Keep the selected point above the sheet on smaller screens

In portrait, the map center no longer accounts for the sheet height. With a 667 pt screen height and no top or bottom widgets, `OAQuickActionsSheetView.adjustFrame` places the sheet's top edge at y=327, while the new pin position puts its tip at y=333.5, underneath the sheet. Top widgets shift the center farther down and can hide the entire pin. This prevents users from seeing the exact point they are selecting. Please constrain the map center to the available area above the sheet. These coordinates follow from the layout formulas and XIB dimensions; this was checked statically, without running the app.

Sources/Controllers/Map/OAMapViewTrackingUtilities.mm
```diff
@@ -555,7 +557,7 @@ - (void) startMoving:(double)finalLat finalLon:(double)finalLon zoomParams:(OAAu
         if (userZoomAnimation)
             animateZoom = NO;
         
-        if (!animateZoom)
+        if (!animateZoom && ![self isQuickActionsSheetVisible])
```
tigrim: [P2] Cancel the active auto-zoom animation when the sheet opens

The updated condition allows `startMoving:` to cancel an existing auto-zoom animation, but opening the sheet does not call this method. If Quick Actions is opened during a smooth auto-zoom transition, the renderer's display link continues advancing that animation until the next location update reaches `startMoving:`. The animation may even finish before then, so the zoom at the moment the sheet opens is still not preserved. Please cancel the active location-services Zoom animation directly when opening the sheet, while keeping location following and rotation active.

tigrim (approved):

Sources/Controllers/Map/OAMapViewTrackingUtilities.mm
```diff
@@ -813,6 +818,12 @@ - (BOOL) shouldAutoZoom:(CLLocation *)location autoZoomFrequency:(int)autoZoomFr
         return (now - _lastTimeAutoZooming) > autoZoomFrequency;
 }
 
+- (BOOL)isQuickActionsSheetVisible
+{
+    OAFloatingButtonsHudViewController *floatingButtonsController = [OARootViewController instance].mapPanel.hudViewController.floatingButtonsController;
```
alex-dev-neo: Method `isQuickActionsSheetVisible` should be moved to `hudViewController`

Sources/Controllers/QuickAction/OAFloatingButtonsHudViewController/OAFloatingButtonsHudViewController.mm
```diff
@@ -664,6 +659,7 @@ - (void)showActionsSheetAnimated:(OAHudButton *)quickActionButton
         [self adjustMapViewPort];
     }];
     [self restorePinPosition];
+    [OARootViewController instance].mapPanel.mapViewController.mapView.mapAnimator->cancelCurrentAnimation(kLocationServicesAnimationKey, OsmAnd::MapAnimator::AnimatedValue::Zoom);
```
alex-dev-neo: Introduce method `cancelAutoZoomAnimation` at `mapViewController`

PR #5766, opened 2026-09-11
Unify coordinate format widgets and grid

### LINKED ISSUE #5512: Unify coordinate format picker across Settings, Search and Widget

vitaliy-sova-ios:
## Related issue / task

Fixes https://github.com/osmandapp/OsmAnd-iOS/issues/5512

## Summary

Unifies coordinate format handling for the coordinates widgets and the map grid:
the grid format preference moves from a legacy int enum to string format ids
(`builtin:*` / `epsg:*`), matching Android. Grid projection/format mapping to the
core enums is centralised in `OAGridFormatMapping.h`, and the per-widget
coordinate format becomes a profile preference.

## Testing

**Tested on:**

- Device: iPhone 11
- iOS: 27.0

### Scenarios

- Coordinates grid: switch format (incl. MGRS, OLC, Maidenhead, Swiss Grid, custom EPSG), zoom limits
- Coordinates widgets: per-widget format, adding a widget for a non-active profile
- Upgrade from a build with the legacy `coordinateGridFormat` int preference

## Relevant checks

- [x] Portrait / Landscape tested
- [x] Light / Dark mode tested
- [x] Different profiles tested
- [x] Localization / RTL checked
- [x] VoiceOver / Accessibility checked
- [ ] CarPlay checked
- [x] Android parity checked

## AI disclaimer

**Implementation:**

- Tool / Agent: Claude Code
- Model: Opus 5

1. Port the changes from branch `fix_coordinates_grid_mgrs_projection` into this branch.
2. Review the branch diff for bugs, regressions and code style, then fix the findings.

- `OAGridFormatMappingBridge` added so Swift gets core enum raw values without mirroring them
- Legacy int migration moved out of the preference getter into a one-shot `MigrationManager` step
- Android parity fixes: Maidenhead granularity 6.0, OLC max zoom 18, grid search limit 50
- Toast instead of a silent no-op when a chosen format cannot be resolved
- EPSG catalog / transformer caching; dead `GridFormatWrapper` API removed

**Final review:**

- Tool / Agent: Claude Code
- Model: Opus 5
- [x] Final diff reviewed

**Significant findings:**

- Legacy grid format values were migrated through the wrong constant table (DMS/DDD/MGRS
  resolved to the wrong format) — fixed
- Supported grid max zoom was clamped to 14 for every format instead of 22 — fixed
- Grid SQL is assembled by string concatenation; separators are now explicit, since a
  trailing-whitespace cleanup silently broke the queries during development


Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
```diff
@@ -95,10 +167,198 @@ final class EpsgCatalogRepository {
...
+        baseAuth: String,
+        baseCode: String
+    ) -> [Int]? {
+        let sql = """
+            SELECT h.code FROM helmert_transformation h \
+            WHERE h.auth_name = 'EPSG' AND IFNULL(h.deprecated, 0) = 0 \
+            AND h.source_crs_auth_name = ? AND h.source_crs_code = ? \
+            AND h.target_crs_auth_name = 'EPSG' AND h.target_crs_code = '4326' \
+            AND h.method_auth_name = 'EPSG' AND h.method_code IN (\(Self.supportedHelmertMethods)) \
+            ORDER BY CASE WHEN lower(IFNULL(h.description, '')) LIKE '%replaced by%' THEN 1 ELSE 0 END, \
```
tigrim, line 325: Please account for the transformation's area of use before ranking and limiting these candidates. For example, for EPSG:26715 (NAD27 / UTM zone 15N), this query returns operation 7698 first, but that operation is valid only in Panama, which does not overlap the CRS's area of use. When the core falls back to the supplied operation code, `resolveProjectionParameters` accepts the first successfully decoded Helmert transformation and applies it across the grid, producing an incorrect datum shift. Successful parameter extraction does not establish that the transformation is geographically applicable.

Sources/Controllers/Settings/Coordinates/OAEpsgCoordinateTransformer.mm
```diff
@@ -94,4 +116,56 @@ - (CLLocation *)toLonLatWithCode:(NSInteger)epsgCode easting:(double)easting nor
...
+- (OAEpsgGridConstants *)constantsForCode:(NSInteger)epsgCode projectionRaw:(NSInteger)projectionRaw
+{
+    auto *transformer = [self transformerForCode:(int)epsgCode];
+    if (!transformer)
+        return nil;
+
+    auto projection = static_cast<OsmAnd::GridConfiguration::Projection>(projectionRaw);
+
+    OsmAnd::PointD lonBounds, latBounds, semiMajor, refLonLat, falseEN, scaleFactor;
+    if (!transformer->getConstants(projection, lonBounds, latBounds, semiMajor, refLonLat, falseEN, scaleFactor))
```
tigrim, line 128: Please normalize the CRS units, or exclude non-metric CRSs from grid selection until they are supported. The new catalog accepts EPSG:2222 (NAD83 / Arizona East (ft)), but the core's `getConstants()` reads linear parameters through `GetProjParm()` and divides them by 1000 without converting feet to meters. Its false easting of 700000 ft therefore becomes 700 km instead of 213.36 km. These values pass the validation below and are forwarded to the renderer, so the grid coordinates and labels are incorrect. The conversion and label units need to be handled consistently.

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
```diff
@@ -95,10 +223,219 @@ final class EpsgCatalogRepository {
...
+            AND hu.object_auth_name = h.auth_name AND hu.object_code = h.code \
+            LEFT JOIN extent he ON he.auth_name = hu.extent_auth_name AND he.code = hu.extent_code \
+            AND IFNULL(he.deprecated, 0) = 0 \
+            WHERE h.auth_name = 'EPSG' AND IFNULL(h.deprecated, 0) = 0 \
+            AND h.source_crs_auth_name = ? AND h.source_crs_code = ? \
+            AND h.target_crs_auth_name = 'EPSG' AND h.target_crs_code = '4326' \
+            AND h.method_auth_name = 'EPSG' AND h.method_code IN (\(Self.supportedHelmertMethods)) \
+            GROUP BY h.code, h.accuracy, h.description, a.area) \
+            SELECT code FROM candidates WHERE intersects = 1 \
+            ORDER BY IFNULL(ROUND(overlap / NULLIF(crs_area, 0.0), 1), 0.0) DESC, \
```
tigrim, line 444: Selecting the transformation with the largest bounding-box overlap still applies one regional datum shift to the entire CRS. For EPSG:26715, this ordering puts operation 1172 (`NAD27 to WGS 84 (3)`, Canada) first because it covers about 61% of the CRS bounding box. `resolveProjectionParameters` then uses that first operation for the full grid, including the CONUS part where operations 1173/1175 are applicable instead. The CRS bounds passed to the renderer are not reduced to the selected operation's area, so coordinates in the US receive the Canadian shift. Please select the operation for the rendered location, constrain the grid to the selected operation's area, or reject CRSs that require multiple regional operations.

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
```diff
@@ -24,24 +32,143 @@ final class EpsgCatalogRepository {
...
+    private static let metricAxesFilter = """
+        AND EXISTS (SELECT 1 FROM axis metric_axis \
+        WHERE metric_axis.coordinate_system_auth_name = crs.coordinate_system_auth_name \
+        AND metric_axis.coordinate_system_code = crs.coordinate_system_code \
+        AND metric_axis.uom_auth_name = 'EPSG' AND CAST(metric_axis.uom_code AS INTEGER) = \(metreUomCode)) \
+        AND NOT EXISTS (SELECT 1 FROM axis other_axis \
+        WHERE other_axis.coordinate_system_auth_name = crs.coordinate_system_auth_name \
+        AND other_axis.coordinate_system_code = crs.coordinate_system_code \
+        AND (IFNULL(other_axis.uom_auth_name, '') <> 'EPSG' \
+        OR IFNULL(CAST(other_axis.uom_code AS INTEGER), 0) <> \(metreUomCode)))
```
tigrim, line 69: Filtering the output axes to metres fixes linear output units, but it does not validate the angular units of the conversion parameters consumed by the core. EPSG:22780 passes this filter, while its latitude and longitude of origin are stored in grads (EPSG:9105). `CoordinateTransformer::getConstants()` reads them with `GetProjParm()` and multiplies by pi/180; GDAL reserves `GetNormProjParm()` for conversion to degrees. Consequently, 38 grads is treated as 38 degrees instead of 34.2 degrees and the generated grid is misplaced. Please normalize angular projection parameters in the core or exclude CRSs whose required angular parameters are not in degrees.

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
vitaliy-sova-ios: Went with rejecting such CRSs: a CRS is now accepted only if a single operation covers everything the applicable operations cover inside its area of use (1% tolerance), otherwise it's marked unsupported and dropped from the grid format list as well. That removes 67 formats (EPSG:26715, 26716, ED50 / UTM 30N, Beijing 1954 and Aratu zones, …) while keeping cases like OSGB36 / British National Grid, where the bbox is only wider than the single available operation because of an offshore gap; list and resolver stay in sync — 2587 formats listed, 0 listed-but-unresolvable.

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
vitaliy-sova-ios: Added a units filter on the conversion parameters as well: all seven parameter slots must be in EPSG:9001 (metre), 9102/9122 (degree), 9110 or 9201 (unity), applied both to the grid format list and to getGridDefinition. Worth noting that 9110 (sexagesimal DMS) doesn't need excluding — PROJ decodes it to decimal degrees itself (EPSG:2039 comes back as 31.7343936, not the stored 31.4403817), so only genuinely different units are a problem. That drops exactly two formats, EPSG:22780 and EPSG:29702, both with origin parameters in grads; linear parameters outside metres only occur on CRSs whose axes aren't metres either, which the axis filter already excluded.

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
```diff
@@ -95,10 +245,313 @@ final class EpsgCatalogRepository {
...
+        var inside = [Int: [EpsgArea]]()
+        for (code, areas) in operationAreas {
+            let clipped = areas.compactMap { $0.clipped(to: crsArea) }
+            if !clipped.isEmpty {
+                inside[code] = clipped
+            }
+        }
+        guard !inside.isEmpty else { return false }
+
+        let covered = EpsgArea.unionArea(inside.values.flatMap { $0 })
```
tigrim: This denominator only measures the union of the available transformation extents, so a single operation that covers a tiny sliver of the CRS still passes. For example, EPSG:2164 covers the offshore Côte d'Ivoire bbox (-7.55, 1.02)–(-3.11, 5.19), while its only operation, EPSG:1469, overlaps it only in (-4.22, 5.15)–(-3.85, 5.19), about 0.08% of the CRS bbox; `best == covered`, so the format is listed and that local operation is then used for coordinates throughout the CRS area. Please compare the selected operation's clipped coverage with the CRS area (or otherwise reject coordinates outside its extent), rather than comparing it only with the union of operations.

Sources/Controllers/Settings/Coordinates/EpsgCatalogRepository.swift
```diff
@@ -24,24 +34,163 @@ final class EpsgCatalogRepository {
         LEFT JOIN extent e ON e.auth_name = u.extent_auth_name AND e.code = u.extent_code \
         AND IFNULL(e.deprecated, 0) = 0 
         """
-    
+
+    private static let gridBaseSelect = """
```
tigrim: Could this catalog logic be moved to OsmAnd-shared instead of introducing an iOS-specific implementation? Android already has its own `EpsgCatalogRepository.kt`, while OsmAnd-shared already provides `SQLiteAPI.openByAbsolutePath()` on both Android and iOS. Keeping the SQL, supported-method rules, transformation ranking, area validation, and models in common code would prevent the two implementations from drifting and would let fixes apply to both platforms. Only locating `proj.db` and the renderer bridge need to remain platform-specific. Longer term, an even cleaner boundary would be a high-level catalog API in OsmAnd Core next to PROJ, so application code does not depend directly on PROJ's internal database schema.

tigrim (commented):
One architectural concern after reviewing the database integration:

Sources/Controllers/Map/Widgets/Factory/MapWidgetsFactory.swift
```diff
@@ -41,11 +41,11 @@ final class MapWidgetsFactory: NSObject {
         case .secondNextTurn:
             return OANextTurnWidget(horisontalMini: true, nextNext: true, customId: customId, appMode: appMode, widgetParams: widgetParams)
         case .coordinatesCurrentLocation:
-            let widget = CoordinatesCurrentLocationWidget()
+            let widget = CoordinatesCurrentLocationWidget(customId: customId, appMode: appMode, widgetParams: widgetParams)
```
tigrim, line 44: ?

Sources/Controllers/Map/Widgets/Factory/MapWidgetsFactory.swift
```diff
@@ -41,11 +41,11 @@ final class MapWidgetsFactory: NSObject {
...
             widget.delegate = OARootViewController.instance().mapPanel.hudViewController?.mapInfoController
             return widget
         case .coordinatesMapCenter:
-            let widget = CoordinatesMapCenterWidget()
+            let widget = CoordinatesMapCenterWidget(customId: customId, appMode: appMode, widgetParams: widgetParams)
```
tigrim, line 48: ?

Sources/Controllers/Map/Widgets/CoordinatesBaseWidget.swift
```diff
@@ -50,24 +69,78 @@ class CoordinatesBaseWidget: OABaseWidgetView {
...
+            storedId = value
+        } else {
+            storedId = coordinateFormatPref.get(appMode)
+        }
+
+        let format = resolveFormat(storedId, appMode: appMode)
+
+        let row = section.createNewRow()
+        row.cellType = OAValueTableViewCell.getIdentifier()
+        row.key = "coordinate_format"
```
tigrim, line 102: to rowKey

tigrim:
Static review findings (no build run):

1. **[P1] Reset projection parameters when leaving a custom EPSG grid.** In [`OACoordinatesGridLayer.mm` line 262](https://github.com/osmandapp/OsmAnd-iOS/blob/2a67b9df2314ce91c64d9b74c7fc5ea311af4823/Sources/Controllers/Map/Layers/OACoordinatesGridLayer.mm#L262), the reused `GridConfiguration` updates projection constants and Helmert parameters only when `hasProjectionParameters` is true. After selecting a custom EPSG and then UTM/MGRS, the previous values remain. Core's `setProjectionParameters()` does not clear all of these fields for UTM/MGRS, so the grid can use the wrong projection until the layer is recreated. Please reset the secondary parameters before applying the next format.

2. **[P2] Use the same support criteria in the grid picker and resolver.** [`CoordinateGridFormatProvider.swift` line 62](https://github.com/osmandapp/OsmAnd-iOS/blob/2a67b9df2314ce91c64d9b74c7fc5ea311af4823/Sources/Controllers/Settings/Coordinates/CoordinateGridFormatProvider.swift#L62) treats any EPSG catalog definition as supported, but `resolve()` also requires a mapped projection method and usable projection/ellipsoid parameters. An entry can therefore appear in the grid picker and fail with an “Unavailable” toast when selected. Please filter with the same condition used by `resolve()`.

Sources/Controllers/Map/Widgets/CoordinatesBaseWidget.swift
```diff
@@ -216,3 +349,20 @@ class CoordinatesBaseWidget: OABaseWidgetView {
...
 }
+
+extension CoordinatesBaseWidget: CoordinateFormatSelectorDelegate {
+
+    func coordinateFormatSelector(_ selector: CoordinateFormatSelectorViewController, didSelectFormatId formatId: String) {
+        applySelectedFormat(formatId)
+    }
+
+    func coordinateFormatSelectorDidRequestOtherFormat(_ selector: CoordinateFormatSelectorViewController) {
+        guard let vc = configViewController else { return }
```
tigrim, line 360: guard let configViewController

Sources/Controllers/Settings/Coordinates/CoordinateFormatSelectorViewController.swift
```diff
@@ -258,14 +280,53 @@ final class CoordinateFormatSelectorViewController: OABaseNavbarViewController {
...
+                                addMode: CoordinatesFormatAddViewController.AddMode,
+                                focusSearch: Bool,
+                                onSelected: @escaping (String) -> Void) {
+        let addVC = CoordinatesFormatAddViewController(
+            appMode: appMode,
+            excludedIds: excludedIds,
+            addMode: addMode,
+            focusSearch: focusSearch
+        )
         addVC.onFormatAdded = { id in
```
tigrim, line 321: weak?

Sources/Controllers/Settings/Coordinates/CoordinateGridFormat.swift
```diff
@@ -0,0 +1,56 @@
...
+//  Created by Vitaliy Sova on 20.08.2026.
+//  Copyright © 2026 OsmAnd. All rights reserved.
+//
+
+import Foundation
+
+struct CoordinateGridPoint {
+    let x: Double
+    let y: Double
+    static let zero = CoordinateGridPoint(x: 0, y: 0)
```
tigrim, line 14: code style

Sources/Controllers/Settings/Coordinates/CoordinateGridFormatBridge.swift
```diff
@@ -0,0 +1,98 @@
...
+            scaleFactor = p.scaleFactor.x
+            scaleFactorY = p.scaleFactor.y
+            let e = p.ellipsoidParameters
+            translationsX = e.translationsXY.x; translationsY = e.translationsXY.y
+            translationsZ = e.translationsZW.x; translationsW = e.translationsZW.y
+            rotationsX = e.rotationsXY.x; rotationsY = e.rotationsXY.y
+            rotationsZ = e.rotationsZScale.x; ellipsoidScale = e.rotationsZScale.y
+        } else {
+            hasProjectionParameters = false
+            lonMin = 0; lonMax = 0; latMin = 0; latMax = 0
```
tigrim, line 76: don't use ;

Sources/Controllers/Settings/Coordinates/CoordinateGridFormatProvider.swift
```diff
@@ -0,0 +1,183 @@
...
+            seen.insert(n)
+            result.append(n)
+        }
+        return result
+    }
+
+    // MARK: - Private
+
+    private func normalizeId(_ formatId: String?) -> String? {
+        if let n = CoordinateFormatIds.normalize(formatId) { return n }
```
tigrim, line 79: n?

Sources/Controllers/Settings/Coordinates/CoordinateGridFormatProvider.swift
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+        return nil
+    }
+
+    private func readProjectionConstants(
+        _ epsgCode: Int,
+        projectionRaw: Int32
+    ) -> CoordinateGridProjectionConstants? {
+        guard let c = OAEpsgCoordinateTransformer.sharedInstance()
```
tigrim, line 128: c?

Sources/Controllers/Settings/Coordinates/CoordinateGridFormatProvider.swift
```diff
@@ -0,0 +1,183 @@
...
+            falseEastingAndNorthing: CoordinateGridPoint(x: c.falseEasting, y: c.falseNorthing),
+            scaleFactor: CoordinateGridPoint(x: c.scaleFactor, y: c.scaleFactorY)
+        )
+    }
+
+    private func readEllipsoidParameters(
+        _ epsgCode: Int,
+        operationCode: Int
+    ) -> CoordinateGridEllipsoidParameters? {
+        guard let p = OAEpsgCoordinateTransformer.sharedInstance()
```
tigrim, line 146: ?

Sources/Helpers/GridFormat.swift
```diff
@@ -87,12 +122,54 @@ enum GridLabelsPosition: Int32, CaseIterable {
...
-        format != .utm && format != .mgrs
+
+    static func migratePreferenceValue(_ value: Any?) -> String {
+        if let stringValue = value as? String {
+            if let normalized = CoordinateFormatIds.normalize(stringValue) {
+                return normalized
+            }
+
+            switch stringValue {
+            case "DD_MM_SS":
```
tigrim, line 136: to enum

Sources/Helpers/GridFormat.swift
```diff
@@ -87,12 +122,54 @@ enum GridLabelsPosition: Int32, CaseIterable {
...
+        if let numberValue = value as? NSNumber,
+           let legacyId = legacyStoredGridFormatId(numberValue.intValue) {
+            return legacyId
+        }
+        return CoordinateFormatIds.builtinDdd
+    }
+
+    private static func legacyStoredGridFormatId(_ storedRawValue: Int) -> String? {
+        switch storedRawValue {
+        case 0:
```
tigrim, line 161: use enum

tigrim:
**Question about the grid format default:** In [`OACommonGridFormat.getProfileDefaultValue`](https://github.com/osmandapp/OsmAnd-iOS/blob/2a67b9df2314ce91c64d9b74c7fc5ea311af4823/Sources/Helpers/OAAppSettings.m#L5672-L5675), an unset grid format follows the profile's current primary coordinate format. This means changing the primary format can also change an enabled grid without selecting a new grid format. Android keeps `COORDINATE_GRID_FORMAT` as a separate string preference with `builtin:ddd` as its default and writes the migrated value per profile. Is this coupling intentional? If the grid should retain its own format, the iOS default/migration needs to store an initial ID rather than derive it on every read. This is separate from the projection-state and picker-support findings above.

Sources/Helpers/OACoordinatesGridSettings.mm
```diff
@@ -18,6 +18,7 @@ @implementation OACoordinatesGridSettings
     OsmAndAppInstance _app;
     OAAppSettings *_settings;
     NSInteger _supportedMaxZoom;
+    NSMutableDictionary<NSString *, NSValue *> *_supportedZoomByFormatId;
```
tigrim, line 21: Could we rename _supportedZoomByFormatId to _zoomRangeByFormatId? The current name sounds like it stores a single zoom value, while NSValue actually represents a zoom range

Sources/Controllers/Map/Widgets/Factory/MapWidgetsFactory.swift
vitaliy-sova-ios: The three parameters are what the per-widget format preference needs. The
preference key is built from the widget type plus customId
("coordinates_widget_format_<type>[_<customId>]"), so a second copy of the
widget gets its own format instead of sharing one; appMode is the profile the
value is read from and written to; and widgetParams carries the format chosen
in the creation dialog, which the initialiser stores for the target profile
before the view is built. Same shape as the other widgets in this switch -
OANextTurnWidget and the rest already take customId/appMode/widgetParams.

Sources/Controllers/Map/Widgets/Factory/MapWidgetsFactory.swift
vitaliy-sova-ios: The three parameters are what the per-widget format preference needs. The
preference key is built from the widget type plus customId
("coordinates_widget_format_<type>[_<customId>]"), so a second copy of the
widget gets its own format instead of sharing one; appMode is the profile the
value is read from and written to; and widgetParams carries the format chosen
in the creation dialog, which the initialiser stores for the target profile
before the view is built. Same shape as the other widgets in this switch -
OANextTurnWidget and the rest already take customId/appMode/widgetParams.

PR #5767, opened 2026-09-11
CarPlay Map mode to Vehicle appearance by default

### LINKED ISSUE #5398: CarPlay Appearance Overrides OsmAnd Map Mode Setting

vitaliy-sova-ios:
## Related issue / task

Fixes https://github.com/osmandapp/OsmAnd-iOS/issues/5398

## Summary

Removed CarPlay map appearance migration. Vehicle appearance is now used by default until changed manually.

## Testing

Not run.

**Tested on:**

- Device: iPhone 11
- iOS: 26.0

### Scenarios

- Verified Vehicle appearance remains the default value.

## Relevant checks

- [x] CarPlay checked
- [ ] Android parity checked

## AI disclaimer

**Implementation:**

- Tool / Agent: Cursor Agent
- Model: GPT-5.6 Sol

1. Remove the CarPlay appearance migration.
2. Keep Vehicle appearance as the default until manually changed.

- None

**Final review:**

- Tool / Agent: Cursor Agent
- Model: GPT-5.6 Sol
- [x] Final diff reviewed

**Significant findings:** None

### DIFF
--- a/Sources/Helpers/MigrationManager.swift
+++ b/Sources/Helpers/MigrationManager.swift
@@ -23,7 +23,6 @@ final class MigrationManager: NSObject {
         case migrateRouteRecalculationValues
         case migrateLocationIconSizeAndCourseIconSize
         case migrateAstronomyPreferences
-        case migrateCarPlayMapAppearanceMode
         case migrateTracksSortModeKeysAndFormat
     }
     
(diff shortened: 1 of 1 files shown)


tigrim:
Removing this migration prevents future installations from inheriting the profile map mode, but it does not address users whose settings were already migrated by an earlier version. The migration persisted Day/Night/Auto in `carplay_map_appearance_mode`, and that stored value still takes precedence over the Vehicle appearance default after this update—even if the user never manually selected a CarPlay map mode.

Could we also handle previously migrated values while preserving explicit user selections? Otherwise, affected users will continue seeing the old mode after upgrading.

tigrim (approved):

PR #5772, opened 2026-09-13
Benchmark the C++ router against the OsmAndShared planner inside the app

alex-dev-neo:
Companion to the routing-to-OsmAnd-shared series in the android repository (osmandapp/OsmAnd#25999 to #26026). It needs the planner classes from that series in `OsmAndShared.framework` (`OASRoutePlannerFrontEnd`, `OASBinaryMapIndexReader`, `OASRoutingConfiguration`), so it compiles only against a shared module built from `kmp-routing-26-native-tuning` or later; it is here for review and for the numbers, and lands after the series does.

`OARoutePlannerBenchmarkTest` answers the question gate 1 of that series asks: inside the app, on the routes the jvm and Kotlin/Native benchmarks use (Noord-Holland, Upper Bavaria, Lower Austria), how does the shared planner compare with the C++ router the app runs today? It runs each route two ways - the C++ router the way `OARouteProvider` drives it (car profile from the app's `routing.xml`, complex mode, 256 MB, one regional map per route opened as `checkInitialized` opens it) and the shared planner from the framework with the same inputs - and reports wall time, segments, distance, the segments the search settled, the tiles it loaded and the process footprint after each. The report goes into the result bundle as activities and an attachment: the host's log goes nowhere when the host is the Mac.

Not part of a normal test run: it takes a minute and only prints numbers. It is run alone, and in Release, because Debug links the -O0 core and the debug Kotlin framework:

```
xcodebuild test -workspace OsmAnd.xcworkspace -scheme 'OsmAnd Maps' -configuration Release \
  -destination 'platform=macOS,variant=Designed for iPad' ENABLE_TESTABILITY=YES \
  -only-testing:'OsmAnd MapsTests/OARoutePlannerBenchmarkTest'
xcrun xcresulttool export attachments --path <xcresult> --output-path <dir>
```

`ENABLE_TESTABILITY` because without `-export_dynamic` the app's classes are undefined for the test bundle in Release; and the test targets' Release signing is "iPhone Distribution", which a development Mac does not have - for the run it has to be switched to Automatic / Apple Development the way Debug has it. That is left as it is here, not changed for one test.

What it found on one Mac (Designed for iPad, best of two after a warmup, ms):

| route | C++ router | shared planner |
|---|---|---|
| Amsterdam to Schiphol 17 km | 1 114 | 858 |
| Amsterdam to Haarlem 20 km | 1 224 | 990 |
| Amsterdam to Den Helder 80 km | 820 | 972 |
| Munich to the airport 35 km | 639 | 791 |
| Munich to Rosenheim 65 km | 1 656 | 1 951 |
| Munich to Garmisch 90 km | 650 | 728 |
| Vienna to Schwechat 20 km | 1 890 | 1 838 |
| St. Pölten to Krems 30 km | 392 | 514 |
| St. Pölten to Wiener Neustadt 70 km | 849 | 768 |

The shared planner takes 0.8 to 1.3 of the C++ router's time. Its segments, distances and visited counts are the same as on the jvm and on the simulator; the C++ router settles 5 to 20% fewer segments on the routes where it is faster, and on St. Pölten to Wiener Neustadt it picks a slightly different road (449 segments, 97.8 km against 446 and 96.7 km everywhere else).

A second table, the gpx approximation (osmandapp/OsmAnd#26038): the same nine routes as tracks - the C++ router's route, its road points thinned to one every 15 m and moved two metres off the nodes, as a recording never lies on them - attached back to the roads by the C++ core and by the shared copy, the routing-based and the geometry-based way (the latter is what `OARouteProvider.calculateGpxApproximation` runs). Wall time in ms, Release, Designed for iPad:

| route | C++, routing-based | shared, routing-based | C++, geometry-based | shared, geometry-based |
|---|---|---|---|---|
| Amsterdam to Schiphol 17 km | 86 | 94 | 19 | 20 |
| Amsterdam to Haarlem 20 km | 96 | 105 | 18 | 19 |
| Amsterdam to Den Helder 80 km | 164 | 179 | 39 | 51 |
| Munich to the airport 35 km | 223 | 214 | 28 | 31 |
| Munich to Rosenheim 65 km | 304 | 288 | 38 | 40 |
| Munich to Garmisch 90 km | 246 | 253 | 42 | 44 |
| Vienna to Schwechat 20 km | 923 | 807 | 17 | 17 |
| St. Pölten to Krems 30 km | 138 | 134 | 18 | 21 |
| St. Pölten to Wiener Neustadt 70 km | 212 | 222 | 43 | 46 |

The shared copy takes 0.87 to 1.09 of the C++ core's time on the routing-based way and 1.03 to 1.32 on the geometry-based one, a few milliseconds either way; both attach the same segments to the same track points and settle the same segments (Schiphol 205 segments over 429 points, 1 831 settled), Vienna being the slow one for both on the routing-based way (50 374 settled).


alex-dev-neo:
Closing: the comparison this was built for is done - the two planners were measured side by side inside the app for gates 1 to 3, and the routing itself landed in master with #5781. The harness is not worth carrying in the test bundle, and the branch stays, so it can be brought back if the two planners need to be measured against each other again.

PR #5792, opened 2026-09-15
Fix Plan Route undo/redo crashes and stale lines

### LINKED ISSUE #5782: Incorrect route line rendering after trimming route and crash on Undo/Redo actions
### Description

When trimming a route to a specific point, the removed segment of the route line remains visually rendered on the map. Additionally, performing subsequent Undo and Redo operations causes the application to crash unexpectedly

[Crash logs.zip](https://github.com/user-attachments/files/32194206/Archive.15.zip)

### Steps to reproduce

1. Open the app and start planning/editing a route (Plan Route).
2. Create/load a route with multiple points.
3. Trim the route to a specific point ("Trim route to point" / "Trim before/after").
4. Observe the rendered route line on the map.
5. Tap the Undo button.
6. Tap the Redo button (or alternate Undo/Redo).

### Actual result

Step 4: The trimmed part of the route line is still displayed on the map, even though the route points were removed.

Step 6: The app crashes after interacting with the Undo/Redo buttons following the trim operation.

<video src="https://github.com/user-attachments/assets/d460bf8b-923a-421e-971a-81b11819c80a"></video>
</details> 

### Expected result

- [ ] Step 4: The line corresponding to the trimmed route segment is removed from the map view.

- [ ] Step 6: Undo/Redo operations update the route state correctly without causing an application crash.

### Your Environment
OsmAnd Version: 5457
Android/iOS version: 27
Device model: 15

aleksandr-tata:
## Related issue / task

Fixes #5782.

## Summary

Fix Plan Route crashes and incorrect track state after Trim → Add → Undo/Redo, and remove leftover lines after editing.

- Finish Trim inside the clear command so its first execution and Redo leave the same complete point list.
- Capture both editing halves for Clear/Split history and copy the road-geometry dictionary on capture and restoration.
- Reject invalid Trim/Delete indices before adding a command to history.
- Restore the previous point profile when undoing deletion at a segment boundary.
- Remove obsolete route lines when the number of rendered segments decreases.
- Ignore ordinary map taps during Move/insertion so they cannot commit an extra point; explicit Add one more, Apply and Cancel remain available.
- Add 23 regression tests. Temporary diagnostic logging has been removed.

## Testing

**Tested on:**

- Device: iPhone 11 — iOS 26.0.1 (manual verification by the author).
- Simulator: iPhone 14 Pro Max — iOS 26.5 (XCTest).

### Automated verification

- [x] Initial baseline: 15 of 20 new history/render tests failed, reproducing the Trim history crashes, geometry loss and stale lines.
- [x] Move follow-up baseline: 2 of 3 additional tests failed; ordinary tapping after Cancel passed.
- [x] After fixes: all 36 selected tests passed — 23 `OAPlanRouteHistoryTests` and 13 existing `OAPlanRouteEditingBridgeTest` tests.
- [x] After removing temporary diagnostics: Objective-C++ syntax check of the edited bridge passed; the command manager and base command match their original versions.
- [x] `plutil -lint OsmAnd.xcodeproj/project.pbxproj` and `git diff --check` passed.

The 36-test runtime run preceded the logging-only cleanup. History tests use the real editing context, commands and bridge with a headless map layer; the line-cleanup test uses the native line collection and waits for test-host initialization.

### Scenarios

Manually verified on the device:

- [x] Delete a point, insert before a middle point and Apply: unchanged sections retain their appearance; Undo/Redo restores the insertion.
- [x] Trim before/after: removed sections leave no residual lines, including after panning and zooming.
- [x] Trim → one or three additions → Undo/Redo → another Undo: correct point counts/order and no crash.
- [x] Two successive Trims with repeated Undo/Redo: the second snapshot remains valid.
- [x] Add one more with Apply/Cancel: the original track suffix is preserved.
- [x] Trim/Clear → Undo restores calculated road geometry; undoing deletion of a segment endpoint restores the original gap.
- [x] One-point and empty-track transitions remove and restore lines correctly.
- [x] Save a copy after Redo Trim and reopen it: retained track data is preserved.
- [x] Taps during Move do not add points; Cancel restores the original track, and ordinary tapping works after leaving Move.

## Relevant checks

- [ ] CarPlay checked
- [x] Android parity checked — source comparison of snapshots, editing-mode tap guards and renderer cleanup; Android runtime was not tested.

## AI disclaimer

**Implementation:**

- Tool / Agent: Codex
- Model: GPT-6

1. Investigate issue #5782 using the supplied videos, crash screenshot and QA reproduction steps.
2. Review an external analysis, formalize automatic/manual regression scenarios, and compare Android behavior.
3. Add failing tests, implement the fixes, rerun the tests and add temporary diagnostics.
4. Investigate the device-reported extra point during Move, reproduce it in tests and block conflicting map taps.
5. Remove temporary logs and prepare the PR using the team's format after manual verification.

- Normalize the point list within the Trim command instead of after every history operation, preserving normal insertion boundaries.
- Copy dictionary containers while retaining point identity for road-geometry keys.
- Prune obsolete lines only from the main route collection.
- Reject conflicting map taps before hit detection instead of extending point selection into an unfinished edit.
- Use a headless map layer for command tests and an initialization-aware native collection test for rendering.

**Final review:**

- Tool / Agent: Codex (same agent as implementation)
- Model: GPT-6
- [x] Final diff reviewed

**Significant findings:**

- The initial Trim merged the editing halves in the bridge; Redo bypassed that merge and left subsequent commands with incompatible indices.
- Clear history captured only the first half and shared mutable geometry storage; Undo could lose points or empty its own geometry snapshot.
- Deleting a segment endpoint changed the previous point's profile, which Undo did not restore.
- The renderer retained lines whose IDs were no longer generated.
- Move temporarily removes the selected point; tapping a point in the second half could fall through to Add and leave a ninth point after Cancel.
- Save during an unfinished edit, other operation-switching paths, asynchronous routing races and performance are outside the validated scope of this PR.


tigrim (approved):

alex-dev-neo:
Reviewed the change — the fixes are correct and target the actual cause, no blockers from my side. A few things below; the first one is the same defect this PR fixes, so it would be good to cover it here rather than in a separate PR.

### 1. The same road-segment aliasing bug is still present in two other commands

The root cause you fixed is that `roadSegmentData` is exposed as a live `NSMutableDictionary` (`OAMeasurementEditingContext.h:59`, no `copy`), while `clearSegments` → `clearSnappedToRoadPoints` calls `removeAllObjects` on the very object a command is holding as its snapshot. Android does not have this because `getRoadSegmentData()` returns `new HashMap<>(roadSegmentData)` and `setRoadSegmentData()` copies as well.

Two commands still capture the live dictionary:

- `OAReversePointsCommand.mm:37` — `_oldRoadSegmentData = editingCtx.roadSegmentData;`, and `executeCommand` immediately calls `clearSnappedToRoadPoints` on that same dictionary, so the snapshot is emptied during execute; `undo` (line 85) then restores an empty map. Reproducible: route snapped to roads → "Reverse route" from the options menu → Undo → calculated geometry is gone.
- `OAJoinPointsCommand.mm:38` — `_roadSegmentData = ctx.roadSegmentData;`, and `undo` calls `clearSegments` (line 46) *before* `setRoadSegmentData:` (line 47) — literally the same ordering `OAClearPointsCommand` had before this PR.

Both are one-liners (`copy` on capture, `mutableCopy` on restore). Alternatively, fix the root and make the `roadSegmentData` accessor defensive like Android's — broader in scope, but it removes the whole class of bugs.

### 2. `OARemovePointCommand.mm:48` — unguarded array access in `undo`

```objc
if (_position > 0)
{
    OASWptPt *previousPoint = ctx.getPoints[_position - 1];
```

`execute` got a bounds check, `undo` did not. Android guards both with `position > 0 && position <= points.size()`. `getPoints` is only the `_before` half and it shrinks outside the command stack (`splitSegments` when entering move/insert mode), so this can go out of range → `NSRangeException`. I could not find a path through the current UI (the point-editing view covers the Undo toolbar), but given the topic of this PR the symmetric check is worth adding.

### 3. `OAClearPointsCommand.mm:35` — `_pointPosition = ctx.selectedPointPosition;` runs before the `ctx == nil` check

Harmless in practice (messaging nil yields 0), but it reads like a bug — move the line below the guard.

### 4. Inconsistent bound in the bridge

`trimBeforeIndex:` / `trimAfterIndex:` validate the index against `getAllPoints.count`, while every neighbouring entry point (`showPointOptionsAtIndex:`, `addPointBeforeIndex:`, `addPointAfterIndex:`) uses `getPointsCount`. `selectedPointPosition` indexes `_before`, so `getPointsCount` is the semantically correct bound. Not a live bug — `_after` is empty on that path.

### 5. Unrelated formatting churn

`- (void) drawRouteSegments` → `- (void)drawRouteSegments`, `- (BOOL) execute` → `- (BOOL)execute`, `- (void) undo` → `- (void)undo`. These files use the spaced style throughout, so the reformatting only adds diff noise.

### 6. `testRendererRemovesLinesAfterJoiningEditingHalves`

It reaches into the private `_collection` ivar via `class_getInstanceVariable` plus manual `ivar_getOffset` pointer arithmetic and a `reinterpret_cast`. If the collection type ever changes, this breaks at runtime instead of at compile time. It also waits up to 60s for test-host initialization, which makes it an integration test living in the unit-test target. A test-only accessor exposing the line count would be sturdier.

### Android

Moving `splitSegments` inside the command diverges from Android, which has the same bug: `MeasurementToolFragment.java:1309` calls `editingCtx.splitSegments(...)` after `execute`, so redo after Trim is broken there as well. Worth filing an Android issue so the platforms do not drift.

---

The rest looks good to me: the line cleanup in `drawRouteSegments` is safe (`_collection` is used only there with ids from 100, `_lastLineCollection` is separate, and `getLines()` returns a `QList` by value so removing while iterating the copy is fine), the `onTouch` guard matches Android's `selectedPointPosition != -1` behaviour including taps on existing points, and `getPoints` → `getAllPoints` in `OASplitPointsCommand` fixes real loss of the second half on undo.

aleksandr-tata:
Addressed the review in four commits:

- `d1b0e0dc3` — Reverse and Join now copy road-geometry snapshots and restore mutable copies. The repeated Join test also exposed a lost gap profile causing an exception on Redo; Undo now restores that profile.
- `941f89b34` — Added the bounds check in Remove Undo, moved the Clear context guard before reading the selected point, used `getPointsCount` for the bridge Trim bounds, and reverted the unrelated signature formatting changes.
- `1fc45e3d9` — Replaced renderer-test ivar arithmetic with Debug-only typed access and isolated collection setup. The test exercises the real line collection (2 → 1 → 0) without waiting for app initialization.
- `b9efe5e7d` — Testing Join from the next segment exposed an `NSNotFound` versus `-1` mismatch in boundary detection. Fixed it and added coverage for both segment boundaries and both Join directions.

### Verification

- Before the fixes, all five new baseline tests failed, reproducing geometry loss, the Remove Undo exception, incorrect Trim acceptance, and the Join Redo exception.
- After the fixes, all **43 selected tests passed** on iPhone 14 Pro Max / iOS 26.5: 30 history tests and 13 bridge tests, with zero failures and `TEST SUCCEEDED`.
- The follow-up manual scenarios were completed with **no crashes reported**. Visual geometry, gap restoration, and saved-track results have not been separately confirmed yet.
- Diff validation and project-file lint passed. No temporary production logging was added.

alex-dev-neo (approved):
All six points from my earlier comment are addressed correctly, and I checked the new commits against Android. Approving.

Verified:

- `copy` / `mutableCopy` in `OAReversePointsCommand` and `OAJoinPointsCommand`.
- `OARemovePointCommand.undo` now guards with `_position > 0 && _position <= ctx.getPointsCount`, matching Android, and `testDeleteUndoWithShortenedBeforeHalfDoesNotThrow` covers exactly that case.
- nil-check ordering in `OAClearPointsCommand`, `getPointsCount` bound in the bridge, formatting reverted to the local style.
- The renderer test now uses `#if DEBUG` accessors instead of ivar offset arithmetic. `DEBUG=1` is defined at project level and inherited by the app target and both test targets, so the conditional compiles consistently.

Two things beyond my comment are good catches:

- Saving and restoring the gap point's profile in `OAJoinPointsCommand`. iOS `joinPoints:` mutates the original `WptPt` in place, while Android replaces it with a copy (`MeasurementEditingContext.java:650`), so undo really did lose the gap.
- `i != -1` → `i != NSNotFound` in `isBorderPointSelected:`. `NSNotFound == NSIntegerMax`, so an `NSInteger` result is never `-1` and the loop always answered from the first segment; Android's `-1` is correct only because Java's `indexOf` returns `-1`. I scanned `Sources/` for other `indexOfObject:` results compared against `-1` and found none.

Two asks before this is merged, since it targets `r5.4`:

1. The approval on this PR was submitted on `b8e5e543`, five hours before the four follow-up commits, so those have not been looked at by anyone but the author. @tigrim could you take another pass, at least over `b9efe5e7`?
2. `b9efe5e7` goes beyond the original issue: it touches shared `OAMeasurementEditingContext` and changes visible behaviour — on tracks with gaps, points that start or end the second and later segments now get "Join segments" where they previously got the split options (`OAPointOptionsBottomSheetViewController.mm:128`). The unit test covers the predicate, but the menu itself is not in the manual scenarios, the test run after these four commits is not reflected in the description, and there is no CI on the branch. Please re-run the suite, update the description (it still says "Add 23 regression tests", "all 36 selected tests passed", and that the line-cleanup test waits for test-host initialization), and eyeball the point menu on a track with a gap.

Non-blocking follow-ups, both unreachable through the current UI:

- `OAJoinPointsCommand.mm:40` still snapshots only the before half (`ctx.getPoints`) while its `undo` calls `clearSegments`, which clears both — the same pattern fixed in `OASplitPointsCommand` via `getAllPoints` in this PR. Harmless today because join is reached from the point menu with an empty after half. `OAReversePointsCommand` is fine here, it uses `clearPoints`.
- `OAMeasurementEditingContext.mm:620` initializes `gapPoint` to a non-nil `[[OASWptPt alloc] init]` where Android uses `null`, so the `if (gapPoint)` guard never fails; with `gapIndex == -1` (join at position 0) that reaches `removeObjectAtIndex:` with `-1`. Not reachable, because the options sheet skips the join item when `isFirstPointSelected:YES`. Aligning `joinPoints:` with Android — replace the point with a copy instead of mutating it — would also make the new `_gapPointProfile` workaround unnecessary.

I did not build or run the tests myself.

PR #5793, opened 2026-09-15
Fix main thread hang when opening the context menu for a large station

### LINKED ISSUE #5788: Main thread hang (9000+ ms) after opening POI context menu for a railway station
### Description

Tapping the railway station **Київ-Пасажирський** (Kyiv-Pasazhyrskyi, Kyiv) opens the POI context menu, and immediately after the menu is presented the main thread blocks. iOS hang detection shows one continuous hang that grows past its 9000+ ms display cap and has still not recovered when the 17 s screen recording ends — about 14 s of frozen UI and counting.

The hang starts after the map re-centering animation finishes and the menu is fully drawn, which rules out the tap hit test as the blocking step. The system status bar and Dynamic Island keep updating, so this is an app main-thread hang, not a device-level freeze.

### How to reproduce

1. Install the offline map for Ukraine / Kyiv —  Kyiv oblast, 1 sep 2026
2. Open the map at approximately the 100 m scale over the Kyiv-Pasazhyrskyi railway station.
3. Single-tap the railway station POI.

### Actual result

<table style="width:100%">
  <tr>
    <td><video src="https://github.com/user-attachments/assets/c6b68996-ebce-42b0-b323-80ec5735f558"  width="300" height="auto"/> </td>
  </tr>
</table>   

The context menu appears with the correct title and action row, then the UI freezes: the map does not pan or zoom and no control responds. Hang counter values read off the recording, measured from the tap:

* Logs – [logs.zip](https://github.com/user-attachments/files/32230758/logs.zip)

| Time after tap | Hang counter |
| --- | --- |
| 0.2 s | 257 ms |
| 1.2 s | 1221 ms |
| 5.1 s | 5119 ms |
| 9.0 s | 9000+ ms (display cap) |
| 14.2 s | 9000+ ms (recording ends, still hanging) |

### Expected result

The context menu opens and stays responsive. Any lookup that can exceed ~100 ms runs off the main thread, with the menu showing a placeholder until it resolves.

### Environment

- OsmAnd version: 5459
- iOS version: 27
- Device: iPhone 16 Pro Max 
- Map data: Kyiv oblast, 1 sep 2026

alex-dev-neo:
Fixes #5788 — the main thread hangs for over two minutes after opening the POI context menu for Київ-Пасажирський. Needs osmandapp/OsmAnd-core#1132.

## What happens

Tapping the station leads to `OAPOIViewController.setup:`, which resolves nearby transport stops and their routes synchronously. The watchdog report from the issue points at a block on the main queue, and a profile of the hung process shows where it goes:

```
_dispatch_call_block_and_release
 └ closure #1 in closure #1 in PlaceDetailsViewController.resolveDetailedObjectInBackground()
   └ -[OAPOIViewController setup:]
     └ -[OATransportStopsBaseController processTransportStop]
       └ -[OATransportStopsBaseController addRoutes:…]
         └ OsmAnd::ObfDataInterface::getTransportRoutes(…)
           └ ObfReaderUtilities::readSInt32 → CodedInputStream::ReadVarint32
```

530 of 532 main-thread samples are in `addRoutes`, and the leaves are the route geometry decode loop. Two things make it explode at this POI: a major station is referenced by 27 train relations, whose geometry is by far the largest in the data, and every installed obf returns its own copy of each stop — including live update files, which carry the region transport data as well. With 122 live files the station's routes were read 122 times over, geometry included, none of which the menu displays.

## Changes

* `OATransportStopsBaseController` gets a single `searchTransportStopsIn:` used by both searches, which merges the copies of a stop the way `TransportStopsRouteReader.mergeTransportStops` does on Android: the newest copy supplies the stop, later copies are compared by the route ids they carry, and only routes no earlier copy supplied are read. `deletedRoutesIds` keep an older copy from adding a route back but never remove what a newer one supplied, and the newest copy alone decides whether a stop is gone or a `#Missing Stop` placeholder — none of this was handled before.
* Copies are ordered by source file name before merging, as Android sorts obfs, so the copy that represents a stop does not depend on `QHash` traversal order.
* `addRoutes` uses the routes attached during the merge; it still reads them from the obf for stops that did not come from this search, such as a transport stop passed in from the map.
* Routes for the menu are read with `skipGeometry`. The rows need ref, name, type and colour, `initStopIndex` needs `forwardStops`; the geometry is only needed to draw the route, so `OATransportStopRoute.routeWithGeometry` reads it back and returns it. It does not touch the `shared_ptr` the menu holds, which `OATransportRouteController` reads on the main thread; `OATransportStopsLayer` keeps the returned route and publishes it inside `runWithRenderSync`.
* `findPOI` calls the new `performSearchInFile` with the `ObfFile` it already holds, instead of passing a path that core then looked up by scanning the whole collection.

Android gates live update files out of the transport search behind `USE_OSM_LIVE_FOR_PUBLIC_TRANSPORT`, off by default. iOS has no such preference and none is added here: live data stays in the transport search, which is the intended default for iOS.

## Measurements

iPhone 17 simulator, debug build, Kyiv oblast map plus 122 live update files, `[ContextMenu]` log:

| | before | after |
| --- | --- | --- |
| `Details update`, main thread | 142 156 ms | 551 ms |
| `Nearby POI scan`, background | 963 ms | 62 ms |

The same tap with only 2 live files installed takes 295 ms and 6 ms, so the amplification by the number of files is essentially gone.

## Checked on the simulator

* The station's route list matches OsmAnd for Android on the same data, with OSM Live for public transport enabled there: 25 routes, same set, including the ones only a live update file has (105, 106, 67, 719, 720, 722, 725, 733, 734) and without the ones its `deletedRoutesIds` removes.
* Subway entrance (Вокзальна, the 400 m `processTransportStopsForAmenity` branch): 57 routes, 281 ms. Android shows the same title and type there, and every route visible in its collapsed list is present; its list is truncated, so the full sets were not compared.
* Tapping a route opens it and draws the line on the map, i.e. the lazy geometry path works.

Not tested on a device.

The first commit keeps one copy per stop instead of merging them, which the third replaces; happy to squash if you prefer the branch without that step.


Sources/Transport/OATransportStopRoute.mm
```diff
@@ -192,6 +196,24 @@ - (OATransportStopRoute *) clone
...
+    const auto& point31 = OsmAnd::Utilities::convertLatLonTo31(OsmAnd::LatLon(_stop.latitude, _stop.longitude));
+    const auto bbox31 = (OsmAnd::AreaI)OsmAnd::Utilities::boundingBox31FromAreaInMeters(kRouteGeometrySearchRadiusMeters, point31);
+    const int zoomShift = 31 - OsmAnd::TransportStopsInAreaSearch::TRANSPORT_STOP_ZOOM;
+    const auto tbbox31 = OsmAnd::AreaI(bbox31.top() >> zoomShift, bbox31.left() >> zoomShift, bbox31.bottom() >> zoomShift, bbox31.right() >> zoomShift);
+
+    const auto& obfsCollection = [OsmAndApp instance].resourcesManager->obfsCollection;
+    const auto dataInterface = obfsCollection->obtainDataInterface(&tbbox31, OsmAnd::MinZoomLevel, OsmAnd::MaxZoomLevel, OsmAnd::ObfDataTypesMask().set(OsmAnd::ObfDataType::Transport));
+    const auto route = dataInterface->getTransportRouteWithGeometry(_route);
+    if (route)
+        _route = route;
```
tigrim: [P1] `_route` is replaced here on a background queue, while the route context menu can concurrently read the same `nonatomic` C++ `shared_ptr` property (for example, `forwardStops` in `OATransportRouteController`). Concurrent reads and writes to the same `std::shared_ptr` object are a data race and can corrupt its reference count or crash. Please synchronize access to the stored pointer, or load the geometry into a separate value and publish it on the same queue used by all readers.

Sources/Controllers/TargetMenu/OATransportStopsBaseController.mm
```diff
@@ -239,25 +247,127 @@ - (OATransportStop *) findNearestTransportStopForAmenity:(OAPOI *)amenity
...
+    for (const auto stopId : stopIds)
+    {
+        if (deletedStopIds.contains(stopId))
+            continue;
+
+        const auto& deletedIds = deletedRouteIds[stopId];
+        QList< std::shared_ptr<const OsmAnd::TransportRoute> > routes;
+        for (const auto& route : routesById[stopId])
+        {
+            if (!deletedIds.contains(route->id.id))
```
tigrim: [P2] This final filtering lets a deletion from an older live-update file remove a route restored by a newer file. The files are processed newest first, so a newer copy may already have supplied route A; if an older copy lists A in `deletedRoutesIds`, `deletedIds` still removes it here. The same ordering issue exists for `deletedStopIds`. Android preserves data already accepted from the newer file: later/older deletion records only prevent older data from being added. Please apply deletions according to file precedence instead of accumulating them globally and filtering the final result.

Sources/Transport/OATransportStopRoute.mm
alex-dev-neo: Confirmed and fixed in 074ef1a0e7. The property is `nonatomic, assign` and `OATransportRouteController` reads `_transportRoute.route->forwardStops` on the main thread while this ran on a background queue, so the write was a race I introduced — before this branch `_route` was only set at construction.

`loadGeometryIfNeeded` is now `- (std::shared_ptr<const OsmAnd::TransportRoute>) routeWithGeometry`, which reads the route and returns it without touching the stored pointer. `OATransportStopsLayer` keeps the result in its own `_stopRouteGeometry` and assigns it inside `runWithRenderSync`, so publishing and reading both happen on the queue that does the drawing; `hideRoute` resets it there as well. The trade-off is that the geometry is re-read each time a route is shown, which is one background read per user action.

Sources/Controllers/TargetMenu/OATransportStopsBaseController.mm
alex-dev-neo: You are right, and fixed in 074ef1a0e7. I had convinced myself the global filter was equivalent because the route list matched Android on my test data — but there all deletion records came from the newest file, which is exactly the case where the two rules coincide. The case you describe was not reachable in that data.

Deletions are now applied by precedence, as in `mergeTransportStops`: copies are walked newest first, a copy's `deletedRoutesIds` only keep older copies from adding a route back, routes from the newest copy are added unconditionally, and the final filtering is gone. `deletedStopIds` follows the same rule — only the newest copy decides whether a stop is gone, and an older deleted or `#Missing Stop` copy is skipped instead of dropping the stop.

One caveat on verification: with the live update cleanup in this branch, a daily file older than its month's monthly is deleted at startup, so I could not build a dataset on the simulator where an older file deletes a route a newer one supplies — within one month it cannot occur any more. The rule itself was checked line by line against `mergeTransportStops`, and the station's route list still matches Android on the data I do have.

tigrim (approved):

PR #5798, opened 2026-09-16
Default appearance original icon option missing

### LINKED ISSUE #5739: "Original" icon option is missing when adding a folder and cannot be saved in Default appearance
## Description
Two related issues with the **Original** icon option for **Favorites** folders:

1. The **Original** option is missing from the **Icon** section when adding a new folder.
2. Setting the **Icon** to **Original** through **Default appearance** is not saved and resets to the last used icon. On the first attempt the **Done** button stays disabled, so the change cannot be saved at all.

### How to reproduce

**First issue:**
* Go to **Menu** → **My Places** → **Favorites**
* Tap the menu in the navigation bar → **Add new folder**
* In the **Icon** section, tap the last used icon

**Second issue:**
* Go to **Menu** → **My Places** → **Favorites**
* Long tap on a folder → **Default appearance**
* Change the **Icon** to **Original**

### Actual result

<table style="width:100%">
  <tr>
    <th>First issue</th>
    <th>Second issue</th>
  </tr>
  <tr>
    <td></td>
    <td><video src="https://github.com/user-attachments/assets/4313fde6-44a4-4c9d-92af-39f162b95968" width="300" height="auto"></video></td>
  </tr>
</table>

**First issue:**
* The **Original** option is missing from the icon list

**Second issue:**
* Changing the **Icon** to **Original** cannot be saved, because the **Done** button remains disabled
* When both the color and the icon are changed, the **Original** icon state is not saved

### Expected result
* The **Original** option is available in the **Icon** section when adding a new folder
* Selecting **Original** in **Default appearance** enables the **Done** button and the choice is saved

**Environment**
OsmAnd Version: 5456
iOS version: 27 Beta

Vladyslav-lys:
## Related issue / task

Fixes #5739

## Summary

- Added `Original` to the icon selector when creating a Favorites folder and made it the initial selection.
- Fixed change detection and `Done` button availability when selecting `Original` in `Default appearance`.
- Preserved the initial `Original` selection when creating a folder without switching to another icon and back.
- Stopped filling an existing group’s empty icon from its points, preserving the group’s `Original` setting.
- Preserved individual point icons when clearing the folder’s icon override.
- Matched Android’s folder-creation behavior by keeping the recently used icons list unchanged.
- When selecting an existing folder in the favorite point editor, matched Android’s icon resolution order: explicit folder icon, original POI icon when available, saved point icon, then the editor’s fallback icon. Updated the icon selector to reflect the resolved icon.

## Testing

**Tested on:**

- Device / Simulator: iPhone 15
- iOS: 26.6.2

Additional automated controller-level checks were run during implementation on an iPhone 17 Pro Simulator with iOS 26.5.

### Scenarios

- Verified GPX export/import of an `Original` folder on iOS and persistence after restarting the application.
- Switched a favorite between folders with explicit icons and `Original`, saved it, and reopened the editor to compare the displayed and saved icons.
- Compared regular-favorite behavior with Android: selecting an `Original` folder can retain a previously saved folder icon instead of restoring an earlier manually selected icon.
- Checked the untouched initial `Original` selection with and without recently used icons.
- Checked explicit-icon initialization and switching between a concrete icon and `Original`, including the value passed for saving and subsequent editor initialization.
- Checked that creating a folder does not modify the recently used icons list.
- Checked reopening with no icon history, unrelated icon history, and an icon already present in `Last used`.
- Checked saved-point-icon fallback, original POI icon priority, and icon-selector synchronization when selecting an existing `Original` folder.

## Relevant checks

- [ ] CarPlay checked
- [x] Android parity checked

## AI disclaimer

**Implementation:**

- Tool / Agent: Codex
- Model: GPT-6 Astra

1. Fix the missing `Original` icon option when creating a Favorites folder.
2. Fix saving `Original` in `Default appearance`.
3. Review regression risks, compare the implementation with Android, and revise or revert changes as requested.
4. Remove `OAFavoriteGroupAppearanceTest` and `ParityChecks.m`.
5. Fix the initial `Original` selection being replaced by `Last used` when creating a folder without changing the selection.
6. Match Android’s icon-category behavior when reopening a newly created folder.
7. Fix icon handling when moving a favorite into an `Original` folder.
8. Provide testing scenarios, review the current branch, and verify the accuracy of the GitHub description.

- Used the existing empty-icon representation for `Original` and normalized nil/empty values when detecting appearance changes.
- Prevented a new folder’s empty name from causing its icon selector to use points from the default Favorites group.
- Removed automatic filling of an existing group’s empty icon from individual points.
- Removed obsolete default-icon fallback methods from the folder editor.
- Used temporary simulator checks without adding permanent test sources to the final branch changes.

**Final review:**

- Tool / Agent: Codex
- Model: GPT-6 Astra
- [x] Final diff reviewed

**Significant findings:**

- Addressed the additional default-selection case reported during manual testing: `Original` must persist without switching away from it and back.
- Investigated the category changing to `Last used` after folder creation and removed the creation-time update to recent icons. Icons already in recent history retain Android’s category-selection behavior.
- Fixed the displayed and selected icon when moving a favorite into an Original folder, following Android’s behavior.
- Reviewed additional risks involving GPX persistence, repeated folder switching, unavailable POIs, uncommon imported icons, and synchronous POI lookup. These reviews do not constitute complete Android parity or performance testing.


tigrim:
There is one issue with applying **Original** to existing points.

In `OAFavoritesHelper.updateGroup`, an empty `iconName` now skips point updates even when `updatePoints == YES`:

```objc
if (updatePoints && iconName.length > 0)
```

As a result:
- **Apply to existing** becomes a no-op for the icon because the group value is not updated either.
- **Apply to all** clears only the group override, while existing points keep their previous concrete icon instead of switching to Original.

This also differs from Android: when the selected icon is empty, Android updates every affected point to its original POI icon, or clears the icon ID when an original POI icon is unavailable.

Please handle the empty `iconName` branch explicitly for `updatePoints == YES`, restoring/clearing each point icon as appropriate.

Sources/Controllers/TargetMenu/PointEditing/OABaseEditorViewController.mm
```diff
@@ -437,13 +438,9 @@ - (void) setupIconHandler
...
     
-    if (!_isNewItem && self.editIconName)
-        _selectedIconName = self.editIconName;
-    else
-    {
-        _selectedIconName = [self getDefaultIconName];
-        self.editIconName = _selectedIconName;
-    }
+    // An empty icon means Original, including the initial selection for a new folder.
+    _selectedIconName = self.editIconName ?: @"";
```
tigrim, line 442: This changes the untouched default for every newly created folder to `Original`. Issue #5739 asks to make `Original` available, but the previous behavior selected the last-used/default icon, and Android still initializes a new group with `DEFAULT_ICON_NAME`. Could we keep the existing initial selection and only add `Original` as an available option, unless changing the default was explicitly confirmed as a product requirement?

Sources/Helpers/OAFavoritesHelper.mm
```diff
@@ -772,8 +774,26 @@ + (void)updateGroup:(OAFavoriteGroup *)group
...
+        {
+            NSString *pointIconName = iconName;
+            if (iconName.length == 0)
+            {
+                // Original restores the POI icon or clears the point's explicit icon.
+                pointIconName = nil;
+                NSString *originName = [point getAmenityOriginName];
+                if (originName.length > 0)
+                {
+                    OAPOI *poi = [OAAmenitySearcher findPOIByOriginName:originName
```
tigrim, line 813: This performs a synchronous OBF amenity search once per favorite while handling the save action on the main thread. Applying `Original` to a large folder can therefore block the UI for N sequential searches. Android currently has equivalent resolution semantics, but copying it here also copies this scaling problem. Please move the bulk resolution off the main thread or avoid repeated searches by using a cached/persisted original icon.

Sources/Controllers/TargetMenu/PointEditing/OABaseEditorViewController.mm
Vladyslav-lys: confirmed by product manager

tigrim:
Creating a new folder with `Original` from the favorite point editor does not follow the same icon resolution path as selecting an existing `Original` folder.

The new folder returns an empty `iconName`, which `addGroupWithName:` assigns directly to `_selectedIconName`. As a result, when adding a favorite from a POI, its original POI icon can be replaced with an empty/default icon.

Could we resolve the icon through `iconNameForGroup:` after creating the group, or reuse `onGroupChanged:` so newly created and existing `Original` folders behave consistently?

tigrim:
The latest fix resolves the original case for a newly created folder with `Original`, but it introduces a regression when the entered folder name resolves to an existing group.

`resolvedIconName` is now unconditionally replaced with the result of `iconNameForGroup:`. The group editor allows reusing an existing group name with a different appearance, so if the user explicitly selects a concrete icon, that selection is now ignored: the point receives the existing group icon (or the resolved original POI icon) instead.

Previously, the explicitly selected `iconName` was preserved in this flow. Could we call `iconNameForGroup:` only when `iconName` is empty (`Original`) and keep non-empty icon names unchanged?

tigrim:
The latest change prevents the existing-group icon regression by disabling that flow, but it introduces a behavior regression of its own.

`OAFavoriteGroupEditorViewController` now sets `validatesGroupUniqueness = YES` for every new folder. Previously this was caller-controlled: My Places explicitly enabled strict uniqueness, while the favorite point editor left it disabled so an existing parent-only group (and an existing group with changed appearance) could be reused. The new assignment makes `allowsExistingGroupFor:` and `allowsValidationForGroupName` ineffective in all new-folder flows, so Done is now disabled for cases that were previously supported.

Could we keep the caller-specific uniqueness behavior and instead fix icon resolution directly: preserve a non-empty `iconName`, and call `iconNameForGroup:` only when the selected icon is empty (`Original`)?

tigrim (approved):

Sources/Controllers/TargetMenu/PointEditing/OABaseEditorViewController.mm
```diff
@@ -678,20 +656,15 @@ - (void)reloadCollectionData
...
 {
-    if (_isNewItem)
-        return;
-    
     NSIndexPath *indexPath = [self.tableView indexPathForCell:cell];
     if (indexPath)
         [self.tableView reloadRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationNone];
-    if ([category isEqualToString: @"original"])
+
+    if ([category isEqualToString:_poiIconCollectionHandler.ORIGINAL_KEY])
```
alex-dev-neo, line 663: Selecting `Original` sets `editIconName` to `""`. Switching to another category afterwards does not change it back: `selectCategory` only swaps the icons shown in the palette.

Possible case (derived from the code, not checked on a simulator): a folder has the `special_star` icon. Open `Default appearance`, tap `Original`, then switch to the category that contains the star without tapping an icon. The star probably stays highlighted, `Done` is enabled, and saving applies `Original`.

Before this PR, `Original` just restored the group's own icon, so this mismatch could not happen. Android handles it in `FavoriteEditorIconController.setSelectedCategory`: for a non-`Original` category it selects `getSelectedIconKey()` or falls back to `requireIcon()`. Could we do the same here, i.e. restore a concrete icon when a non-`Original` category is selected?

Sources/Controllers/TargetMenu/PointEditing/OABaseEditorViewController.mm
```diff
@@ -599,7 +577,7 @@ - (BOOL)isAppearanceChanged
     OAFavoriteGroup *groupExist = [self existingGroupFor:self.editName];
     return !groupExist
             || ![self.editBackgroundIconName isEqualToString:groupExist.backgroundType]
-            || ![self.editIconName isEqual:groupExist.iconName]
+            || ![self.editIconName isEqualToString:groupExist.iconName ?: @""]
```
alex-dev-neo, line 580: `isAppearanceChanged` now treats `nil` and `""` as the same icon, but the check in `textViewDidChange:` (line 700) does not: `![groupExist.iconName isEqualToString:self.editIconName]`. After a folder is loaded from GPX, an `Original` folder has `iconName == nil`. Any edit of the name field then counts as an appearance change, and `Done` is enabled even when nothing has changed. Could we use the same `?: @""` there?

Sources/Controllers/TargetMenu/PointEditing/OAFavoriteGroupEditorViewController.mm
```diff
@@ -26,6 +28,7 @@ - (void)postInit
     
     if (self.isNewItem)
     {
+        self.validatesGroupUniqueness = YES;
```
alex-dev-neo, line 31: Blocking existing names for every new folder matches Android: `GroupEditorFragment.checkEnteredName` rejects any existing name when `pointsGroup == null`, and `groupExists` ignores case. The change is only half done, though. `postInit` now always sets `validatesGroupUniqueness = YES`, and no caller sets it back to `NO`. As a result:

- `isParentOnlyGroup:` and the second branch of `allowsExistingGroupFor:` can no longer be reached;
- `viewController.validatesGroupUniqueness = true` in `FavoriteListViewController+Actions.swift:428` does nothing.

Please either remove the property together with the code that can no longer run, or keep it caller-controlled. Having both makes the next change error-prone.

Sources/Controllers/TargetMenu/OAEditPointViewController.mm
```diff
@@ -1299,21 +1300,24 @@ - (void)addGroupWithName:(NSString *)name
...
         }
         else
         {
             [OAFavoritesHelper addFavoriteGroup:editedGroupName
                                           color:color
                                        iconName:iconName
                              backgroundIconName:backgroundIconName];
+            group = [OAFavoritesHelper groupByTrimmedName:editedGroupName];
         }
+        resolvedIconName = [self iconNameForGroup:group];
```
alex-dev-neo, line 1320: `iconNameForGroup:` is called no matter which icon the user picked. This is safe today only because existing names are now blocked (see the comment on `OAFavoriteGroupEditorViewController.mm:31`). If that check is ever relaxed, an explicitly selected icon will again be replaced by the existing group's icon. `resolvedIconName = iconName.length > 0 ? iconName : [self iconNameForGroup:group];` would make this code safe on its own.

Sources/Controllers/TargetMenu/PointEditing/OAFavoriteGroupEditorViewController.mm
```diff
@@ -181,13 +214,37 @@ - (void)addPointsGroup
...
+            [self finishEditingPointsGroup:updatePoints updateGroupValues:updateGroupValues];
+        }];
+    }
+    else
+    {
+        [self finishEditingPointsGroup:updatePoints updateGroupValues:updateGroupValues];
+    }
+}
+
+- (void)beginSaving
```
alex-dev-neo, line 238: The `_isSaving` flag is fragile. The "saving in progress" state now lives in three places: `_isSaving`, `modalInPresentation` and `view.userInteractionEnabled`. It is also checked by hand at every entry point: `onRightNavbarButtonPressed`, `onLeftNavbarButtonPressed`, `onGestureRecognizerShouldBegin`, the `dismissViewController` override and `editPointsGroup`. This will break on future changes:

- Any new entry point (a button, a gesture, a delegate call) has to remember to check the flag.
- `dismissViewController` does nothing without any signal while the flag is set, which is surprising for callers.
- `finishEditingPointsGroup` must reset the flag before calling `dismissViewController`. If someone reorders these lines, the editor stays open and locked.
- If a future path in `updateGroup` returns without calling `completion`, the editor is locked for good: both navbar buttons and the gesture are blocked.

Could we let UIKit do the blocking instead? Show an `MBProgressHUD` on `self.navigationController.view` (it is already used, e.g. in `OAWeatherCacheSettingsViewController`) and set `modalInPresentation = YES` while the lookup runs. The completion then hides the HUD and dismisses the editor. The HUD covers the whole container, navbar included, so the flag, the four guards and the `dismissViewController` override can all be removed. The user also sees that something is happening; right now the screen just stops responding. Please check that no gesture recognizer on a parent view can still fire under the HUD in the presentation paths used here.

Sources/Controllers/TargetMenu/PointEditing/OABaseEditorViewController.mm
Vladyslav-lys: I checked the full call chain. When switching to a non-Original category containing icons, BasePoiIconCollectionHandler.onMenuItemSelected calls selectCategory and then selectIconName for the first icon. This triggers onCollectionItemSelected and updates editIconName, so the saved value should match the highlighted icon.

This differs from Android, which restores the previously selected icon, but the mismatch described in the comment does not appear to occur through this path. I’m keeping the existing iOS behavior of selecting the first icon in the category.

Sources/Controllers/TargetMenu/PointEditing/OAFavoriteGroupEditorViewController.mm
```diff
@@ -181,13 +164,48 @@ - (void)addPointsGroup
 
 - (void)editPointsGroup:(BOOL)updatePoints updateGroupValues:(BOOL)updateGroupValues
 {
-    if (![self.editIconName isEqual:_favoriteGroup.iconName])
+    [self.view endEditing:YES];
+    self.modalInPresentation = YES;
+
+    // Disable the container so its navigation gestures cannot receive touches through the HUD.
+    UIView *containerView = self.navigationController.view ?: self.view;
+    containerView.userInteractionEnabled = NO;
```
alex-dev-neo: **The lock must be limited to the editor itself.**

`self.navigationController` here is the app's root navigation controller. My Places and this editor are both pushed onto it (`MyPlacesNavigator.swift:82`, `FavoriteListViewController+Actions.swift:18`). Disabling `userInteractionEnabled` on its view freezes the whole app for as long as the lookup runs, not just this screen:

- If `completion` is ever missed, the whole app stays frozen, not just the editor.
- If something else changes navigation while the lookup runs, for example `MyPlacesNavigator` calling `popToRootViewController` after an import, the editor leaves the stack. The HUD and the disabled root view then stay on top of the map until `completion`.
- The code changes state it does not own, and the restore sets `YES` unconditionally instead of the previous value.

Please lock only what belongs to the editor: its own view, its own `navigationItem` and its own gesture-delegate hook. Do not change the navigation controller's view, its gestures or other screens.

Also, `self.modalInPresentation = YES/NO` (lines 168 and 200) has no effect here. This editor is only ever pushed, never presented, so these two lines should be removed.

PR #5800, opened 2026-09-16
Fix split-label thread safety in OAGPXLayer

alex-dev-neo:
`processSplitLabels:doc:` enqueues its work on `_splitLabelsQueue`, which has no concurrency limit, so those operations run in parallel with each other and with the main thread. They guarded themselves by comparing `_splitCounter` — but they read that strong ivar directly, while `resetSplitCounter` replaced the object on the main thread under `_splitLock`. That is an ARC load of a pointer that may be in the middle of being released.

The guard was also a check-then-act: `splitCounter == _splitCounter` was evaluated separately from the `appendSplitLabels:` and `decrementSplitCounter` that followed, so a reset landing in between let a stale operation append into a label list the main thread had just cleared.

Changes:

- Add `_splitGeneration`, bumped by `resetSplitCounter` under `_splitLock`. Operations capture the generation as a plain integer and no longer read layer state off-thread.
- `appendSplitLabels:generation:` and `decrementSplitCounterForGeneration:` verify the generation and mutate under one lock; the decrement returns -1 for a stale generation so it cannot touch a fresh counter.
- The per-point loop keeps only `weakOperation.isCancelled`, which is lock-free and already covers a restart — `refreshStartFinishPoints` always cancels before resetting. No lock in the hot path.
- `_startFinishProvider` was torn down outside `_splitLock` in `resetLayer` and `refreshStartFinishPoints` but rebuilt inside it in `refreshStartFinishProvider`; all three now go through `removeStartFinishProvider` under the lock.

Verification: built and run on the iOS Simulator (master build, ANGLE renderer; the file is byte-identical on both branches). Three tracks of 40–55 km with a time split were shown at once, so each finished track load re-entered `refreshStartFinishPoints` — cancel plus generation reset — on top of the previous track's in-flight split operations, and visibility was toggled repeatedly on top of that. Split labels and start/finish markers rendered correctly and came back after every cycle; no crashes over four launches.

This does not demonstrate that the data race itself is gone — that needs Thread Sanitizer, and the argument here is that no unsynchronized read of the strong ivar remains.

Context: one of the crash signatures in the September iOS reports was `objc_msgSend$visualization3dByType` on a dead receiver from `-[OAGPXLayer refreshGpxTracks:reset:]`. This PR does not claim to fix it — the layer's `_gpxFiles` and `_cachedTracks` are reached only from the main thread, so that crash has another cause. This is the thread-safety defect that does exist in the class.


Sources/Controllers/Map/Layers/OAGPXLayer.mm
```diff
@@ -192,7 +193,7 @@ - (void) resetLayer
     [super resetLayer];
 
     [self.mapView removeTiledSymbolsProvider:_waypointsMapProvider];
-    [self.mapView removeTiledSymbolsProvider:_startFinishProvider];
+    [self removeStartFinishProvider];
```
tigrim, line 197: `resetLayer` removes the provider but leaves the current split generation and queued/running operations valid. An operation can therefore finish after `hideRecGpxTrack`, append the old labels, and dispatch `refreshStartFinishProvider`, re-adding the provider that this method just removed. Please cancel/invalidate the split work here, and make the deferred provider refresh verify its captured generation before rebuilding.

alex-dev-neo: Confirmed, and fixed in 0fcd28e.

`resetLayer` now calls `cancelSplitLabels`, which cancels the queue, bumps the generation through `resetSplitCounter` and clears the labels. The deferred refresh rechecks its captured generation before rebuilding, because the reset can land between the `dispatch_async` and the block running, and the cancellation alone would not cover that window.

Worth noting this is not something the PR introduced: on r5.4 the old guard compares the captured `OAAtomicInteger *` against `_splitCounter`, and `resetLayer` does not touch that either, so the same operation would have rebuilt the provider. It is a gap in the fix rather than a regression, which is why it belongs here.

### Verified on a simulator

One visible track with a split interval set, breakpoint callbacks on `cancelSplitLabels`, `appendSplitLabels:generation:` and `refreshStartFinishProvider`.

Normal path first: enabling a split builds the labels and they appear on the map, so retiring the work in `resetLayer` does not lose them — every path that resets and still expects labels continues into `refreshStartFinishPoints`, which re-enqueues.

Then the race itself. An operation was suspended at the entry of `appendSplitLabels:generation:`, before it takes `_splitLock`, leaving it in flight on the operation queue while the main thread stayed free:

```
 +0.12 s  HELD      q=NSOperationQueue 0x12f4b2e00
+42.14 s  CANCEL    gen=28      <- hiding the track, via resetLayer
+42.21 s  CANCEL    gen=29
+42.22 s  REBUILD               <- the legitimate rebuild after the reset
+42.22 s  RESUMED               <- the stale operation released
          (nothing further)
```

No rebuild followed the stale operation, and the map stayed clean: no line, no start/finish icons, no labels. Before the change that operation would have appended and dispatched a refresh with its generation still current.

One observation from the same session, unrelated to this PR: suspending inside `appendSplitLabels:` rather than at its entry stalls the main thread for as long as it is held, which the app's own log reports as `runWithRenderSync block took 4316.9 ms`. `_splitLock` is held across the append and across the whole provider construction in `refreshStartFinishProvider`, so background split work and the main thread contend on it. Not something I want to change in this PR, but it may be worth a look separately.

tigrim (approved):

PR #5802, opened 2026-09-16
Cut the main-thread cost of refreshing GPX tracks

alex-dev-neo:
`OASelectedGPXHelper` queues one `GpxLoadOperation` per visible track, and every completion calls `updateGpxTracksOnMapObservable.notifyEvent`. `onUpdateGpxTracks` had no coalescing, so each event queued its own `refreshGpxTracks` on the main queue — and `initRendererWithGpxTracks` always passes `reset:YES`, which emptied `_cachedTracks`, `_cachedColors` and `_cachedWallColors` and recolorized every track from scratch. On top of that, each pass walked all visible tracks through `getGPXItem`, which runs `isAnalyseNeeded` and stats the file on the calling thread.

Three changes, all on the same path:

**1. Coalesce the refresh.** `onUpdateGpxTracks` keeps at most one pending refresh (`std::atomic_bool`, because `notifyEvent` also arrives from background threads). The flag is cleared before the rebuild, so an update arriving during it still schedules the next one instead of being dropped.

**2. Keep the caches that a reset does not have to drop.** `refreshGpxTracks:reset:` still calls `resetLayer` — `drawLine:` always appends to `_linesCollection` and only the recording track is updated in place, so the collection has to be rebuilt. But `dropCachedTracksChangedIn:` now drops only tracks that disappeared, whose `OASGpxFile` was replaced by a reload, or the current track, which is re-cached every pass anyway.

**3. Stop stating files where only the database entry is needed.** `hideRemovedGpx` wants to know whether an entry exists; `refreshStartFinishPoints` reads appearance parameters. `GpxDbHelper.getItem` returns `dataItems[file]` either way — `readIfNeeded` only decides whether a background re-read is scheduled — so `getCachedGPXItem:` returns the same object without the stat. `addTrackToCached` keeps the reading variant, since that is what schedules analysis for a track entering the layer.

### Measurements

lldb breakpoints on a running app, five visible tracks. These are invocation counts, not durations, so they do not depend on the device or frame rate.

Refreshes and lookups, same breakpoint set before and after change 1:

| | before | after |
|---|---|---|
| `-[OAMapViewController refreshGpxTracks]` | 4 | 1 |
| `-[OAGPXLayer refreshGpxTracks:reset:]` | 4 | 1 |
| `-[OAAppSettings hideRemovedGpx]` | 5 | 2 |
| `-[OAGPXDatabase getGPXItem:]` | 52 | 16 |
| `-[OASelectedGPXHelper shouldLoadFileAtPath:]` | 20 | 8 |

Two caveats on that table. Coalescing bounds the number of *pending* refreshes, not the total: completions that arrive while a refresh is pending collapse into one, completions spread out in time still get their own pass. And the absolute counts shift with measurement overhead — under a heavier breakpoint set the same build settles at 2 refreshes rather than 1, reproducibly. So read the table as "the multiplier is much smaller", not as "exactly one refresh".

Change 2, measured on one binary by letting the first refresh settle and then triggering a second over an unchanged track set from lldb: the cold refresh ran colorization twice (only two of the five test tracks use a gradient coloring type) and cached five tracks; the repeat refresh ran colorization zero times and cached zero. The "before" figure is not measured here — the previous code emptied the caches unconditionally and colorization is gated on `_cachedColors[qKey].isEmpty()`, so it necessarily recomputed.

Change 3, three consecutive runs, identical results each time:

```
getGPXItem (stats the file)      8
getCachedGPXItem (no file access) 24
```

Every `getCachedGPXItem` call site was a `getGPXItem` call site before, and control flow is unchanged, so this is 32 disk-touching lookups down to 8.

No visual regression after any of the three: all five tracks render on the first and on a repeat refresh, gradient colorization intact, start/finish markers present, no duplicated lines.

### Scope

This addresses the largest cluster in the September crash reports — six of twelve are `0x8BADF00D` watchdog terminations where the main thread was busy when the process was asked to exit, three of them inside exactly this refresh. It does not claim to eliminate that cluster. Still on the main thread: `fileExistsAtPath` in `hideRemovedGpx`, which is what that function exists to check, and `shouldLoadFileAtPath`, already minimised by 2cd15301e2. Removing those means making `buildGpxList` asynchronous, which changes its `BOOL` contract across four call sites; after coalescing the remaining cost is on the order of N stats once per burst, which did not look worth that risk. Unrelated to this PR, `releaseAllResources` and `OsmAnd::ReleaseCore()` still wait without a timeout during teardown.


Sources/Controllers/Map/Layers/OAGPXLayer.mm
```diff
@@ -267,16 +267,31 @@ - (void) refreshGpxTracks:(NSDictionary<NSString *, OASGpxFile *> *)gpxFiles res
     if (reset)
     {
         [self resetLayer];
-        [_cachedTracks removeAllObjects];
-        _cachedColors.clear();
-        _cachedWallColors.clear();
+        [self dropCachedTracksChangedIn:gpxFiles];
```
tigrim: [P1] This preserves stale colorization data when the palette changes. onPaletteChangedEvent: reaches this method with reset:YES, but an edited or replaced palette keeps the same name and the same OASGpxFile instance. The checks in refreshGpxTracks then see a non-empty cache and an unchanged palette name, so the track continues using the old palette. _cachedWallColors has the same issue when the palette changes but the wall coloring type does not. Please fully invalidate these caches for palette-change refreshes, or include the palette revision or content in the cache key.

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -184,6 +185,7 @@ @implementation OAMapViewController
     
     dispatch_queue_t _mapSourceUpdateQueue;
     BOOL _mapSourceInvalidated;
+    std::atomic_bool _gpxTracksRefreshScheduled;
```
tigrim, line 188: [P2] Please initialize this atomic explicitly. This target uses C++17, where default construction of std::atomic_bool does not initialize its value, so the first exchange(true) formally reads an indeterminate state and may discard the first refresh. Initializing it to false in init, as OATrackPreviewMapRenderer does for _cancelled, avoids relying on Objective-C allocation details.

Sources/GPX/OAGPXDatabase.mm
```diff
@@ -203,12 +203,21 @@ + (EOAGPX3DLineVisualizationPositionType)lineVisualizationPositionTypeForName:(N
 }
 
 - (OASGpxDataItem *_Nullable)getGPXItem:(NSString *)filePath
+{
+    return [[OASGpxDbHelper shared] getItemFile:[self gpxFileForPath:filePath]];
+}
+
+- (OASGpxDataItem *_Nullable)getCachedGPXItem:(NSString *)filePath
+{
+    return [[OASGpxDbHelper shared] getItemFile:[self gpxFileForPath:filePath] readIfNeeded:NO];
```
tigrim, line 212: [P2] This is not actually a no-file-access lookup. GpxDbHelper.getItem(file, readIfNeeded: false) still executes file.exists() whenever the cached item is non-null; readIfNeeded only skips the analysis and re-read branch. In hideRemovedGpx that stat is then followed by another fileExistsAtPath: check. Please either expose a true cache-only getter from the shared helper or update the API comment and performance claims so callers do not assume this is safe from main-thread filesystem I/O.

Sources/Controllers/Map/Layers/OAGPXLayer.mm
alex-dev-neo: Confirmed, this is a real regression and a good catch.

`onPaletteChangedEvent:` passes `[_gpxFiles copy]`, so the `OASGpxFile` instances are identical and `dropCachedTracksChangedIn:` keeps every entry. The checks in `refreshGpxTracks` then compare `prev_coloring_type`, `prev_color_palette` (the name), `colorization_scheme` and the emptiness of `_cachedColors` — an edited or replaced palette changes none of the four, so the colors are never recomputed. `_cachedWallColors` has the same hole through its `prev_wall_coloring_type` check.

Rather than widen the cache key, I invalidate where the staleness is known: `onPaletteChangedEvent:` now calls `dropCachedColors`, which clears `_cachedColors` and `_cachedWallColors` before the refresh. That makes both the gradient branch and the wall branch recompute, since each of them also tests the corresponding cache for emptiness.

I checked the other inputs colorization depends on and did not find a second case: the per-track palette name and both coloring types are covered by the existing checks; `applicationMode`'s `getMaxSpeed` never reaches the layer, because `onAppModeChanged` does not refresh tracks — speed gradients were already stale across a profile switch before this PR; subscription state has no path here either.

Verified on a simulator with breakpoint counters on `refreshGpxTracks:reset:`, `onPaletteChangedEvent:`, `dropCachedColors` and `-[OARouteColorize initWithGpxFile:...]`, using one visible track switched to Altitude coloring. A plain `reset:YES` refresh with unchanged documents (terrain toggle → `initRendererWithGpxTracks`) gives 1 refresh and **0** colorizations, which is the cache-preserving behaviour this PR introduces; a palette change event in the altitude category gives `onPaletteChangedEvent:` → `dropCachedColors` → refresh → **1** colorization. Same process, same track, same code path, the only difference being the cache drop. The event I drove was `Added` (a palette duplicate) rather than an in-place content edit, since `isRoutePaletteChangeEvent:` treats Added, Updated, Replaced and Removed identically.

Sources/Controllers/Map/OAMapViewController.mm
alex-dev-neo: Done — initialized in `commonInit`, next to the other ivar setup.

For the record, the ivar is not indeterminate in practice today: the Objective-C runtime zero-fills instance memory on allocation, and `std::atomic_bool` has a trivial default constructor before C++20, so no `.cxx_construct` is emitted for it and the zeroed value survives. But that is exactly the allocation detail you are pointing at, the explicit initialization costs nothing, and the precedent you cite is right there in the tree.

Sources/GPX/OAGPXDatabase.mm
alex-dev-neo: The filesystem part does not hold, as far as I can tell from the shared sources.

In `GpxDbHelper.getItem(file:callback:readIfNeeded:)` the guard is

```kotlin
if (readIfNeeded && GpxDbUtils.isAnalyseNeeded(item) && GpxDataItem.isRegularTrack(file))
```

`readIfNeeded` is the first operand, so with `false` Kotlin short-circuits before `isAnalyseNeeded` — which is where `file.lastModified()` lives — and before `isRegularTrack` (itself a pure string prefix check). `exists()` occurs exactly twice in the whole of `GpxDbHelper`, in `loadGpxItems` and `loadGpxDirItems`, i.e. the blocking bulk load at startup, and neither is reachable from `getItem`. `isPathEmpty()` is pure string work as well. So the lookup itself performs no file access, on a cache hit or a miss.

Your conclusion about the comment stands, though. It promised something about the call site that the method cannot guarantee, and `hideRemovedGpx` does follow it with a `fileExistsAtPath:` on the main thread, so the caller is still not free of main-thread I/O. I reworded it to state only what is actually skipped: the freshness check that stats the file, and the re-read it would schedule. Moving that remaining `fileExistsAtPath:` off the main thread is a separate change I have deliberately kept out of this PR.

If you were reading a different revision of the shared module than the one this branch builds against, tell me which and I will recheck.

Sources/Controllers/Map/Layers/OAGPXLayer.mm
```diff
@@ -267,16 +276,31 @@ - (void) refreshGpxTracks:(NSDictionary<NSString *, OASGpxFile *> *)gpxFiles res
...
```
tigrim: [P1] Palette import and restore still leaves these caches stale. OAFileSettingsItem.installItem invalidates the palette repository and refreshImportedPalette only evicts the repository cache; neither emits a PaletteChangeEvent. BackupUtils.updateCacheForItems then sends only updateGpxTracksOnMapObservable. That path reaches reset:YES here with the same OASGpxFile instances, so the imported palette is loaded by the repository but _cachedColors and _cachedWallColors survive and visible tracks keep the pre-import colors. Before this change reset:YES cleared them. Please make the backup/import refresh invalidate the color caches as well, or propagate an invalidation reason that distinguishes this path from an unchanged-document refresh.

tigrim: [P1] Selecting a different palette for an individual track is another stale-cache path that is not covered by dropCachedColors. The Track Appearance screen mutates gpx.gradientPaletteName in place and sends updateGpxTracksOnMapObservable, so this method keeps the cache because the OASGpxFile instance is unchanged. refreshGpxTracks notices the new palette name, but when the wall coloring type matches the line gradient type it reuses the non-empty _cachedWallColors as the new line cache; it then records the new palette name, while the wall branch skips recomputation because the wall type is unchanged. If the types differ, the line is recalculated but the wall still keeps the old palette. Please invalidate both color caches when prev_color_palette differs before either cache can be reused, or track the wall palette independently.

alex-dev-neo: Confirmed, and this one is on me twice over: your first comment already asked for full invalidation on palette-change refreshes, and I implemented the narrowest reading of it — an invalidation in `onPaletteChangedEvent:`, which only covers the paths that emit an event.

This path emits none. `PaletteRepository.invalidatePalette` is just `cachedPalettesForId.remove(id)` with no `notifyListeners`, `refreshImportedPalette` calls only that, and `updateCacheForItems` sends `updateGpxTracksOnMapObservable` alone. The refresh then arrives here with the same `OASGpxFile` instances and an unchanged palette name, so nothing in the chain can tell the colors expired.

Rather than invalidate at one more entry point, I removed the cache reuse: `reset:YES` clears `_cachedTracks`, `_cachedColors` and `_cachedWallColors` again, exactly as before this PR. Keying the cache on a palette revision is the real fix, but that is a change in the shared module and does not belong in a release branch as an afterthought.

alex-dev-neo: Confirmed. And you had already named this case in the first review — the wall cache going stale when the palette changes but the wall coloring type does not — so it should not have taken a second round.

The sequence is as you describe: the gradient branch is entered because the palette name differs, it writes the new name into `prev_color_palette` straight away, and then `shouldCalculateColorCache` can come out false when the wall type matches the line type and `_cachedWallColors` is non-empty, at which point `_cachedColors[qKey] = _cachedWallColors[qKey]` stores colors computed with the previous palette under the new name. The wall branch below skips its own recomputation for the same reason.

Fixed by removing the cache reuse altogether: `reset:YES` clears all three caches as it did before this PR, so neither cache can outlive a palette change.

What remains in the PR is the coalescing of `onUpdateGpxTracks` and the cache-only database lookup, which is where the measured main-thread saving came from: a burst of N track loads goes from N full rebuilds to one, `getGPXItem` calls from 52 to 16, `shouldLoadFileAtPath` from 20 to 8, and of the remaining database lookups only the ones that genuinely need a re-read touch the filesystem.

Sources/Controllers/Map/Layers/OAGPXLayer.mm
alex-dev-neo: Correction to this thread: the claim above that I had found no second case was wrong, and so was the fix it described. You had pointed at the palette itself being mutable under a stable name; I only covered the one path that emits an event, and both cases you raised afterwards — import and restore, and picking a palette for a track — sit outside it.

The reuse is gone as of c54788f, so `reset:YES` clears the colorization caches exactly as it did before this PR, and `dropCachedColors` along with it.

tigrim (approved):

PR #5803, opened 2026-09-17
Give the shutdown waits a deadline

alex-dev-neo:
Depends on osmandapp/OsmAnd-core#1134 — the options used here do not exist without it, so this does not build on its own.

`applicationWillTerminate:` has about five seconds, and the renderer and core teardown it runs joined their worker threads without a deadline. A worker inside a file read or a driver call outlives that budget, the process is killed, and the user gets a crash report. Six of the twelve September reports are `0x8BADF00D` watchdog terminations, and one of them shows the main thread inside exactly this chain: `applicationWillTerminate:` → `onApplicationDestroyed` → `releaseContext:` → `postReleaseRendering` → `QWaitCondition::wait`.

The core change makes those waits bounded but leaves the limits off by default. This sets them:

- `MapRendererSetupOptions::maxTeardownWaitTime` = 500 ms, which covers three waits — the resources worker join, the request pool drain, and the GPU worker join.
- `OsmAnd::ReleaseCore(1000)` for the Qt application thread.

Worst case is then 2000 ms (the wait for the GPU sync stage, which was already bounded and is untouched here) + 3 × 500 + 1000 = 4.5 s, leaving the rest of shutdown inside the budget. In a healthy app these waits finish in milliseconds and the limits never apply. Both numbers are named constants, so tightening them is a one-line change; the pre-existing two seconds dominates the sum if more headroom is wanted.

Safe on iOS specifically because `releaseContext:` has exactly one caller, `onApplicationDestroyed`, so it only ever runs when the app is terminating and nothing continues afterwards.

### Evidence

Reproduced on the Simulator by suspending the resources worker from lldb and terminating the app from the app switcher, which is what actually delivers `applicationWillTerminate:` — swiping a suspended app away does not.

| | without | with |
|---|---|---|
| `Releasing context` logged | yes | yes |
| `Releasing render and frame buffers` logged | no | yes |
| breakpoint on `releaseRenderAndFrameBuffers` | 0 hits | hit |
| process after the swipe | alive after 2 minutes, killed by hand | exited on its own |

Without the change, `sample` showed 317 of 317 samples parked in `MapRendererResourcesManager::stopWorkerThread` at `REPEAT_UNTIL(_workerThread->wait())`. A healthy termination, with nothing suspended, behaves as before and logs `applicationWillTerminate` followed by the full teardown.

Durations are not measured — the teardown log lines go through `printf` without timestamps, so this shows that teardown completes, not how long it takes. The budget arithmetic above is from the code, not from a stopwatch.


Sources/Views/OAMapRendererView.mm
```diff
@@ -808,6 +811,10 @@ - (void)createContext
     rendererSetup.pathToOpenGLShadersCache = QString::fromNSString(NSTemporaryDirectory());
     rendererSetup.gpuWorkerThreadEnabled = true;
     rendererSetup.displayDensityFactor = _displayDensityFactor;
+    // releaseContext: runs from applicationWillTerminate:, where the whole shutdown has about five
+    // seconds. Three teardown waits are bounded by this value, on top of the two seconds
+    // postReleaseRendering already spends waiting for the GPU sync stage.
+    rendererSetup.maxTeardownWaitTime = kTeardownWaitTime;
```
tigrim, line 817: This timeout makes the dependent core change abandon live workers, but their lifetime is not detached from the renderer. The GPU thread is started with `std::bind(&MapRenderer::gpuWorkerThreadProcedure, this)`, and `MapRendererResourcesManager`/its running tasks retain raw pointers back to the renderer. After `releaseRendering` returns, this method tears down the framebuffers and EAGL contexts, and `onApplicationDestroyed` then sets `_mapView = nil`, allowing its `_renderer` shared pointer to be destroyed. A worker that was blocked for more than 500 ms can therefore resume before the process is actually killed and dereference freed renderer/GL state, turning the watchdog fix into a shutdown use-after-free. Please make the core abandonment path retain the renderer and its dependent GL state until process exit (or otherwise ensure abandoned workers cannot touch them) before opting iOS into the timeout.

PR #5804, opened 2026-09-17
Fix route time after Attach to roads: recalculate GPX route time and …

### LINKED ISSUE #5787: Attach to roads: route time calculated as pedestrian after snapping (Driving profile shows ~5h instead of ~46min)
## Summary

After applying **Attach to roads** on a track segment with the **Driving** profile, the estimated route time switches from a correct ~46 min to ~4 h 56 min (≈ 3.6 km/h — walking speed), even though the distance barely changes (17.8 km → 17.7 km). The incorrect time persists in the navigation start banner.

## Steps to Reproduce

1. Navigate → **Follow track** → select "Thu 13 Aug 2026", Segment 1
2. Tap **Attach to the roads** → **Continue**
3. Profile: **Driving**, threshold: **50 m** → **Apply**
4. Return to route preparation / start navigation

## Observed vs. Expected

| State | Distance | Time |
|---|---|---|
| Before attachment | 17.8 km | **46 min** ✅ |
| After Apply (Driving) | 17.7 km | **4 h 56 min** ❌ |
| In navigation | 18 km | 4 h 56 min ❌ |

4 h 56 min for 17.7 km ≈ **3.6 km/h** — this matches a pedestrian/walking speed, not a driving profile.

## Screenshots

<table style="width:100%">
  <tr>
<th> </th>
 <th> </td>
  </tr>
</table>

**Before attachment — 17.8 km · 46 min:**
_(route preparation screen showing correct time)_

**After Apply — 17.7 km · 4 h 56 min on Driving profile:**
_(route preparation screen showing wrong time)_

**Navigation start — OsmAnd Maps 580 ms hang banner + 4 h 56 min:**
_(system hang indicator visible at top of screen)_

## Device / Build

- **Build:** OsmAnd iOS 5.4.1 (5459), branch `r5.4`, commit `7dadb5d`, built 11 Sep 2026
- **Device:** iPad Air 5th gen, iPadOS 26.5, landscape 1180×820 pt
- **Source:** TestFlight

## Notes

- PR [#5740](https://github.com/osmandapp/OsmAnd-iOS/pull/5740) fixed the main Attach to roads regression (empty Plan Route instead of track points — see #5735). This is a **separate regression** introduced or exposed in the same flow.
- Suspected root cause: after snapping, the routing time is recalculated using the wrong profile or a default pedestrian speed instead of the selected navigation profile.
- **Suggested verification:** compare behaviour in OsmAnd 5.3 iOS and OsmAnd Android with the same track and the same Driving profile; check whether `routingHelper` picks up the correct profile after `attachToRoads` applies the approximated GPX.
- Found during automated XCUITest run on 11 Sep 2026 (build 5459 regression sweep).

## Related

- #5735 — Attach to roads opened empty Plan Route (fixed in #5740)
- #5740 — PR that fixed #5735 (confirmed in 5459)

vitaliy-sova-ios:
Fix route time after Attach to roads: recalculate GPX route time and speed

## Related issue / task

Fixes #5787

## Summary

**Problem.** After applying *Attach to the roads* to a followed track, the estimated route
time changed from ~46 min to ~4 h 56 min for virtually the same distance (17.8 → 17.7 km),
i.e. ≈3.6 km/h, and the wrong value persisted in the navigation start banner.

**Root cause.** It is neither a wrong profile nor the pedestrian default speed: no iOS
profile has a 1.0 m/s default (PEDESTRIAN is 1.11 m/s, which would yield 4 h 26 min for
this distance, not 4 h 56 min). The value is the track's own recorded pace, which reaches
the route this way:

1. `speed` of the track points is copied into the route locations —
   `OARouteProvider.createLocation:`.
2. The *Attach to the roads* screen operates on a GPX rebuilt from the current route, and
   `makeGpxFromRoute` synthesises point timestamps out of those speeds —
   `OAGPXUIHelper.mm:116`, called from `OARouteInfoView.mm:1070`.
3. The approximation treats them as external timestamps and overwrites `segmentSpeed` /
   `segmentTime` of every segment — `OARouteProvider.mm:127`.
4. Those values are written into the attached GPX (`speed`, `segmentTime`) by
   `OARouteExporter` and read back unchanged by `OARouteImporter`.
5. `calculateGpxRoute:` returns the segments as is, so the directions get
   `averageSpeed = distance / segmentTime` and the ETA equals the recording duration.

Steps 1–4 behave identically on Android. Step 5 is the divergence:
`GpxRouteHelper.calculateGpxRoute` recalculates time and speed with the current profile
whenever `!calculatedRouteTimeSpeed` (GpxRouteHelper.java:59-62). On iOS the counterpart
`calculateGpxRouteTimeSpeed:gpxRouteResult:` existed but was never called from anywhere —
dead code since the port, and `calculatedRouteTimeSpeed` is never set.

**Fix.** Call it, at the same point Android does. The method itself is also cleaned up: the
segment vector is taken by reference instead of by value, the native routing lock is
acquired the same way `findVectorMapsRoute` does, and empty/`nullptr` inputs are guarded.

**Behaviour change worth knowing.** For a GPX planned with different profiles per part, the
ETA is now computed entirely with the current navigation profile instead of the per-part
speeds stored in the file. That is exactly what Android does — `calculateTimeSpeed` takes a
single router from the context — so this PR removes a divergence rather than creating one.
Geometry and turns stay per-part; only time and speed are unified.

## Testing

**Tested on:**

- Device: iPhone 11
- iOS: 27.0

### Scenarios

- Issue repro on a synthetic 17.8 km track carrying `<speed>` on every point (recorded pace
  3.6 km/h, 4 h 56 min): before the fix 23 min before Apply → **4 h 51 min** after Apply;
  after the fix **40 min**.
- Negative control — the same geometry without `<speed>` on the points: the time stays
  correct both before and after Apply, and the fix does not change it. Confirms the entry
  point of the bug is the point speed, not the point timestamps.
- Long track: 207 km, 2244 points, Kyiv and Kyiv oblast, attached with the Driving profile
  and a 50 m threshold, 527 route segments. 8 route recalculations triggered by switching
  profiles — car, bicycle, pedestrian, truck, motorcycle, moped — the recalculation runs
  for each of them and the route is built correctly.
- Performance A/B on the same long track, with the new call enabled and disabled through a
  temporary kill switch (see below).

## Relevant checks

- [x] Different profiles tested
- [ ] CarPlay checked
- [x] Android parity checked
- [x] Performance impact checked

No UI code is touched, so orientation, theme, localization and accessibility are not
affected. CarPlay consumes the same `OARoutingHelper.getLeftTime` value and therefore gets
the fix for free, but it was not exercised separately.

**Android parity.** Condition, guard, call site and the position relative to the reverse
branch match `GpxRouteHelper.calculateGpxRoute`. The connecting first/last segments are
added after the recalculation on both platforms. `RouteResultPreparation.calculateTimeSpeed`
(Java) and `calculateTimeSpeed` (C++) are line-for-line equivalent — same Naismith/Scarf
constants, same 15 m/s slow-down threshold, same `defineVehicleSpeed` → `getDefaultSpeed`
fallback — so the numbers match, not merely the structure.

**Performance.** Measured with temporary instrumentation on the 207 km / 527-segment route,
42 installed map resources, 9 samples:

| Phase | Median | Mean | Min–max |
|---|---|---|---|
| Routing environment build | 18.1 ms | 19.5 ms | 13.8–34.0 |
| `calculateTimeSpeed` (527 segments) | 27.0 ms | 27.1 ms | 22.8–31.1 |
| **Total per route calculation** | **45.1 ms** | **46.6 ms** | 40.3–56.7 |

Against ~160 ms of total route calculation time for this track, on a background thread.
Native routing lock wait was 0.001 ms. `checkInitialized` reports `newlyInitialized = 0`
from the second call onwards, so no map files are re-opened. Scaling: ~51 µs per route
segment, and the environment build grows with the number of installed maps (~9 ms of it is
the resource scan at 42 resources).

Caching the routing environment was considered and rejected — it would remove only the
19 ms half, while the remaining 27 ms is the work Android performs as well.

## AI disclaimer

**Implementation:**

- Tool / Agent: Claude Code (desktop app)
- Model: Claude Opus 5

1. Investigate, read-only, whether the behaviour described in the issue is possible at all
   and where it originates; no code changes.
2. Generate GPX fixtures that reproduce it, then correct them after the first attempt
   failed on the device, which disproved the initial hypothesis about point timestamps.
3. Cross-check the analysis against the issue text and the attached screenshots.
4. Implement the fix.
5. Code review of the change against Android, looking for regressions.
6. Measure the performance cost of the added call on a long track.

- Call site placed inside the `gpxRouteResult.size() > 0` branch so it covers both return
  paths and runs after the reverse-route-points branch, mirroring Android.
- Signature changed from by-value to by-reference for the segment vector.
- `@synchronized (_nativeRoutingLock)` added around the environment build; Android has no
  lock there, but every other native routing entry point on iOS takes it. Lock ordering was
  checked against `checkInitialized`'s `@synchronized (self)` — no inversion.
- `!gpxParams.calculatedRouteTimeSpeed` guard kept although the flag is never set on iOS,
  for parity and as the hook for a future online GPX engine.
- A related precision divergence in `makeGpxFromRoute` (`(long)(dist / speed) * 1000` on
  iOS vs `(long)(dist / speed * 1000)` on Android) was deliberately left out of scope.
- An unrelated optimisation opportunity — `containsData` re-evaluated on every
  `checkInitialized` for non-matching resources, ~9 ms on each route calculation
  app-wide — was left for a separate PR.

**Final review:**

- Tool / Agent: Claude Code (desktop app)
- Model: Claude Opus 5
- [x] Final diff reviewed

Reviewed by the same agent and model that wrote the change, i.e. not an independent review.

**Significant findings:** the multi-profile ETA behaviour change described in the Summary;
no other correctness or crash findings survived the review. The initially suspected cost of
`writeMapFilesCache` was measured at 0.008–0.022 ms and dismissed.


tigrim:
merge conflict

tigrim (approved):

alex-dev-neo (changes requested):
**Requesting changes: one blocker**

`calculateGpxRouteTimeSpeed:` builds a full shared routing environment, but it only needs the router. `TurnPreparation.calculateTimeSpeed` reads nothing from the request except `getRouter()`. Going through `calculateSharedRoutingEnvironment:` has side effects on this path:

- it calls `checkInitializedForParams:` and `sharedRouteReaders`, so with `useSharedRouting` off it opens an `OASBinaryMapIndexReader` for every routing obf. That is the ~600 ms mentioned in the description, and those readers then stay open until the installed maps change, with nothing else in that configuration using them;
- it takes `_nativeRoutingLock`, which the whole-route GPX branch never took before. A recalculation of a followed track now waits for any search holding the lock. For example, `calculateSharedGpxApproximation` holds it for the whole `searchGpxRouteGctx` call while a track is being attached to roads.

A plain `RoutingRequest` is enough:

```objc
- (void) calculateGpxRouteTimeSpeed:(OARouteCalculationParams *)params gpxRouteResult:(NSArray<OASRouteSegmentResult *> *)gpxRouteResult
{
    OsmAndAppInstance app = [OsmAndApp instance];
    OASRoutingConfigurationBuilder *builder = [app getSharedRoutingConfigForMode:params.mode];
    OASGeneralRouter *generalRouter = [app getSharedRouter:builder mode:params.mode];
    if (!generalRouter)
        return;

    OASRoutingConfiguration *cf = [self buildSharedRoutingConfig:builder params:params generalRouter:generalRouter];
    OASRoutingRequest *request = [[OASRoutingRequest alloc] initWithConfig:cf calculationMode:OASRouteCalculationMode.normal];
    [OASTurnPreparation.shared calculateTimeSpeedRequest:request result:gpxRouteResult];
}
```

`getSharedRouter:` and `buildSharedRoutingConfig:` are already called outside the lock in `calculateSharedRoutingEnvironment:`, so this adds no new unlocked path. I haven't built or run this snippet.

With this change, the edit to the comment on `calculateSharedRoutingEnvironment:` can be reverted, because the track approximation is its only caller again. As written, the new comment is also inaccurate: this caller passes a real start and end, and `checkInitializedForParams:` uses them.

**Non-blocking**

- Recalculating after the reversed-route branch, mutating the segments in place and redoing the calculation on every recalculation all match Android's `GpxRouteHelper`. Fine as is.
- When no shared router can be built, the recalculation is silently skipped and the recorded-pace ETA remains. A log line would help diagnose it.

PR #5805, opened 2026-09-17
Add activity for Plan route synchronization with Android

### LINKED ISSUE #5736: Plan a route saves the profile's default activity instead of the selected one
## Description

When a track is saved from **Plan a route**, the activity written into the GPX metadata ignores the activity configured in the profile settings and sets it to **None** (empty).

In contrast, on Android, it falls back to the profile's default activity, whereas on iOS the value is omitted entirely.

### Steps to reproduce

1. Enable the **Trip recording** plugin.
2. Go to **Settings** → **Cycling** → **Trip recording** → **Activity**.
3. Change the activity to **Gravel Biking** (or any non-default activity).
4. Go to **Menu** → **Plan a route** → **Create new route** → **Route between points** → select the **Cycling** profile.
5. Add route points to generate a track → tap **Done** / **Save**.
6. Open the saved track from **My Places** → **Tracks** and inspect its **Info** section.

### Actual result

* **iOS:** The track is saved with Activity set to **None** (the profile setting `Gravel Biking` is completely ignored).
* **Android (for comparison):** The track is saved with the profile setting `Gravel Biking`

<table style="width:100%">
  <tr>
<th> </th>
 <th> </th>
 <th> </th>
  </tr>
</table>  
</details> 

### Expected result

- [x] The saved track should inherit the Activity type selected in the profile's Trip Recording settings (e.g., `Gravel Biking`) - **as in Andoid**

### Environment

**OsmAnd Version:** OsmAnd Maps 5456, released 07.09.26
**iOS Version:** iOS 27.1
**Device model:** iPhone 15

DmitrySvetlichny:
## Related issue / task
Fixes https://github.com/osmandapp/OsmAnd-iOS/issues/5736

## Summary
* Tracks saved from **Plan a route** were written without the activity selected in the profile's **Trip recording → Activity** setting (Activity was None). The activity handling was removed together with `OASaveGpxRouteAsyncTask` in the Plan Route refactor (#5484). The earlier iOS implementation (#5414) also used the profile's default activity instead of the selected one and had no multi-profile check.
* Ported Android's `SaveGpxRouteAsyncTask.savePreselectedRouteActivity` and `shouldSyncRouteActivityWithEditedProfile` to `OAPlanRouteEditingBridge`. They run before the file is written in **Save**, **Save as…**, **Save as copy** and **Append to existing track**.
* Added `OAMeasurementEditingContext.isInMultiProfileMode`, ported from Android `MeasurementEditingContext`.
* The activity comes from `currentTrackRouteActivity` for the route profile. Straight-line routes use the current app profile. The activity is synced with the profile only when an edited track has route points and uses a single profile; otherwise an existing activity is kept.
* As on Android, saving an edited track to the same file and appending to a track keep the source track's metadata (description, keywords, creation time, activity) instead of replacing it with empty metadata. New files and copies get empty metadata.

## Testing
**Tested on: iPhone 17 Pro simulator

### Scenarios
* **Cycling → Trip recording → Activity** set to **Gravel Biking**, new Cycling route → **Save** and **Save as…** → the track activity is Gravel Biking.
* New Walking route with the default activity → Walking. Cycling activity set to None → no activity.
* Straight-line route with the Cycling app profile → Gravel Biking; with Browse map → no activity.
* Recorded track with Hiking and a description → **Edit** → **Save** → Hiking and the description are kept.
* Planned single-profile (Walking) track with Hiking → **Edit** → **Save** → Walking (synced with the profile, as on Android), the description is kept.
* The same track edited with two profiles → Hiking is kept.
* Track with "й" in the file name → **Edit** → **Save** → metadata is kept, no duplicate track.
* **Save as copy** → the activity of the profile, no description; the original track is unchanged.
* **Append to existing track** into a track with Hiking → activity and description are kept; into a track without an activity → the profile activity.
* Editing points while **Append to existing track** is saving → no crash.
* Trip recording, segment saving and navigation from Plan Route work as before.

## Relevant checks
* [x] Different profiles tested
* [x] Localization / RTL checked
* [x] Android parity checked
* [x] Performance impact checked

## AI disclaimer
**Implementation:**
* Tool / Agent: Claude Code
* Model: Claude Opus 5

1. Analyze issue #5736 and the history of activity handling in Plan Route; check whether it is a regression of the new Plan Route UI.
2. Compare with Android `SaveGpxRouteAsyncTask` and `MeasurementEditingContext` and review the initial fix: minimal, Android parity, no regressions.

**Final review:**
* Tool / Agent: Claude Code
* Model: Claude Opus 5
* [x] Final diff reviewed
**Significant findings:**
* `ctx.appMode` is set the same way as on Android (`ChangeRouteModeCommand`, opening a track), so the ported logic receives the same inputs.
* Copied `metadata.bounds` do not affect track bounds in OsmAnd (they are computed from points). `metadata.time` keeps the original creation time, as on And


tigrim (commented):
[P2] Assign an activity when saving a segment

`saveSegmentWithPointIndexes` exports a fresh GPX and writes it directly, bypassing `savePreselectedRouteActivity`, so saving an individual Plan Route segment produces a file without `osmand:activity`. Please assign the activity before writing, deriving it from the selected point range rather than from the whole editing context because a single-profile segment can belong to a multi-profile route.

Sources/Controllers/PlanRoute/OAPlanRouteEditingBridge.mm
```diff
@@ -1741,6 +1741,10 @@ - (void)performSaveWithFileName:(NSString *)fileName
     NSString *gpxRootPath = OsmAndApp.instance.gpxPath;
     NSString *folderPath = (folder.length > 0) ? [gpxRootPath stringByAppendingPathComponent:folder] : gpxRootPath;
     NSString *outFile = [[[folderPath stringByAppendingPathComponent:trackName] stringByAppendingPathExtension:@"gpx"] stringByStandardizingPath];
+    if (!asCopy && [originalGpxPath.decomposedStringWithCanonicalMapping isEqualToString:outFile.decomposedStringWithCanonicalMapping])
```
tigrim, line 1744: [P1] Preserve the source activity when saving a copy

`exportGpx` creates a GPX with fresh metadata, but the source metadata is copied only when `asCopy` is false and the output path matches the original path. For an existing multi-profile route, `shouldSyncRouteActivityWithEditedProfile` is false, so Save As Copy (or Save As under another path) will either replace the explicitly selected route activity with the current profile preference or remove it when that preference is empty. Android copies the metadata for every edited GPX before calling `savePreselectedRouteActivity`; please preserve the source metadata here as well and let the helper override only the single-profile synchronization case.

DmitrySvetlichny: I checked the Android caller chain: “Save as new track” calls saveNewGpx, which passes null to SaveGpxRouteAsyncTask. This creates fresh metadata and assigns the activity from the profile’s setting, even when editing an existing track. Our iOS condition follows the same behavior: source metadata is preserved when saving changes to the original file, while a new file or copy gets the profile’s configured activity. For example, if the source activity is Hiking and the profile is configured with Gravel Biking, the new track gets Gravel Biking on both platforms; if the profile setting is None, both leave it unset. I suggest keeping the current condition for Android parity, since preserving the source activity in copies would be a separate behavior change.

tigrim (approved):

PR #5811, opened 2026-09-18
Add a tag with the OSM object URL to the favourite

### LINKED ISSUE #5749: Save OSM link for Favorite point
- [x] Add a tag with the OSM object URL to the favourite point if the OpenStreetMap plugin is on.

DmitrySvetlichny:
## Related issue / task
Fixes #5749

Android: https://github.com/osmandapp/OsmAnd/issues/25464, https://github.com/osmandapp/OsmAnd/pull/25738

## Summary
- When the OpenStreetMap editing plugin is enabled, a favorite added from the context menu now stores the OSM object URL in the `osm_url` GPX extension (same as Android PR #25738).
- The URL comes from the same source objects as on Android: POI, place details (`BaseDetailsObject` synthetic amenity), map objects without a POI (`OARenderedObject`) and OSM edit points (POI lookup by OSM id).
- `osm_url` is written to the favorites GPX in `OAFavoriteItem.toWpt`. It survives an app restart and is included in export, sharing and backup.
- `osm_url` is hidden from the POI info rows in `AdditionalInfoBundle`, same as Android.

## Testing
**Tested on: iPhone 17 Pro simulator

### Scenarios
- Enable the OpenStreetMap editing plugin.
- Add a POI (node) to favorites → GPX contains `<osmand:osm_url>https://www.openstreetmap.org/node/9806497618</osmand:osm_url>`.
- Add an area without a POI (`landuse=residential`, way) → `https://www.openstreetmap.org/way/1333681864`.
- Add an unnamed cafe (node) → `https://www.openstreetmap.org/node/2324856936`.
- Add a location that has no OSM object → no `osm_url`.
- Checked through the OSM API that the saved URLs point to the correct objects (names, types and coordinates match).

## Relevant checks
- [x] Android parity checked

**Implementation:**
Personal implementation/Agent: Claude Code

1. Keep the change narrow: store only the OSM URL and leave the existing POI handling of favorites unchanged.
2. Check every construct against existing patterns in the project and against Android.
3. Do a final code review, walk through the test scenarios in the code, and write a device test plan.

**Final review:**
- Tool / Agent: Codex
- [x] Final diff reviewed

**Significant findings:**
- On iOS the context-menu target of a POI is replaced with a `BaseDetailsObject` (`presentContextMenu`), so `getTargetPointPoi` returns `nil` and the favorite handler gets no POI. `setOsmUrl:` therefore takes the source object from the current target point, the same way `loadEntity:` in `OAOpenStreetMapLocalUtil`/`OAOpenStreetMapRemoteUtil` does.
- iOS `toWpt` writes only a fixed set of keys, and `WptPt.setAmenity` does nothing (`extensionsWriters` is `nil` for a new `WptPt`). That is why `osm_url` is written explicitly. Favorite amenity tags are not exported on iOS: this is an existing issue outside this PR.
- For a POI the saved URL is the same as the OSM link row in the context menu (`getOsmUrlForId` on the same object).
- For OSM edit points the POI lookup runs synchronously on the main thread, as on Android, and only when the plugin is enabled.


tigrim (approved):

Sources/Controllers/TargetMenu/PointEditing/OAFavoriteEditingHandler.mm
```diff
@@ -64,6 +67,38 @@ - (instancetype) initWithLocation:(CLLocationCoordinate2D)location title:(NSStri
     return self;
 }
 
+- (void)setOsmUrl:(OAPOI *)poi
+{
+    if (![OAPluginsHelper isEnabled:OAOsmEditingPlugin.class])
+        return;
+    
+    id object = poi ?: [OARootViewController.instance.mapPanel getCurrentTargetPoint].targetObj;
```
alex-dev-neo, line 75: `setOsmUrl:` takes the source object from the map panel's current target point rather than from its caller. This works today because `targetPointAddFavorite` is the only path to `initWithLocation:title:address:poi:`, and the target point is still set when the handler is created.

The handler, however, silently depends on whatever the context menu happens to show. Suppose a favorite is later added from another place (search results, a list, a quick action) while a context menu is open. It would then get the `osm_url` of an unrelated object, with no error.

Could the target object be passed explicitly instead? For example, `targetPointAddFavorite` could pass `self.targetMenuView.targetPoint.targetObj` through `OAEditPointViewController` into the handler initializer, next to `poi`. Then the handler would use only what its caller gives it.

Not a blocker, since the current single caller behaves correctly.

DmitrySvetlichny: Good point, thanks. Fixed: `targetPointAddFavorite` now passes `targetPoint.targetObj` through
`OAEditPointViewController` into `OAFavoriteEditingHandler`, and `setOsmUrl:targetObject:` uses only
its own parameters (`poi ?: targetObject`), so the handler no longer reads the map panel.

Behaviour is unchanged, because the caller passes exactly the object the handler used to look up
(`getCurrentTargetPoint` returns the same `targetMenuView.targetPoint`). The waypoint call sites and
`OAPlanRouteEditingBridge` pass `nil`, and the waypoint handler does not take the parameter at all.

This also brings the flow closer to Android, where `FavoritePointEditor.add(latLon, title, address, object)`
receives the object from `MapContextMenu.buttonFavoritePressed()` instead of looking it up itself.

PR #5814, opened 2026-09-18
Fix waypoint color when selecting or reselecting a group

### LINKED ISSUE #5644: Clicking on a point group always applies the red color to the point
### Description

When you select a group for a new waypoint, it defaults the color do red, regardless of that group color

### Steps to reproduce

- Create new waypoint
- Create any new groups with different colors
- Select any color for that point
- Select different groups

### Actual result

It always defaults back to red

https://github.com/user-attachments/assets/ae69639d-5b3c-4f5b-8eda-f032901400a9

### Expected result

Should apply the color of the corresponding group

```
OsmAnd Version: 5.4.0
iOS version: 27 
Device model: iPhone mini 13
```

aleksandr-tata:
# Fix waypoint color when selecting or reselecting a group

## Related issue / task

Fixes https://github.com/osmandapp/OsmAnd-iOS/issues/5644.

## Summary

After choosing a manual waypoint color, selecting or reselecting Waypoints could turn the point red because the group-color lookup omitted the default group. Pending groups also used incorrect lookup keys.

- Build the selection lookup from the same groups shown by the editor, including the default and pending groups.
- Use configured group-color metadata when available. Preserve the existing point-derived fallback for legacy groups; use the waypoint default when no color is available.
- Keep the selected palette item synchronized when a group color is not yet in the local palette. An unknown group does not invoke the red fallback.

Production changes are limited to `OAGpxWptEditingHandler.mm` and `OAEditPointViewController.mm`. GPX saving, recording/SQLite, group-appearance editing, the Points tab and route planning are unchanged. Independent persistence of group and point colors is outside this PR.

## Testing

**Tested on:**

- Host: macOS, targeted iOS ARM64 syntax checks and isolated Objective-C++ XCTest harness.
- Simulator: iPhone 13 / iOS 26.5; all 11 tests in the reduced suite passed.
- Physical device: verification of this reduced revision pending.

### Automated verification

- [x] Seven handler tests: original code produces 10 failed assertions; the fix passes all seven tests.
- [x] Both changed production files pass targeted iOS ARM64 syntax checks.
- [x] Eleven simulator tests passed with no failures (`/tmp/osmand-5644/minimal/result.xcresult`).
- [x] Final diff and Xcode project plist validation.

Coverage: default group, named and pending groups, explicit group metadata differing from point color, legacy fallback, missing color, configured red, group visibility and source-point preservation, repeated selection, card/list consistency, missing local palette item, and unknown group.

The isolated harness stubs UIKit/GPX dependencies. Simulator tests compile the two changed production classes under temporary names into an XCTest bundle hosted by an existing application build. They invoke editor callbacks programmatically using real UIKit/GPX dependencies; they do not perform screen taps. No full project/workspace build was run.

### Scenarios

- [ ] Follow the issue steps: choose blue, tap Waypoints, and verify the configured group color is applied instead of red.
- [ ] Choose another manual color and tap the same group again: icon and selected palette color match the group.
- [ ] Switch between named groups with different configured colors.
- [ ] Select a newly created group with a custom color.
- [ ] Confirm an explicitly red group still applies red.

## Relevant checks

- [ ] CarPlay checked
- [ ] Android parity checked

## AI disclaimer

**Implementation:**

- Tool / Agent: Codex
- Model: GPT-6

1. Investigate and fix issue #5644, with regression tests.
2. Review the changes for scope and regressions.
3. Limit the final fix to applying the configured group color when selecting or reselecting a waypoint group.

- Reuse the editor group list for color lookup instead of maintaining a second lookup implementation.
- Preserve legacy color fallback and guard unknown groups.
- Use targeted compilation and an application-hosted test bundle instead of a full build.

**Final review:**

- Tool / Agent: Codex (same agent as implementation)
- Model: GPT-6
- [x] Final reduced diff reviewed

**Significant findings:**

- The default Waypoints group was missing from the color lookup.
- Pending-group entries were stored under literal title/color keys rather than the group name.
- A group color newly added to the shared palette must also be reflected in the editor's local palette.


Sources/Controllers/TargetMenu/PointEditing/OAGpxWptEditingHandler.mm
```diff
@@ -226,45 +238,12 @@ - (OASGpxFile *)getGpxDocument
...
-        newGroups[@"title"] = _newGroupTitle;
-        newGroups[@"color"] = _newGroupColor.toHexARGBString;
-        groups = newGroups;
-    }
-
-    return groups;
+    NSMutableDictionary<NSString *, NSString *> *colors = [NSMutableDictionary new];
+    for (NSDictionary<NSString *, NSString *> *group in [self getGroups])
+        colors[group[@"title"]] = group[@"color"];
+    return colors;
```
tigrim: `getGroupsWithColors` uses the localized display title as the group identifier. An imported GPX can contain a real category named `Waypoints` (or its localized equivalent), in which case it collides with the display title of the default group. One entry will silently overwrite the other in this dictionary, so selecting the default card may apply the named group’s color. Please keep the raw category key (`@""` for the default group) separate from its display title, or apply the color directly from the selected group record instead of performing another lookup by title.

Sources/Controllers/TargetMenu/PointEditing/OAGpxWptEditingHandler.mm
```diff
@@ -178,7 +178,19 @@ - (OASGpxFile *)getGpxDocument
...
+        NSMutableDictionary<NSString *, NSString *> *group = [category mutableCopy];
+        int color = _gpxDocument.pointsGroups[group[@"title"]].color;
+        if (color == 0)
+            color = [UIColor toNumberFromString:group[@"color"]];
+        if (color == 0)
+            color = [[OADefaultFavorite getDefaultColor] toARGBNumber];
+        group[@"color"] = UIColorFromARGB(color).toHexARGBString;
+        [groups addObject:group];
+    }
+    return groups;
```
tigrim: These methods have immutable return types, but the implementation returns `NSMutableArray`/`NSMutableDictionary` instances, including mutable dictionaries nested in the groups array. Please return immutable copies (for example, `[group copy]`, `[groups copy]`, and `[colors copy]`) so the runtime behavior matches the declared API contract.

OsmAnd MapsTests/PointEditing/OAWaypointGroupSelectionTest.mm
```diff
@@ -0,0 +1,137 @@
...
+    XCTAssertEqual(point.getColor, UIColor.redColor.toARGBNumber);
+    XCTAssertEqual(file.getPointsList.count, 1);
+}
+
+- (void)testPendingGroupSelectionRefreshesMissingPaletteItem
+{
+    OAEditPointViewController *editor = [self editorWithFile:[[OASGpxFile alloc] initWithAuthor:@"test"]];
+    OAGpxWptEditingHandler *handler = [editor valueForKey:@"pointHandler"];
+    UIColor *color = UIColorFromARGB(0xFF123456);
+    [handler setGroup:@"New group" color:color save:NO];
```
tigrim, line 152: This test calls `setGroup`, which reaches `getColorItemWithValue`; for `0xFF123456`, that method creates and persists a palette item. There is no teardown restoring the palette, so the app-hosted test leaks state into subsequent tests or runs and may hide the missing-palette scenario it is intended to cover. Please remove the item during teardown or use an isolated palette repository.

Sources/Controllers/TargetMenu/OAEditPointViewController.mm
```diff
@@ -1320,6 +1333,7 @@ - (void)addGroupWithName:(NSString *)name
         if (!_pointHandler.gpxWptDelegate)
             _pointHandler.gpxWptDelegate = self.gpxWptDelegate;
         [((OAGpxWptEditingHandler *) _pointHandler) setGroup:editedGroupName color:color save:YES];
+        _selectedWaypointGroupKey = editedGroupName;
```
tigrim, line 1336: The selected raw key is updated here, but the completion block below still resolves `selectedIndex` through `_groupNames` (display titles). If the new category is named `Waypoints` or matches the localized default-group title, that lookup resolves to the default card and scrolls to the wrong item. Please use `_waypointGroupKeys` for waypoint groups there, consistently with `generateData` and `onGroupSelected:`.

aleksandr-tata:
Added a follow-up fix in 4a52875cc8 for an issue found during manual verification.

Creating several groups before saving a waypoint previously kept only the last new group: creating `Test5` made `Test4` disappear. The editor stored just one pending group name/color. It now retains all pending groups, matching the current Android behavior for this GPX-file workflow.

- All newly created groups remain available for selection.
- Saving the waypoint also saves the new groups, including empty groups, with their color, icon and background shape. Changing the waypoint color does not overwrite the new group's configured color.
- Discarding the editor does not write pending groups to the GPX document. Existing group metadata is preserved.

This extends the original scope described in the PR: pending group metadata now passes through the existing waypoint save/update paths, including the route-planning delegate. The previous fix for scrolling by raw category key remains in place.

Validation: 19 tests passed on iPhone 13 / iOS 26.5 Simulator, covering multiple pending groups, GPX write/reload with an empty group, discarding pending groups, existing-point editing and duplicate prevention. Targeted iPhone arm64 compiler checks passed for the affected production classes, as did `git diff --check`. Simulator tests use the changed classes in an isolated test bundle hosted by an existing app build; no full rebuilt-app or manual UI verification is claimed.

The separate SharedLib compatibility change in `OASavingTrackHelper.mm` is excluded from this commit.

OsmAnd MapsTests/PointEditing/OAWaypointGroupSelectionTest.mm
```diff
@@ -0,0 +1,339 @@
...
+    XCTAssertEqual(delegate.file.pointsGroups.count, 2);
+    XCTAssertEqual(delegate.file.pointsGroups[@"Test5"].color, UIColor.greenColor.toARGBNumber);
+    XCTAssertEqual(delegate.file.getPointsList.firstObject.getColor, UIColor.redColor.toARGBNumber);
+    XCTAssertEqual(delegate.file.pointsGroups[@"Test4"].points.count, 0);
+    XCTAssertEqualObjects(delegate.file.pointsGroups[@"Test5"].backgroundType, @"square");
+
+    NSString *path = [NSTemporaryDirectory() stringByAppendingPathComponent:[NSUUID.UUID.UUIDString stringByAppendingPathExtension:@"gpx"]];
+    [self addTeardownBlock:^{ [[NSFileManager defaultManager] removeItemAtPath:path error:nil]; }];
+    OASKFile *file = [[OASKFile alloc] initWithFilePath:path];
+    XCTAssertNil([OASGpxUtilities.shared writeGpxFileFile:file gpxFile:delegate.file]);
```
tigrim, line 272: This part of the test verifies the shared library’s GPX serialization round trip rather than the iOS adapter behavior introduced by this PR. That contract should be covered in the shared/Android test suite. The iOS test can stop after verifying that all pending groups and their metadata are handed to the destination file/delegate correctly.

OsmAnd.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,15 +0,0 @@
-{
```
tigrim, line 1: This tracked resolution file is unrelated to the waypoint-color change and was present in both parents of the merge commit, but the merge result deletes it. Please restore `Package.resolved` so this PR does not unintentionally remove the workspace dependency lock.

Sources/Controllers/Map/OAMapViewController.mm
```diff
@@ -3823,11 +3823,22 @@ - (BOOL) saveFoundWpt
...
+{
+    return [self addNewWpt:item.point gpxFileName:gpxFileName groupsFromItem:item];
+}
+
+- (BOOL)addNewWpt:(OASWptPt *)wpt gpxFileName:(NSString *)gpxFileName groupsFromItem:(OAGpxWptItem *)item
 {
     if (!gpxFileName)
     {
         OASavingTrackHelper *helper = [OASavingTrackHelper sharedInstance];
+        [item applyPendingGroupsToFile:helper.currentTrack];
```
tigrim, line 3841: Applying the pending groups only to `helper.currentTrack` does not persist their metadata in the recording database. `addWpt:` writes just the selected waypoint row (category plus the point color/icon/background), and `collectDBPoints` later reconstructs groups exclusively from those rows. After an app restart, empty pending groups disappear, and the selected group’s configured appearance can be replaced by the point’s appearance (for example, a green group containing a red point is reconstructed as red). Please persist the pending group metadata through the recording/SQLite path as well, or otherwise restore it independently of waypoint rows.

aleksandr-tata:
Pushed 927d7e6e3e to narrow this PR back to the original issue #5644: selecting or reselecting a waypoint group should apply that group's color instead of unexpectedly switching the point to red.

Retained the review fixes for raw group keys versus display titles (including a named `Waypoints` category), scrolling to the correct group, immutable return values, palette synchronization, and test palette cleanup.

Removed the additional change that preserved multiple newly created groups when saving a waypoint, including its GPX/model, map, and route-planning integration and persistence tests. This PR no longer changes the recording/SQLite persistence path or introduces independent persistence of group appearance. The restart/empty-group persistence scenario raised in review is outside the reduced scope; it is not being claimed as fixed here and should be handled separately if required.

Restored `Package.resolved` exactly as in the PR base, without dependency updates. The remaining production changes are limited to the waypoint editing handler, the point editor, and the group-selection list.

Validation: all 15 tests in the reduced simulator suite passed; targeted iOS ARM64 syntax checks and diff/project validation passed. SwiftLint has no new warnings. Simulator tests invoke the changed editor classes in a test bundle hosted by an existing app build; no full application build was run.

PR #5815, opened 2026-09-19
Read the travel guides through OsmAndShared

alex-dev-neo:
Reads the travel guides through OsmAndShared instead of OsmAndCore. First of the three iOS steps of the OBF reader port; the shared side is [osmandapp/OsmAnd#26114](https://github.com/osmandapp/OsmAnd/pull/26114).

Issue: https://github.com/osmandapp/OsmAnd-Issues/issues/2935

## What moves

`TravelObfHelper.swift` becomes a wrapper over the shared helper, 987 lines down to 294. Its public surface is unchanged, so none of the 27 call sites needed touching. The travel half of `OATravelGuidesHelper` goes with it: the four `searchAmenity:` variants, `createWptPt:`, `fetchSegmentsAndPoints:`, `buildGpxFile:` and `searchTravelGpx:` - 442 lines.

What stays on the app side is the app's own `TravelArticle` model, the background task with the spinner that OsmAndShared has no notion of, and writing a built gpx to disk and into the tracks database.

New:

- `OAObfFileList` - the installed travel and map files by path, in the order android's `AmenitySearcher.getAmenityRepositories` consults them.
- `SharedObfReaders` - one reader and one repository per file, opened lazily, closed only once no travel work is in flight.
- `SharedTravelContext` - the `TravelObfContext` and `TravelBookmarks` hooks, plus the conversion between the app's article model and the shared one.

## Three behaviours now follow Android

- `getGPXName` sanitizes the file name and falls back to the route id; it used to replace only `/` and `"`.
- `isTravelGpxTags` requires a `route_id`; the brackets were grouped differently here.
- A search that has been superseded by a newer keystroke gives up where it stands instead of reading every travel file to the end.

## Two decisions worth objecting to

The Swift `TravelArticle`/`TravelGpx` stay the UI model - 27 files know them - and the shared model is converted at the boundary. Making the shared model the UI model would turn this PR into a rewrite of all of Travel Guides.

`OARouteProvider._sharedReaders` is **not** merged into the new reader pool. Routing has its own lock and its own lifetime for stale readers around live routing environments; merging the two is a separate task.

## Not needed after all

The plan listed an `Amenity` -> `OAPOI` converter for this step. It turned out unnecessary: the shared helper hands back `TravelArticle`/`TravelGpx` directly. `OAPOI` will be needed in I2, where the map selection works with amenities.

`searchGpxMapObject` is kept in `OATravelGuidesHelper`: the OSM routes v1 click path in `OAMapSelectionHelper` still uses it. Whether that path goes away is the open question for I2.

## Verified on the simulator

Debug build, iPhone 17, 17 map regions plus `Europe_wikivoyage.travel.obf`.

OSM routes: tapping a route in Slovakia builds the track (Tatranská magistrála, 50.2 km, one segment, 815-2170 m profile); the Czech file answers too (Stezka Českem, 96.4 km); a route across the whole country (Cesta hrdinov SNP, 738 km) builds in about a minute at 100% CPU, with memory returning to where it started.

Wikivoyage: popular destinations, article text and images, contents, the navigation hierarchy, search, bookmarks, reopening a saved article (the app -> shared -> app round trip), and the gpx written to disk. For Balassagyarmat: 21 waypoints in 6 groups with their icons.

Those 21 were checked against the data rather than by eye. A throwaway test read `Europe_wikivoyage.travel.obf` with the java reader and the copy at once: both return the same 25 results for the name search, and exactly 21 of the points carry the article's route id with `lang_yes:en`. The article one level up genuinely has no points of its own - the single point the name search turns up there belongs to a different article, and the matcher is right to drop it.

## Note for whoever builds this

The shared framework has to be built from `kmp-obf-7-travel-helper` until [osmandapp/OsmAnd#26114](https://github.com/osmandapp/OsmAnd/pull/26114) is merged - master's OsmAndShared has no travel helper yet.


alex-dev-neo:
Reopened as #5816 from a branch that carries only this step. The branch behind this one picked up the commits of the next step by mistake; nothing here changed.

